Issue #11, opened 2017-05-22
Bootkube example documentation


bzub:
I've successfully created a new Kubernetes cluster with [Bootkube](https://github.com/kubernetes-incubator/bootkube) using kube-router instead of kube-proxy/flannel.  Creating this issue to remind me to document how it was done and share with this project.  Thanks!

murali-reddy:
Thats awesome @bzub Looking forward for the documentation.

bzub:
Here's the quick and dirty:

Follow [documentation](https://github.com/kubernetes-incubator/bootkube#guides) to setup arguments for `bootkube render`.  After rendering the assets:
- delete `assets/manifests/kube-proxy.yaml` and `assets/manifests/kube-flannel{,-cfg}.yaml`
- add `assets/manifests/kube-router.yaml` and `assets/manifests/kube-router-cfg.yaml` from [my gist](https://gist.github.com/bzub/b5225b7e3635cf70f6879767562dea70)
- run kubelet on a node, and then run `bootkube start` pointing to the assets directory

Right now I'm running into issue #14 with this setup, so the gist example may change if I made an error in kube-router setup.  Also I will need to figure out how to get kube-router access to the apiserver insecure port on non-master (worker) nodes, hopefully without making apiserver insecure port bound to an external IP.

bzub:
Quick question: Is there any scenario where kube-router can communicate with the apiserver over the secure port?  The README makes it sound like insecure port is only required contextually.

> caveat: Please use unsecure API port when using kubeconfig file option for now. Need further work on documentation about setting up certificate for kube-router to access API server. Kube

murali-reddy:
bzub thanks a lot for the documentation. Since i am pretty new k8s ecosystem i was kind of struggling to get automated way to get clusters up on AWS with Kube-router. Finally i am able to do so with your script :)

To answer your question. Yes, kube-router has to speak to api server over secure port. I think that caveat is not required, i will remove it.

Basically in the daemon set manifest  of kube-ruter expose certs and key files through volumeMounts in case kubeconfig has reference to files. If cert and key are embedded in kubeconfig then just volumemount for kubeconfig file should be good. I slightly modified your manifest to access API server over secure port. I will share the details in a while.

bzub:
Great, then I will also try to use the secure port for Bootkube. I'll write up more detailed docs sometime soon and share here and also with the Bootkube project.

bzub:
[...]

PR #15, opened 2017-05-26
Bootkube Integration

bzub:
Manifests and instructions for Bootkube + Kube-router.

Fixes: #11


murali-reddy:
Thanks. Much needed step for users to try out kube-router. Woderful first contribution.

Issue #9, opened 2017-05-18
Hairpin nat does not work


murali-reddy:
Accessing cluste IP from a pod, which is end point of the service fails, when traffic gets load balanced to same pod as the pod from which request originated. 

We need a way to have hair pin nat , so that this scenario works.

bzub:
I've gotten this to work by adding a rule like this for every active service endpoint:
```
iptables -t nat -A POSTROUTING -s $ENDPOINT_IP -d $ENDPOINT_IP -m ipvs --vaddr $SVC_IP -j SNAT --to-source $SVC_IP
```

The result is that loopback traffic back to the pod is seen as coming from the service IP -- a possible drawback?  Also, assuming lots and lots of services and/or endpoints, would the number of rules be too many?

If this looks reasonable I will submit a patch to implement it.

murali-reddy:
@bzub seems reasonable solution. 

"The result is that loopback traffic back to the pod is seen as coming from the service IP -- a possible drawback?"

Yeah, its has to be either node ip or service ip. Either of which will have problem with network policies if its permitting traffic only form certian set of pod IP's. But there is nothing much can be done.

"Assuming lots and lots of services and/or endpoints, would the number of rules be too many?"
We will only be adding rules for the pods running on the node. still it can lead to lots of rule.

We might need to check Kube-proxy how they are addressing this problem. Please take a look at kube-proxy. If there is better alternative go with the PR.

bzub:
I believe kube-proxy and kubelet delegates this to the CNI plugin when CNI is used.  I was able to get the "bridge" CNI plugin to set the sysctl `hairpin_mode` to `1` for all the veth devices, but that wasn't enough.  I will see about what kubernetes does with "kubenet" mode which is the only other option other than CNI.  That I believe sets hairpin-mode to "promiscuous bridge" but when I tested setting the `kube-bridge` interface to promiscuous mode it broke things.  So I'll have to look further.

murali-reddy:
@bzub yes, it seems its offloaded to CNI plugin. Not sure merit of this use case.

Please go ahed with your current fix, but keep flag '--hairpin-mode'  like kubelet does (perhaps for kubenet). So if some one wants they can take any perf hit (if there is any) and enable this flag.

bzub:
[...]

PR #42, opened 2017-07-04
network-services-controller: Support hairpin NAT

bzub:
WIP.  Do not merge.  Comments welcome in the meantime.

Still need to finish up
- [x] iptables rules compare/sync
- [x] Add iptables rule to target hairpinChain
- [x] Add to Cleanup()
- [ ] Update README/docs
- [ ] Test

For #9

### DIFF
--- a/app/controllers/network_services_controller.go
+++ b/app/controllers/network_services_controller.go
@@ -172,8 +172,9 @@ func (nsc *NetworkServicesController) OnServiceUpdate(serviceUpdate *watchers.Se
 	}
 }
 
-// sync the ipvs service and server details configured to reflect the desired state of services and endpoint
-// as learned from services and endpoints information from the api server
+// sync the ipvs service and server details configured to reflect the desired
+// state of services and endpoint as learned from services and endpoints
+// information from the api server
 func (nsc *NetworkServicesController) syncIpvsServices(serviceInfoMap serviceInfoMap, endpointsInfoMap endpointsInfoMap) {
 
 	start := time.Now()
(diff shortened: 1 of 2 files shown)


app/controllers/network_services_controller.go
```diff
@@ -375,6 +379,159 @@ func ensureMasqueradeIptablesRule(masqueradeAll bool, podCidr string) error {
...
+
+	// TODO: Factor out this code
+	for _, chain := range chains {
+		if chain == hairpinChain {
+			hasHairpinChain = true
+		}
+	}
+
+	// Delete the chain for hairpin rules, if needed
+	if hasHairpinChain != true {
```
murali-reddy: Should that be hasHairpinChain == true?

app/controllers/network_services_controller.go
```diff
@@ -375,6 +379,159 @@ func ensureMasqueradeIptablesRule(masqueradeAll bool, podCidr string) error {
...
+	iptablesCmdHandler, err := iptables.New()
+	if err != nil {
+		return errors.New("Failed to initialize iptables executor" + err.Error())
+	}
+
+	// TODO: Factor this static jump rule out
+	jumpArgs := []string{"-m", "ipvs", "-j", hairpinChain}
+	err = iptablesCmdHandler.Delete("nat", "POSTROUTING", jumpArgs...)
+	if err != nil {
+		glog.Errorf("Unable to delete hairpin jump rule from chain \"POSTROUTING\": %e", err)
```
murali-reddy: We can return from here. If there is reference from nat POSTROUTING chain to hairpin chain, delete hairpin chain will fail anyways.

app/controllers/network_services_controller.go
```diff
@@ -375,6 +379,159 @@ func ensureMasqueradeIptablesRule(masqueradeAll bool, podCidr string) error {
 	return nil
 }
 
+// syncHairpinIptablesRules adds/removes iptables rules pertaining to traffic
+// from an Endpoint (Pod) to its own service VIP. Rules are only applied if
+// enabled globally via CLI argument or a service has an annotation requesting
+// it.
+func (nsc *NetworkServicesController) syncHairpinIptablesRules() error {
```
murali-reddy: I am assuming we will call this function only if gloabal flag is set or atleast one service is annotated. Right?

PR #61, opened 2017-07-13
Fix logging to stderr

bzub:
The last commit broke logging to stderr.  I'm going to let the image build and see if this fixes it.


bzub:
Works OK now. Merging.

PR #82, opened 2017-07-28
support for network policy GA

### LINKED ISSUE #16: support for network policy GA
very likely network policy is going to be GA in 1.7 [1]

there are some semantic changes not compatible with v1beta1 so existing implementation in kube-router does not work for GA/1.7 network policies. Refactor code so that it works for both GA and v1beta1 semantics of network policies

[1] https://github.com/kubernetes/kubernetes/pull/39164

murali-reddy:
with this refactoring support for network policy V1 (or GA) is added.
Changes are backward compatible so beta network policy semantics
are still available for k8s ver 1.6.* and less

Fixes #16


murali-reddy:
Tested on kubernetes 1.6 and 1.7 versions. Kube-router supports both network policy Beta and GA semantics.

PR #111, opened 2017-08-08
Enable pod egress masquerading by default

### LINKED ISSUE #107: Replace or remove "--cluster-cidr"
Kubernetes conformance tests assume that pods are able to reach the internet by default. I think most Kubernetes users expect this behavior by default as well. We already determine the Pod CIDR at runtime.

So, I think we should:
- Remove `--cluster-cidr`.
- Default to enable SNAT for pod traffic destined outside the cluster.
- Add a new flag to disable the last item, `--disable-pod-egress` or similar.

bzub:
- ~~Removes flag "--cluster-cidr"~~
- Depreciates flag `--cluster-cidr`
- Adds flag `--enable-pod-egress` (default: true)
- Removes previously created iptables rule if option is changed to false
- Uses ipset of PodCIDRs from active nodes for iptables matching

Fixes #107 if workable.

~~@murali-reddy this uses `currentCidr` which is the Pod subnet local to the node, rather than the entire PodCIDR for all pods. Do you see any issues from this? Seems that any particular node shouldn't care about masquerading traffic for another node's Pod subnet, but I may be mistaken.~~

I'm running k8s conformance tests now, will update when complete.

### DIFF
--- a/Documentation/README.md
+++ b/Documentation/README.md
@@ -69,25 +69,27 @@ Also you can choose to run kube-router as agent running on each cluster node. Al
 ### command line options
 
 ```
-  --run-firewall                    If false, kube-router won't setup iptables to provide ingress firewall for pods. true by default.
-  --run-router                      If true each node advertise routes the rest of the nodes and learn the routes for the pods. false by default
-  --run-service-proxy               If false, kube-router won't setup IPVS for services proxy. true by default.
-  --cleanup-config                  If true cleanup iptables rules, ipvs, ipset configuration and exit.
-  --masquerade-all                  SNAT all traffic to cluster IP/node port. False by default
-  --cluster-cidr                    CIDR range of pods in the cluster. If specified external traffic from the pods will be masquraded
-  --config-sync-period duration     How often configuration from the apiserver is refreshed. Must be greater than 0. (default 1m0s)
-  --iptables-sync-period duration   The maximum interval of how often iptables rules are refreshed (e.g. '5s', '1m'). Must be greater than 0. (default 1m0s)
-  --ipvs-sync-period duration       The maximum interval of how often ipvs config is refreshed (e.g. '5s', '1m', '2h22m'). Must be greater than 0. (default 1m0s)
-  --kubeconfig string               Path to kubeconfig file with authorization information (the master location is set by the master flag).
-  --master string                   The address of the Kubernetes API server (overrides any value in kubeconfig)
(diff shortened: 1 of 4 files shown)


murali-reddy:
@bzub conformance tests MAY pass, but it breaks [fundamental assumptions](https://kubernetes.io/docs/concepts/cluster-administration/networking/) from Kubernetes networking perspective.

- all containers can communicate with all other containers without NAT
- the IP that a container sees itself as is the same IP that others see it as
- all nodes can communicate with all containers (and vice-versa) without NAT

Whole network policy semantics need source IP to be preserved.

bzub:
```
[10138.798644] golang[5]: Ran 149 of 652 Specs in 3951.314 seconds
[10138.799520] golang[5]: SUCCESS! -- 149 Passed | 0 Failed | 0 Pending | 503 Skipped PASS
[10138.808301] golang[5]: Ginkgo ran 1 suite in 1h5m51.854043594s
[10138.811313] golang[5]: Test Suite Passed
[10138.828900] golang[5]: 2017/08/08 04:34:04 util.go:133: Step './hack/ginkgo-e2e.sh --ginkgo.focus=\[Conformance\]' finished in 1h5m52.461830863s
[10138.829175] golang[5]: 2017/08/08 04:34:04 e2e.go:80: Done
```

bzub:
You're right, this would masquerade pod traffic between nodes. I'll have to exclude other node's pod cidrs.

bzub:
@murali-reddy This now uses/updates an ipset to match Pod egress traffic to be masqueraded. Local testing looks good. Verified pod-to-pod does not get masqueraded between nodes or any time. Verified `--enable-pod-egress=false` deletes previously created iptables rule.

Running conformance test now.

murali-reddy:
@bzub thats nice. One caveat though. Kops has [minifest](https://github.com/kubernetes/kops/blob/master/upup/models/cloudup/resources/addons/networking.kuberouter/k8s-1.6.yaml.template#L70) with rolled in to the binary. I need to get rid of that. Till then can you please just keep the flag. It will not have any effect, but to ensure corba library will not fail.

bzub:
OK good point. I could just depreciate the flag. Good to get interface changes done while we're in alpha.

murali-reddy:
Thanks @bzub. i will raise PR in kops, once we do next release.

murali-reddy:
LGTM

PR #133, opened 2017-08-20
Consolidate multiple stop channels into one

Thoro:
Removes the multiple stop channels. The same effect can be reached by a single stop channel that is closed. 

This removes complexity.


murali-reddy:
LGTM

PR #134, opened 2017-08-20
Add build-image folder, which allows to build kube-router on mac os

Thoro:
Because of netlink kube-router can't be built on MacOS.

This adds a folder with a "make.sh" script to build kube router.

Integration into Makefile outstanding.


murali-reddy:
Verified the patch by building on Mac. Thanks @Thoro for the patch.

PR #150, opened 2017-09-07
Kube-proxy like behaviour to listen on all ip's for NodePort service.

### LINKED ISSUE #139: NodePort type service choose NodeIP
Something I want to know, in my node has internal IP, but if I want to reach the NodePort through external Ip, the code return the first internal IP default, how can I  get service through external Ip?

murali-reddy:
NodePort service listens on all ip/interfaces in case of kube-proxy. Currently kube-router listens only on node IP. This fix introduces flag `nodeport-bindon-all-ip` with which you can have kube-proxy like behaviour. If not specified only nodeIP will be open for connections.

Fixes #139


murali-reddy:
When `nodeport-bindon-all-ip` flag is added to the manifest in existing deployment, ipvs is automatically reconfigured to listen on localhost

```
TCP  localhost:31317 rr
  -> 10.1.1.33:http               Masq    1      0          0
  -> 10.1.1.34:http               Masq    1      0          0
  -> 10.1.1.35:http               Masq    1      0          0
```

PR #154, opened 2017-09-08
Adding ability to disable IP-in-IP tunnelining

### LINKED ISSUE #119: User should be able to turn off IPIP tunneling
For cross subnet pod-to-pod connectivity Kube-router currently uses IPIP tunnelining. While having it default enabled is safe bet. User should be able to disable IPIP tunnelining.

If routing infrastructure can route traffic between the pods across the subnets then we dont need IPIP tunneling.

A node sends packet destined for the pod on node in different subnet, to ite default gateway. If default gateway can route the traffic then there is no need for tunneling.

murali-reddy:
Adding ability to disable IP-in-IP tunnelining for cross node pod-to-pod connectivity where nodes are in different subnet. With tunneling disabled its expected that
default gateway has learned the pod CIDR's allocated for all the nodes and can route
the pod-to-pod traffic across nodes in different subnets

Fixes #119

### DIFF
--- a/app/controllers/network_routes_controller.go
+++ b/app/controllers/network_routes_controller.go
@@ -54,6 +54,7 @@ type NetworkRoutingController struct {
 	globalPeerAsnNumber  uint32
 	bgpFullMeshMode      bool
 	podSubnetsIpSet      *ipset.IPSet
+	enableOverlays       bool
 }
 
 var (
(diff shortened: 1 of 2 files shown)


murali-reddy:
Have verified deploying below cluster where nodes in us-west-2a and nodes in us-west-2b are in different subnets
```
kops create cluster \
  --node-count 3 \
  --zones us-west-2a,us-west-2b \
  --master-zones us-west-2a \
  --dns-zone aws.cloudnativelabs.net \
  --node-size m4.large \
  --master-size m4.large \
  --networking kube-router \
  --yes \
  --kubernetes-version 1.7.3 \
  --name mycluster.aws.cloudnativelabs.net
```

I have shuffled `--enable-overlay` to verify tunnels get created and cleaned up as the flas is set to `true` and `false` respectively.

PR #156, opened 2017-09-09
Dynamic gobgpd listener config, fix leaked listener

### LINKED ISSUE #155: BGP server fails to start if ipv6 is disabled via ipv6.disable Kernel parameter
When IPv6 is disabled using Kernel parameter "ipv6.disable=1" kube-router fails silently. The node gobgp server errors that the "address family not supported by protocol" but on restart attempt finds 0.0.0.0:179 is bound / already in use:
```
E0908 14:36:41.037123       1 network_routes_controller.go:144] Failed to start node BGP server: Failed to start BGP server due to : listen tcp6 [::]:179: socket: address family not supported by protocol
I0908 14:37:40.824288       1 network_routes_controller.go:150] Retrying start of node BGP server
E0908 14:37:40.827157       1 network_routes_controller.go:144] Failed to start node BGP server: Failed to start BGP server due to : listen tcp4 0.0.0.0:179: bind: address already in use
```
It would be ideal if we could customize the listen interface(s) and supported protocol for the BGP server.

bzub:
partially addresses #155 

This will automatically detect address families that are supported to avoid errors when ipv4 or ipv6 are disabled.  A followup PR will make the address customizable via CLI flags.

### DIFF
--- a/app/controllers/network_routes_controller.go
+++ b/app/controllers/network_routes_controller.go
@@ -69,7 +69,7 @@ const (
 	podSubnetIpSetName   = "kube-router-pod-subnets"
 )
 
-// Run runs forever till until we are notified on stop channel
+// Run runs forever until we are notified on stop channel
 func (nrc *NetworkRoutingController) Run(stopCh <-chan struct{}, wg *sync.WaitGroup) {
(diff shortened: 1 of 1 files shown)


bzub:
Will merge once e2e tests pass.

Issue #163, opened 2017-09-15
Support MD5 authentication for external and node BGP peering


bzub:
- [x] Add flag to specify external router's MD5 password #164 
- [x] Add node annotation to specify external router's MD5 password #164 
- [ ] Add flag to specify full-mesh node-to-node MD5 password

aauren:
Closing as stale

PR #164, opened 2017-09-16
Add flag to specify external router's MD5 password

bzub:
Addresses the following item for #163
- Add flag to specify external router's MD5 password

### Other changes, please review.
- **BREAKING**: `--peer-router` and `--peer-asn` flags now take a slice rather than a comma separated string. In hindsight the previous type (comma separated string) may be easier to maintain consistency with annotations.
- Consolidated NRC peer fields into a []config.NeighborConfig
  to store address, asn, and password for each peer.
- Use net.IP type directly in flags. pflags library will validate the input so we don't have to.

### DIFF
--- a/Documentation/README.md
+++ b/Documentation/README.md
@@ -97,28 +97,30 @@ Also you can choose to run kube-router as agent running on each cluster node. Al
 
 ```
 Usage of ./kube-router:
-    --advertise-cluster-ip            Add Cluster IP to the RIB and advertise to peers.
-    --cleanup-config                  Cleanup iptables rules, ipvs, ipset configuration and exit.
-    --cluster-asn string              ASN number under which cluster nodes will run iBGP.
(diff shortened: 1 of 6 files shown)


bzub:
The user interface for BGP peer config gets a little bit awkward if peering with more than one external router + passwords. We might want to consider defining a CustomResourceDefinition for BGP configurations, so that we have native JSON/YAML types to work with and so that users can modify the config with kubectl.

murali-reddy:
@bzub can you update BGP documentation as well

https://github.com/cloudnativelabs/kube-router/blob/master/Documentation/bgp.md

murali-reddy:
LGTM, plase update the BGP documentation, and documentation on [command line options](https://github.com/cloudnativelabs/kube-router/tree/master/Documentation#command-line-options)

bzub:
I went ahead and made the CLI flags and annotations consistent with one another. Needed to be done before beta. The CLI BGP related flags still take single items and comma separated lists :).

murali-reddy:
👍 LGTM

PR #169, opened 2017-09-25
Delete ipset dependency + delete unused ipsets

ryarnyah:
I add some improvement to remove old ipset set used by old networkpolicies.

* Remove ipset dependency to use thread safe ipset
* Delete unused ipset sets on clean

Feel free to comment or suggest any modification to this PR.

Edit: Seems that travis can't push to docker.io....


murali-reddy:
@ryarnyah thanks for your PR. Yes we have pending issue to fix Travis CI. Please ignore that.

We will review, and get back to you.

murali-reddy:
@ryarnyah Overall looks good. Somehow cleaning up ipsets slipped all this while, there is no tracking bug as well :) So thanks for fixing it.

I understand its just a wrapper around ipset command, so its okay if it needed to be replaced with what you wrote. But i am not sure how the new implementation is thread-safe?

ryarnyah:
@murali-reddy The main differencies between the go-ipset implementation and this is that "ipsetPath" variable is initialized per IPSet object and not by globally initialized by the first call to New (which in go-ipset always create a new set).

You can call this implementation more "thread-safe" because IPSet object carry is own "ipsetPath" and init it in each IPSet objects and not shared with all threads and not initialized by only one(which can generate an error with call to Swap per example).

Also this implementation support Save and Restore to Sync with real ipset sets.

Is it more clear? (i'm sorry if my english is not as clear as i want...)

PR #174, opened 2017-10-04
Prevent masquerading pod -> NodeIP traffic

### LINKED ISSUE #172: Pod traffic to remote nodeIP should not use the tunnel interface
With default settings and nodes in different subnets, ipip tunneling is used between nodes and pod CIDR egress is masqueraded. This works for most scenarios however a special case is when a pod accesses an external facing address of another node. I believe this is where problems start (communication hangs after reaching the target node).

### Example
Node A (controller-01) has
- External IP: 147.75.95.177/31
- Pod CIDR: 10.2.0.0/24

Node B (worker-01)has
- External IP: 147.75.95.175/31
- Pod CIDR: 10.2.1.0/24

I start a pod on Node B, and try to ping Node A's external interface:
```
$ k run alpine --rm -ti --image=alpine ping 147.75.95.177
# hangs
```

I see pings coming in to Node A on the tunnel interface, but no reply:
```
core@controller-01 ~ $ sudo tcpdump -vnni tun-1477595175 icmp 
dropped privs to tcpdump
tcpdump: listening on tun-1477595175, link-type RAW (Raw IP), capture size 262144 bytes
13:07:42.696948 IP (tos 0x0, ttl 63, id 39743, offset 0, flags [DF], proto ICMP (1), length 84)
    147.75.95.175 > 147.75.95.177: ICMP echo request, id 2048, seq 91, length 64
13:07:43.697104 IP (tos 0x0, ttl 63, id 40620, offset 0, flags [DF], proto ICMP (1), length 84)
    147.75.95.175 > 147.75.95.177: ICMP echo request, id 2048, seq 92, length 64
13:07:44.697249 IP (tos 0x0, ttl 63, id 41481, offset 0, flags [DF], proto ICMP (1), length 84)
    147.75.95.175 > 147.75.95.177: ICMP echo request, id 2048, seq 93, length 64
```

The source IP was masqueraded from the Pod IP to Node B's external IP: `147.75.95.175 > 147.75.95.177`

Now, on Node B, if I turn off masquerading just for traffic going to Node A, source Pod IP is retained, still going over the tunnel, and replies are received successfully.

On Node B:
```
core@worker-01 ~ $ sudo iptables -t nat -I POSTROUTING -m set --match-set kube-router-pod-subnets src -d 147.75.95.177 -j RETURN
```

On Node A:
```
core@controller-01 ~ $ sudo tcpdump -vnni tun-1477595175 icmp
dropped privs to tcpdump
tcpdump: listening on tun-1477595175, link-type RAW (Raw IP), capture size 262144 bytes
13:14:07.183842 IP (tos 0x0, ttl 63, id 63653, offset 0, flags [DF], proto ICMP (1), length 84)
    10.2.1.4 > 147.75.95.177: ICMP echo request, id 2048, seq 8, length 64
13:14:07.183885 IP (tos 0x0, ttl 64, id 38404, offset 0, flags [none], proto ICMP (1), length 84)
    147.75.95.177 > 10.2.1.4: ICMP echo reply, id 2048, seq 8, length 64
```

### LINKED ISSUE #173: Pod traffic to any node should not have NAT applied
Currently there is a bug with the pod egress masquerading feature which causes traffic from a pod to another node across networks to be masqueraded as well. Instead, for these specific cases Pod source IP should not be changed, while still applying masquerading to egress traffic outside the cluster.

Related to #172

bzub:
Fixes #172 
Fixes #173

### DIFF
--- a/app/controllers/network_policy_controller.go
+++ b/app/controllers/network_policy_controller.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"os/exec"
 	"strconv"
 	"strings"
 	"sync"
@@ -20,7 +19,6 @@ import (
 	"github.com/golang/glog"
 	"k8s.io/client-go/kubernetes"
 	api "k8s.io/client-go/pkg/api/v1"
-	apiv1 "k8s.io/client-go/pkg/api/v1"
 	apiextensions "k8s.io/client-go/pkg/apis/extensions/v1beta1"
 	networking "k8s.io/client-go/pkg/apis/networking/v1"
 )
@@ -167,11 +165,6 @@ func (npc *NetworkPolicyController) Sync() error {
 	var err error
 	npc.mu.Lock()
 	defer npc.mu.Unlock()
-
-	_, err = exec.LookPath("ipset")
-	if err != nil {
-		return errors.New("Ensure ipset package is installed: " + err.Error())
-	}
 
 	start := time.Now()
 	defer func() {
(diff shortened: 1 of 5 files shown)


bzub:
Appears to work well, and automatically deletes the problematic masquerading rule upon kube-router upgrade.

Running e2e tests now.

ryarnyah:
@bzub Why add go-ipset when utils/ipset do the same thing?

bzub:
@ryarnyah I hit some bugs and I had some trouble understanding the workflow with IPSet/Set/Entry. I created #176 to get your help.

PR #177, opened 2017-10-05
Cleanup ipset create function + restore

### LINKED ISSUE #176: utils: ipset issues
@ryarnyah Could you take a look at these items and tell me if there's a bug or if I'm doing something wrong:

https://github.com/cloudnativelabs/kube-router/blob/2ac68610995668c43c99ccf78b1d0122a9406393/utils/ipset.go#L169
IPSet.Create() adds a Set with a nil pointer, should point to the IPSet object.

https://github.com/cloudnativelabs/kube-router/blob/2ac68610995668c43c99ccf78b1d0122a9406393/utils/ipset.go#L289
The other issue I saw was trying to use Set.Refresh(), the command line would usually be malformed under the call to Restore() linked above. It would end up with Go symbols in the command line like `'[hash:net` as the type argument. I believe the issue is with the `%v` printing Go representations of slices in this function: https://github.com/cloudnativelabs/kube-router/blob/2ac68610995668c43c99ccf78b1d0122a9406393/utils/ipset.go#L268

Could you also help me understand the workflow for using Save() and Restore() methods? The comments seem more for CLI usage than Go. Thanks for your help!

ryarnyah:
Clean up my mess:
* Add ipset pointer to Create function
* Use strings.Join to append options
* Add some doc to Save and restore functions
* Switch array to map to simplify set use

Sorry for the inconvenience.

Fix #176


bzub:
An image with these changes is available with `cloudnativelabs/kube-router-git:PR177` for testing.

I believe the ipset issue is the cause of #180. So I'm going to test against that issue.

utils/ipset.go
```diff
@@ -170,8 +170,9 @@ func (ipSet *IPSet) Create(setName string, createOptions ...string) (*Set, error
 	set := &Set{
 		Name:    setName,
 		Options: createOptions,
+		IPSet:   ipSet,
```
bzub: map needs to be initialized before use: `ipSet.Sets = make(map[string]*Set)`

utils/ipset.go
```diff
@@ -170,8 +170,9 @@ func (ipSet *IPSet) Create(setName string, createOptions ...string) (*Set, error
...
 	}
-	ipSet.Sets = append(ipSet.Sets, set)
+	ipSet.Sets[setName] = set
```
bzub: Here's where kube-router will panic using Create() with a nil map

utils/ipset.go
ryarnyah: Thanks for the review. Done it in NewIPSet function.

murali-reddy:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x126c4fb]

goroutine 240 [running]:
github.com/cloudnativelabs/kube-router/utils.(*IPSet).run(0x0, 0xc42051b568, 0x2, 0x2, 0xc420450960, 0xc42062b128, 0x1bd963d, 0x6)
	/home/kube/go/src/github.com/cloudnativelabs/kube-router/utils/ipset.go:125 +0x9b
github.com/cloudnativelabs/kube-router/utils.(*Set).Flush(0xc4206cc690, 0xc420852d20, 0x19)
	/home/kube/go/src/github.com/cloudnativelabs/kube-router/utils/ipset.go:298 +0x7f
```

I ran into this panic, does the patch cover this as well?

bzub:
`quay.io/cloudnativelabs/kube-router:PR177` image has been updated with latest changes.

utils/ipset.go
bzub: That panic *should* be fixed with this PR now. I am having success with the latest revision.

bzub:
I'm going to merge this to hopefully get master branch back to a stable state. I will run e2e tests to verify next as well, and if needed raise a new issue if there's additional issues.

PR #212, opened 2017-11-03
Fix typo in doc

sbkoth:


murali-reddy:
thanks @sbkoth  for the patch.

PR #253, opened 2017-12-18
IPVS periodic sync performance improvement

murali-reddy:
During periodic sync of IPVS services there is a check if the required service
already existing in IPVS. For the check the list of currnet IPVS services are
read from IPVS. This is causing performance hit as number of services increases.

With this fix, Kube-router reads once from ipvs and use for further during service sync


iMartyn:
+1 This reduced our load significantly.

Issue #229, opened 2017-11-27
Add Unit Tests


andrewsykim:
I think kube-router can benefit from having a unit test suite. Not sure how easy it may be to add some now, but seems like reasonable task to group the code in a way that's more testable. 

Related: https://github.com/cloudnativelabs/kube-router/issues/28

cc @murali-reddy

murali-reddy:
Indeed it will benefit. We tried to run k8s networking integration tests as part of CI #140 (using packet, dind etc) but it still pending. Atleast unit test can be added.

aauren:
While there are still areas lacking coverage, I think that in the broad spirit of this issue that kube-router has a good amount of unit testing now.

Issue #135, opened 2017-08-20
hook build-image/make.sh into main Makefile


murali-reddy:
as mentioned in #134 kube-router can not be built on Mac. @Thoro submitted patch to build kube-router on Mac. We need to hook that into main Makefile, so that all commands work seamlessly on both Linux and Mac.

roffe:
We can build on Mac fine with BUILD_IN_DOCKER option added in https://github.com/cloudnativelabs/kube-router/pull/377 , Closing the issue, feel free to reopen if needed

PR #261, opened 2017-12-26
package utils unit tests and CI fixes

andrewsykim:
Adds unit tests for `utils` package. @murali-reddy any tips on how we can enable this on travis CI? 

https://github.com/cloudnativelabs/kube-router/issues/229


build-image/Dockerfile
```diff
@@ -1,18 +0,0 @@
-from centos:7
```
andrewsykim: Seemed unnecessary to have this docker file since we can accomplish the same things with `golang` but I may have missed something so glad to undo this if needed.

murali-reddy:
@andrewsykim thanks for adding the unit test. 

> any tips on how we can enable this on travis C

try adding `make test ` to `build/travis-deploy.sh`

build-image/Dockerfile
murali-reddy: `build-image` has script for building the kube-router image from Mac. See #135

This is not hooked into current CI pipeline. So the changes are not related to unit test, so may be keep this PR just for unit test changes?

andrewsykim:
tests passed here https://travis-ci.org/cloudnativelabs/kube-router/jobs/322256904

PR #264, opened 2017-12-28
route controller unit tests

andrewsykim:
Adds unit tests for route controller. 

cc @murali-reddy 

https://github.com/cloudnativelabs/kube-router/issues/229

### DIFF
--- a/app/watchers/services_watcher.go
+++ b/app/watchers/services_watcher.go
@@ -5,11 +5,13 @@ import (
 	"time"
 
 	"github.com/cloudnativelabs/kube-router/utils"
+
 	api "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


app/watchers/services_watcher.go
```diff
@@ -93,7 +95,15 @@ func StartServiceWatcher(clientset *kubernetes.Clientset, resyncPeriod time.Dura
 
 	svcw.clientset = clientset
 	svcw.broadcaster = utils.NewBroadcaster()
-	lw := cache.NewListWatchFromClient(clientset.Core().RESTClient(), "services", metav1.NamespaceAll, fields.Everything())
+	lw := &cache.ListWatch{
```
andrewsykim: Had to switch to this since the fake client does not have an implementation for `clientset.Core().RESTClient()`.

andrewsykim: On a related note, I don't think we should be using global variable for the watchers, but something that can be addressed later.

murali-reddy:
Thanks @andrewsykim i have tested againest real cluster, switch from cache.NewListWatchFromClient to cache.ListWatch did not have any impact.

PR #265, opened 2018-01-02
Unit Tests for Node Update Events

andrewsykim:
Also refactors `activeNodes` to not use a global variable, this makes testing easier as we don't have to clear the map between test cases.


andrewsykim:
cc @murali-reddy

murali-reddy:
Good One. Thanks @andrewsykim

PR #270, opened 2018-01-04
Make kubeadm install docs clearer

barthoda:
These docs didn't read well and possibly led the user to do a `kubeadm reset`.


murali-reddy:
Thanks @barthoda

Issue #273, opened 2018-01-05
IPIP Tunnel Interface not UP due to "add tunnel 'tunl0' failed: File exists"


andrewsykim:
I've noticed that some interfaces for IPIP tunneling were being created, but not set to `UP`. I added some verbose logging and was able to get the output of the `ip tunnel` commands.

```
I0105 20:50:28.279023       1 network_routes_controller.go:791] ip tunnel args: [ip tunnel add tun-138197148207 mode ipip local 138.197.132.93 remote 138.197.148.207 dev eth0]
E0105 20:50:28.281679       1 network_routes_controller.go:793] ip command output: add tunnel "tunl0" failed: File exists
E0105 20:50:28.281705       1 network_routes_controller.go:342] Failed to inject routes due to: Route not injected for the route advertised by the node 138.197.148.207 Failed to create tunnel interface tun-138197148207. error: exit status 1
```
It seems if `tunl0` is already created, we return error since `ip tunnel` returns exit status 1. The early return happens [here](https://github.com/cloudnativelabs/kube-router/blob/master/app/controllers/network_routes_controller.go#L791-L793). This leads to nodes having all the tunnel interfaces configured, but running `netlink.LinkSetUp(link)`. 

Not familiar enough with with IPIP Tunneling to comment as to how `tunl0` is related since the IP command was `ip tunnel add tun-138197148207 mode ipip local 138.197.132.93 remote 138.197.148.207 dev eth0`. The same error happens if run the command locally.
```
# ip tunnel add tun-138197148207 mode ipip local 138.197.132.93 remote 138.197.148.207 dev eth0
add tunnel "tunl0" failed: File exists
```

andrewsykim:
cc @murali-reddy

andrewsykim:
After further investigation, I found out that it's actually `netlink.LinkByName` that's failing with:
```
I0105 21:20:23.209656       1 network_routes_controller.go:786] error getting link by name: numerical result out of range
```

andrewsykim:
It seems netlink has a 15 character limit, which would explain why only _some_ of my routes are not available.

murali-reddy:
Fixed with #274

PR #274, opened 2018-01-05
tunnel interface names cannot be longer than 15 characters

andrewsykim:
Ran into some strange behaviour as reported [here](https://github.com/cloudnativelabs/kube-router/issues/273). Turns out it was because linux interface names cannot be more than 15 characters. Removing the `-` in the interface name ensures that we will always have at most 15 characters for IPv4. However, I do have concerns around backwards compatibility and some more work may have to be done here so that the transition is smoother. @murali-reddy would love your thoughts here.

### DIFF
--- a/app/controllers/network_routes_controller.go
+++ b/app/controllers/network_routes_controller.go
@@ -760,7 +760,7 @@ func (nrc *NetworkRoutingController) injectRoute(path *table.Path) error {
 
 	// check if the neighbour is in same subnet
 	if !nrc.nodeSubnet.Contains(nexthop) {
-		tunnelName := "tun-" + strings.Replace(nexthop.String(), ".", "", -1)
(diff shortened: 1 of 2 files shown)


andrewsykim:
On second thought, if I had used a private IPs (10.X.X.X) I probably wouldn't have run into this issue. However, this is an edge case that other uses _may_ run into.

murali-reddy:
@andrewsykim problem may arise with private ip range of 172.16.*.*-172.31.*.* as well. So it must be fixed. But i know some production users who are using old convention. Modifing the logic to generate the tunnel interface name with your change only if length of IP address without `.` is 12?

andrewsykim:
@murali-reddy that's a great idea, I'll update the PR soon!

andrewsykim:
updated

murali-reddy:
Thanks @andrewsykim

Issue #262, opened 2017-12-27
advertise service VIP from the node only if there is atleast one service endpoint pod on the node


murali-reddy:
Opening an issue for the slack conversation:

https://kubernetes.slack.com/archives/C8DCQGTSB/p1513691858000406

Advertise service VIP (cluster ip, external ip and loadbalancer ip when supported in kube-router) from the node only if there is atleast one service endpoint pod on the node.

this will nicely complement #254. While #254 helps in preserving the client IP which is very essential when providing L4 ingress, this issue will prevent extra hop.

andrewsykim:
Going to take a stab at writing some tests for routes controller before I tackle this, if anyone else is interested on this issue, please go for it :)

wrigtim:
+1 from me. Useful when you have some applications that are less friendly with their traffic profile (e.g. exceptionally high-rate UDP).

murali-reddy:
@andrewsykim its already addressed right?

andrewsykim:
Yup, you have to set service.Spec.ExternalTrafficPolicy=Local as per https://github.com/cloudnativelabs/kube-router/pull/334. We'll also want https://github.com/cloudnativelabs/kube-router/pull/350 implemented so we have faster convergence for BGP routes

murali-reddy:
Thanks for confirming. Closing the bug.

@wrigtim kube-router already has the support for `ExternalTrafficPolicy=Local`, with extra smarts to advertise from the node only if there is atleast on service endpoint.

wrigtim:
@murali-reddy ...but can’t do external policy with ClusterIP? There’s still an extra hop we can take out here?

murali-reddy:
@wrigtim some how documentation is not in sync with changes went in for this feature. I will add documentation. 

You could add kube-router specific annotation to any service `kube-router.io/service.local=true`. Which will result in same behaviour as `externalTrafficPolicy=Local` irrespective of service type.

wrigtim:
Ah fantastic. Appreciate the rapid response - will test it out...

murali-reddy:
@jdconti pointed out missing functionality so reopening the issue.

https://kubernetes.slack.com/archives/C8DCQGTSB/p1523553155000833

uablrek:
Please make this "feature" optional. If you always have *one* service for each external VIP it's fine. Problem is for applications that exposes multiple services with different port but with the same external VIP, e.g. http and sip (just a made-up example).

[...]

PR #277, opened 2018-01-09
support node local service without SNAT

GuoyouZhang:
use case: in VNF, traffic come from external client, and go through router ECMP to k8s node.  The pod needs to know the external client IP.  To meet this case:
1. change iptables rule to avoid SNAT for traffic to the pod on local node
2.  when annotation "kube-router.io/service.local" configured on service, only add local pod as VIP back-end. 
Finally, kube-router will advertise external IP for the node, which really has the local pod and can serve the external request without SNAT.

### DIFF
--- a/app/controllers/network_services_controller.go
+++ b/app/controllers/network_services_controller.go
@@ -99,6 +99,7 @@ type serviceInfo struct {
 	directServerReturnMethod string
 	hairpin                  bool
 	externalIPs              []string
+	local                    bool
 }
 
 // map of all services, with unique service id(namespace name, service name, port) as key
@@ -422,7 +423,7 @@ func (nsc *NetworkServicesController) syncIpvsServices(serviceInfoMap serviceInf
 				Weight:        1,
 			}
 
-			err := ipvsAddServer(ipvsClusterVipSvc, &dst)
+			err := ipvsAddServer(ipvsClusterVipSvc, &dst, false, nsc.podCidr)
 			if err != nil {
 				glog.Errorf(err.Error())
 			}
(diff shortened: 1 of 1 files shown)


murali-reddy:
Excellent @GuoyouZhang, changes look good to me. I will test out briefly and merge.

andrewsykim:
> Finally, kube-router will advertise external IP for the node, which really has the local pod and can serve the external request without SNAT.

I don't see the necessary changes to ensure only services running the service pods will advertise external IPs, am I missing something?

murali-reddy:
@andrewsykim It does not address #262 

@GuoyouZhang annotating service after service is deployed does not work. i will open seperate issue for that.

GuoyouZhang:
@andrewsykim 
if there is no pod on local Node, traffic will be dropped since we ask ipvs only add local pod as backend by annotation service.local.

KJSui:
@GuoyouZhang how to enable the service? Would you like to provide cmd? Thanks.

murali-reddy:
@KJSui please try adding `kube-router.io/service.local=true` annotaion in the service definition. it does not work dynamically (i.e annotating after service has been created) yet (it needs to be fixed). so add the annotation in the service defintion.

KJSui:
@murali-reddy @GuoyouZhang After enabling this service as you suggest, if I curl the service, the client ip doesn't be preserved... If I curl the pod ip directly, I did see client ip appear, do you guys have any idea why the client ip doesn't appear if I curl the external ip of service?

KJSui:
@murali-reddy the cmd should be kube-router.io/service.local="" not "true" after I tried many times.

andrewsykim:
@GuoyouZhang @murali-reddy Another consideration I think we missed in this PR is that the iptable rules removing SNAT are being appended instead of replaced. So instead of seeing only 
```
-A POSTROUTING ! -s 10.38.37.0/26 ! -d 10.38.37.0/26 -m ipvs --vdir ORIGINAL --vmethod MASQ -m comment --comment "" -j MASQUERADE
```

we saw 
```
-A POSTROUTING ! -s <pod-cidr> ! -d <pod-cidr> -m ipvs --vdir ORIGINAL --vmethod MASQ -m comment --comment "" -j MASQUERADE
-A POSTROUTING ! -s <pod-cidr> -m ipvs --vdir ORIGINAL --vmethod MASQ -m comment --comment "" -j MASQUERADE
```

and I think since the previous rule is first in the chain it takes precedence?

murali-reddy:
@andrewsykim i think your observation is right. I will fix it.

PR #281, opened 2018-01-11
Documentation how to install kube-router on generic clusters

roffe:
Docs and examples how to run kube-router as a DaemonSet in a generic cluster


murali-reddy:
Thanks @roffe. Very useful.

PR #285, opened 2018-01-14
create ipip tunnel interface regardless of node subnet

andrewsykim:
Always creating a tunnel interface for an overlay, regardless of its subnets seems like the correct behaviour to me.  This change _might_ be incompatible with some networking topology, e.g. if a global BGP peer in the same "subnet" advertises a default route, this change would mean cluster admins need to configure the default route themselves now.

### DIFF
--- a/app/controllers/network_routes_controller.go
+++ b/app/controllers/network_routes_controller.go
@@ -752,86 +752,78 @@ func (nrc *NetworkRoutingController) addExportPolicies() error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


murali-reddy:
@andrewsykim sorry i don't quite follow. Would you mind articulating the problem and solution in an issue please?

andrewsykim:
Can do, I'm not sure how feasible this PR is due to backwards compatibiliy, I will close it for now and open an issue.

PR #288, opened 2018-01-17
Added cli config for changing prometheus port & path

roffe:
- Updated readme
- Added option to change prometheus port
- Added option to change prometheus path


murali-reddy:
thanks @roffe

Issue #271, opened 2018-01-04
Support to change port for Promethues Metrics


rockwyc992:
`kube-router` will use 8080 port for Promethues Metrics, but it use [hard code](https://github.com/cloudnativelabs/kube-router/blob/master/app/controllers/network_services_controller.go#L136) to define that.

``` go
// register metrics
prometheus.MustRegister(serviceActiveConn)
prometheus.MustRegister(servicePpsIn)
prometheus.MustRegister(servicePpsOut)
http.Handle("/metrics", promhttp.Handler())
go http.ListenAndServe(":8080", nil)
```

It should move this design to somewhere configurable.

xanonid:
If possible, there should be also annotations added to the pods such that prometheus can auto-discover the metrics (cf. https://github.com/prometheus/prometheus/blob/ec94df49d45baa8e680d81e732294eb04fff00de/documentation/examples/prometheus-kubernetes.yml#L244 )

murali-reddy:
Thanks @xanonid for your comments. requirement make sense. we will try to address it in the coming releases.

roffe:
Submitted a PR for this that also adds the option to change the path

https://github.com/cloudnativelabs/kube-router/pull/288

roffe:
PR #288 has been merged and we should be able to close this issue.

@xanonid  I have added docs for what you are asking in https://github.com/cloudnativelabs/kube-router/pull/289 comments welcome!

roffe:
This issue can be closed

Issue #240, opened 2017-12-11
Support log level output in configuration


bzub:
We try to make use of basic log levels in kube-router (INFO, WARN, ERROR), however there is no way to filter out purely informational logs, for example. Users should be able to start kube-router with a verbosity or log level setting.

There are even more sophisticated ways to implemenet this too, such as logging to different streams. Suggestions are welcome on the most practical approach.

roffe:
Since we are using [glog](https://godoc.org/github.com/golang/glog) package i suggest we start using it's levels and agree on a common level system to be used when doing code

different -v= levels:
- 0 default, print info messages such as "started sync loop for services"
- 1 additional info, such as printing each service it's syncinc in the sync service loop e.g "syncing service XYZ"
- 2 detailed info, print out things as protocol, adresses etc in more detail. 
- 3 debug print things such as every event processed with as much details as possible.

the code to use the different levels is realy neat and easy to incorporate in existing code.

roffe:
With #294 I changed log level on a lot of the kube-router output so by default it won't be that talky.

It will obey --v or -v flag to produce more verbose output

Imho we can close this issue now

Issue #5, opened 2017-05-13
Expose node as prometheus endpoint


murali-reddy:
Lots of inforamtion is readily available (Information from conntrack, ipvs, iptables dropped packets etc) to Kube-router that can be exposed as prometheus metric. 

Add capability to present metrics as  prometheus endpoint.

murali-reddy:
@bzub Its fairly easy task. There are plenty of examples out there. Some thing that is close to kube-router:

https://github.com/kobolog/gorb/blob/master/core/prometheus_exporter.go

I will start workign on this as well. We shall coordinate and get this out.

murali-reddy:
metrics exposed as http://nodeip:8080/metrics

```
# HELP kube_router_service_backend_active_connections Active conntection to backend of service
# TYPE kube_router_service_backend_active_connections gauge
kube_router_service_backend_active_connections{backend="100.96.1.2",namespace="kube-system",service_name="kube-dns"} 0
kube_router_service_backend_active_connections{backend="100.96.1.3",namespace="default",service_name="redis-slave"} 0
kube_router_service_backend_active_connections{backend="100.96.1.4",namespace="default",service_name="frontend"} 0
kube_router_service_backend_active_connections{backend="100.96.1.5",namespace="default",service_name="frontend"} 0
kube_router_service_backend_active_connections{backend="100.96.2.3",namespace="kube-system",service_name="kube-dns"} 0
kube_router_service_backend_active_connections{backend="100.96.2.4",namespace="default",service_name="redis-master"} 1
kube_router_service_backend_active_connections{backend="100.96.2.5",namespace="default",service_name="redis-slave"} 0
kube_router_service_backend_active_connections{backend="100.96.2.6",namespace="default",service_name="frontend"} 0
kube_router_service_backend_active_connections{backend="172.20.51.157",namespace="default",service_name="kubernetes"} 1
# HELP kube_router_service_backend_inactive_connections Active conntection to backend of service
# TYPE kube_router_service_backend_inactive_connections gauge
kube_router_service_backend_inactive_connections{backend="100.96.1.2",namespace="kube-system",service_name="kube-dns"} 0
kube_router_service_backend_inactive_connections{backend="100.96.1.3",namespace="default",service_name="redis-slave"} 0
kube_router_service_backend_inactive_connections{backend="100.96.1.4",namespace="default",service_name="frontend"} 0
[...]

PR #293, opened 2018-01-22
Metrics + Logging update

roffe:
PR addresses:

https://github.com/cloudnativelabs/kube-router/issues/240 
Default output of kube-router has been heavily silenced by default.
Output level can be controlled with the glog flag --v

By default kube-router will only output messages such as start & stop messages

Level 1:
Component info, such as "starting ipvs sync loop" etc.

Level 2
Detailed info on component level, such as: "Adding backend to service" or "Received API update"

Level 3
No output on this level yet.

https://github.com/cloudnativelabs/kube-router/issues/271

Port & Path configuration is in place, as well as the option to turn metrics on / off by changing the port number.

The default option has been changed so metrics automaticly is not exposed

https://github.com/cloudnativelabs/kube-router/issues/5
All the traffic stats based on the ipvs services is currently being exposed


roffe:
Also a updated .dockerignore file to not send all of the source code to the docker build daemon when it's not needed

roffe:
Have found bugs i need to address

PR #294, opened 2018-01-22
Metrics + Logging update

roffe:
PR addresses:

#240
Default output of kube-router has been heavily silenced by default.
Output level can be controlled with the glog flag -v & --v

By default kube-router will only output messages such as start & stop messages

Level 1:
Component info, such as "starting ipvs sync loop" etc.

Level 2
Detailed info on component level, such as: "Adding backend to service" or "Received API update"

Level 3
No output on this level yet. but things as raw dumps of variable data etc for heavier debugging should be placed in the higher levels (5+)

#271

Port & Path configuration is in place, as well as the option to turn metrics on / off by changing the port number.

The default option has been changed so metrics automaticly is not exposed

#5
All the traffic stats based on the ipvs services is currently being exposed


roffe:
From earlier kube-router behaviour the config & startup behaviour is the same except for metrics which will NOT be exported unless the user explicitly sets a port, the default is 0. not exposing metrics.

murali-reddy:
@roffe thanks for your work. Is it possible to honour these flags. Some users may be using kube-proxy with kube-router.

Overall looks to good.

```
--run-firewall                        Enables Network Policy -- sets up iptables to provide ingress firewall for pods. (default true)
--run-router                          Enables Pod Networking -- Advertises and learns the routes to Pods via iBGP. (default true)
--run-service-proxy                   Enables Service Proxy -- sets up IPVS for Kubernetes Services. (default true)```
```

roffe:
Thanks for the feedback @murali-reddy however I'm not completely sure what you mean. Those flags should work just as before and my PR should not change the behaviour of those

roffe:
murali.reddy [2:17 PM]
what i am suggesting there will be three buckets of metrics. One for service proxy, one for pod networking, routing stuff etc and last one for network policy.
murali.reddy [2:18 PM]
so if some one is not using kube-router as service proxy, then whole bunch of service metrics are not be published even if kube-router may publish for routing etc.

Makes total sense. Hang on while i implement it

roffe:
@murali-reddy Done.

roffe:
Sorry about the last minute fixes. was observing high cpu usage on kube-router and after some profiling i found that my wait state in metrics_controller was eating a whole cpu core

murali-reddy:
thanks @roffe

PR #295, opened 2018-01-24
Add override next hop option

Thoro:
Added the option to override the nexthop address when peering with multiple peers in different subnets (which are all directly reachable).


murali-reddy:
@Thoro can you please elobrat the commit https://github.com/cloudnativelabs/kube-router/pull/295/commits/a36aab49930e79e2ec7748f8a6ef764c0d5fefeb

I was trying to change the code to use `.conflist` and use `ConfListFromFile` API from CNI. See some overlapping changes, so was wondering.

Thoro:
@murali-reddy I actually just did a quick fix

e.g. 
- remove cni and replace with cni / plugin repos
- remove the setting of the ip in the conf file, and instead replace it with an error message
- use the new ranges api and figure out the first ipv4 addr, and ignore all others

aauren:
This functionality was introduced in #518

PR #303, opened 2018-02-12
support service.spec externalTrafficPolicy=Local

murali-reddy:
support service.spec externalTrafficPolicy=Local. Takes precedence over `kube-router.io/service.local` annotation.

Also dynamically cleanup service endpoints based on the service spec is set for Local only services or not.


murali-reddy:
Image with fix: `cloudnativelabs/kube-router-git:external-traffic-policy`

roffe:
lgtm

murali-reddy:
Thanks @roffe

PR #327, opened 2018-03-04
limit disableSrcDstCheck performed on OnNodeUpdate()

murali-reddy:
OnNodeUpdate dont perform disableSrcDstCheck until first round disableSrcDstCheck() is performed at start

### DIFF
--- a/app/controllers/network_routes_controller.go
+++ b/app/controllers/network_routes_controller.go
@@ -69,6 +69,7 @@ type NetworkRoutingController struct {
 	bgpRRServer          bool
 	bgpClusterId         uint32
 	cniConfFile          string
+	initSrcDstCheckDone  bool
 }
 
 var (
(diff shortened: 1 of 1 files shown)


app/controllers/network_routes_controller.go
```diff
@@ -1276,7 +1278,11 @@ func (nrc *NetworkRoutingController) OnNodeUpdate(nodeUpdate *watchers.NodeUpdat
 	if nrc.bgpEnableInternal {
 		nrc.syncInternalPeers()
 	}
-	nrc.disableSourceDestinationCheck()
+
+	// skip if first round of disableSourceDestinationCheck() is not done yet
+	if nrc.disableSrcDstCheckDone {
```
andrewsykim: Is this supposed to be `if !nrc.disableSrcDstCheckDone`?

murali-reddy: logic is correct, but naming is very poor. I will correct it.

Basically when kube-rotuer pod starts, master nodes try to disable src-dst check for all the nodes just one time.

Until its done, dont try to do it again for node updates as well.

i guess whats happening was, you launch cluster with 30-40 nodes for e.g, then for each node update, same task was being done concurrently so ec2 api serves was throttling the requests.

this check prevents unnecessary calls while in the initial (cluster booting) phase.

PR #333, opened 2018-03-10
Test_addExportPolicies

andrewsykim:
* also fixes bug where default policy was temporarily set to `ACCEPT`   

cc @murali-reddy @roffe

### DIFF
--- a/app/controllers/network_routes_controller.go
+++ b/app/controllers/network_routes_controller.go
@@ -753,18 +753,19 @@ func (nrc *NetworkRoutingController) addExportPolicies() error {
 		err = nrc.bgpServer.AddPolicyAssignment("",
 			table.POLICY_DIRECTION_EXPORT,
 			[]*config.PolicyDefinition{&definition},
-			table.ROUTE_TYPE_ACCEPT)
+			table.ROUTE_TYPE_REJECT)
 		if err != nil {
 			return errors.New("Failed to add policy assignment: " + err.Error())
 		}
-	}
-
-	// configure default BGP export policy to reject
-	pd := make([]*config.PolicyDefinition, 0)
-	pd = append(pd, &definition)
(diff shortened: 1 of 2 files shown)


app/controllers/network_routes_controller.go
```diff
@@ -753,18 +753,19 @@ func (nrc *NetworkRoutingController) addExportPolicies() error {
 		err = nrc.bgpServer.AddPolicyAssignment("",
 			table.POLICY_DIRECTION_EXPORT,
 			[]*config.PolicyDefinition{&definition},
-			table.ROUTE_TYPE_ACCEPT)
```
andrewsykim: assumed this was a bug

andrewsykim:
sorry, don't review yet actually, there's some changes I'd like to make :)

andrewsykim:
ready for review now

murali-reddy:
thanks @andrewsykim

So if i understand the changes in `app/controllers/network_routes_controller.go` though in case of `if !policyAssignmentExists {` we were doing `table.ROUTE_TYPE_ACCEPT`, but as it follwed by subsequnet replace with `table.ROUTE_TYPE_REJECT` things worked?

andrewsykim:
That was the impression I had, I can leave it so that `Replace` is always called, to be consistent with other parts of the function.

murali-reddy:
I verfied change in this PR. Export policies are as expected.

- default export policy to reject everything
- explicit accept policy statement to advertise node CIDR and cluster ip's

```
#gobgp global policy export
Export policy:
Default: reject
Name kube_router:
    StatementName kube_router_stmt0:
      Conditions:
        PrefixSet: any podcidrprefixset
      Actions:
         accept
    StatementName kube_router_stmt1:
      Conditions:
        PrefixSet: any clusteripprefixset
        NeighborSet: any externalpeerset
      Actions:
         accept
```

Issue #182, opened 2017-10-06
On service delete, network route controller should proactivley delete routes advertised for  service cluster ip


murali-reddy:
On service delete, network route controller should proactivley delete routes advertised for  service cluster ip. Currently network route controller control loop advertise (add to the GoBGP RIB) the cluster IP for the active services. But it does not watch for service deletion and proactively delete the advertised route, which it should do.

murali-reddy:
#334 will fix from kube-router code perspective, but if there are any advertised service VIP's they continue to remain in GoBGP rib? Need to verify if that is the case.

murali-reddy:
Fixed in #436

PR #334, opened 2018-03-10
Route advertisement should account for services using externalTrafficPolicy=Local

andrewsykim:
cc @murali-reddy @roffe 

Something to consider for the future. IETF RFC4786 outlines:
```
Frequent advertisements and withdrawals of individual prefixes in BGP
 are known as flaps. Rapid flapping can lead to CPU exhaustion on
 routers quite remote from the source of the instability, and for this
 reason rapid route oscillations are frequently "dampened", as
 described in [RFC2439].
```

Crashing pods or rolling updates of large deployments can cause excessive route flapping, I think the impact it will largely depend on the setup of global peers and the size of the cluster. My guess is that very few people will run into issues with route flapping.

### DIFF
--- a/app/controllers/network_routes_controller.go
+++ b/app/controllers/network_routes_controller.go
@@ -31,14 +31,16 @@ import (
 	"github.com/osrg/gobgp/table"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/vishvananda/netlink"
+	v1core "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
 )
 
 // NetworkRoutingController is struct to hold necessary information required by controller
 type NetworkRoutingController struct {
 	nodeIP               net.IP
-	nodeHostName         string
+	nodeName             string
 	nodeSubnet           net.IPNet
 	nodeInterface        string
 	activeNodes          map[string]bool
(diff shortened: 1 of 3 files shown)


murali-reddy:
@andrewsykim Thanks. Can you please check https://github.com/cloudnativelabs/kube-router/issues/182, does stale service VIP continue to remain, or they get timed out?

murali-reddy:
> Crashing pods or rolling updates of large deployments can cause excessive route flapping, I think the impact it will largely depend on the setup of global peers and the size of the cluster. My guess is that very few people will run into issues with route flapping.

I have no idea how it should be interpreted in the context of data centers or iBGP. 

Any thoughts @SEJeff @jdconti @Thoro

andrewsykim:
With the current PR, stale VIPs for services with externalTrafficPolicy `Local` or `Cluster would be treated the same. Withdrawing a route because the service was deleted vs if a node no longer has a healthy endpoint will have a very different implementation though. I'm happy to investigate further what those would look like, would love to hear what everyone else has to say first :). 

Some thoughts so far though:
* for deleting a service, what we currently have today is _probably_ fine since nodes receiving the packet would drop them anyways. We also stop advertising the route once the service is deleting which means the route is eventually removed on peers. 
* withdrawing a route from a service because it no longer has any endpoints for the service (rolling update, pod failed, etc) requires more work because the routes controllers needs to know whether the route was advertised in the first place. If the bpgServer `DeletePath` call is not expensive and idempotent we might be able to get away with calling `DeletePath` everytime we check a local service that has no endpoints.

Thoro:
I believe you should anyways use route aggregation for the service VIPS and therefore you probably can disable route dampening on your edge / dc router

The /32 should never be advertised to external peers anyways (/24 is the minimum for IPv4 and /48 for IPv6 I believe)

murali-reddy:
> With the current PR, stale VIPs for services with externalTrafficPolicy Local or Cluster would be treated the same. Withdrawing a route because the service was deleted vs if a node no longer has a healthy endpoint will have a very different implementation though.

Sorry you are right. I mixed up both.

andrewsykim:
> I believe you should anyways use route aggregation for the service VIPS and therefore you probably can disable route dampening on your edge / dc router
The /32 should never be advertised to external peers anyways (/24 is the minimum for IPv4 and /48 for IPv6 I believe)

I don't think we aggregate routes for cluster/external IPs on kube-router, in other words we are advertising /32s to external and internal peers.

murali-reddy:
> I don't think we aggregate routes for cluster/external IPs on kube-router, in other words we are advertising /32s to external and internal peers.

Yes Kube-router does advertise /32. Will check how big of problem (or practical limitation) it is. AFAIK Google's Maglev, Microsoft Ananta etc does VIP (/32) advertisement from all the nodes.

PR #335, opened 2018-03-11
[WIP] support for Hostport

murali-reddy:
This PR addes support for hostport.

- CNI conf file to be used is moved to env variable. If env var is not present falls back to previous hard coded value i.e) `/etc/cni/net.d/10-kuberouter.conf`
- support for `.conflist` is added so you can specify multiple plug-ins
- sample [daemonset](https://github.com/murali-reddy/kube-router/blob/hostport/daemonset/kubeadm-kuberouter-all-features-hostport.yaml) is added with 
-- env variable set to `/etc/cni/net.d/10-kuberouter.conflist`
-- configmap with cni conf json with portmap plugin config
-- init container changes to add `.conflist` file
   

Image `cloudnativelabs/kube-router-git:hostport` has the changes in this PR.

### DIFF
--- a/app/controllers/network_routes_controller.go
+++ b/app/controllers/network_routes_controller.go
@@ -66,6 +66,7 @@ type NetworkRoutingController struct {
 	bgpRRClient          bool
 	bgpRRServer          bool
 	bgpClusterId         uint32
+	cniConfFile          string
 }
 
 var (
@@ -87,7 +88,7 @@ const (
 
 // Run runs forever until we are notified on stop channel
 func (nrc *NetworkRoutingController) Run(healthChan chan<- *ControllerHeartbeat, stopCh <-chan struct{}, wg *sync.WaitGroup) {
-	cidr, err := utils.GetPodCidrFromCniSpec("/etc/cni/net.d/10-kuberouter.conf")
+	cidr, err := utils.GetPodCidrFromCniSpec(nrc.cniConfFile)
 	if err != nil {
 		glog.Errorf("Failed to get pod CIDR from CNI conf file: %s", err.Error())
 	}
@@ -100,7 +101,7 @@ func (nrc *NetworkRoutingController) Run(healthChan chan<- *ControllerHeartbeat,
 	}
 
 	if len(cidr.IP) == 0 || strings.Compare(oldCidr, currentCidr) != 0 {
-		err = utils.InsertPodCidrInCniSpec("/etc/cni/net.d/10-kuberouter.conf", currentCidr)
+		err = utils.InsertPodCidrInCniSpec(nrc.cniConfFile, currentCidr)
 		if err != nil {
 			glog.Errorf("Failed to insert pod CIDR into CNI conf file: %s", err.Error())
 		}
@@ -1501,6 +1502,14 @@ func NewNetworkRoutingController(clientset *kubernetes.Clientset,
 	nrc.bgpRRClient = false
 	nrc.bgpRRServer = false
 	nrc.bgpServerStarted = false
+
+	nrc.cniConfFile = os.Getenv("KUBE_ROUTER_CNI_CONF_FILE")
+	if nrc.cniConfFile == "" {
+		nrc.cniConfFile = "/etc/cni/net.d/10-kuberouter.conf"
+	}
+	if _, err := os.Stat(nrc.cniConfFile); os.IsNotExist(err) {
+		return nil, errors.New("CNI conf file " + nrc.cniConfFile + " does not exist.")
+	}
 
 	nrc.ipSetHandler, err = utils.NewIPSet()
 	if err != nil {
--- a/daemonset/kubeadm-kuberouter-all-features-hostport.yaml
+++ b/daemonset/kubeadm-kuberouter-all-features-hostport.yaml
@@ -1,0 +1,184 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kube-router-cfg
+  namespace: kube-system
+  labels:
+    tier: node
+    k8s-app: kube-router
+data:
+  cni-conf.json: |
+    {
+       "cniVersion":"0.3.0",
+       "name":"mynet",
+       "plugins":[
+          {
+             "name":"kubernetes",
+             "type":"bridge",
+             "bridge":"kube-bridge",
+             "isDefaultGateway":true,
+             "ipam":{
+                "type":"host-local"
+             }
+          },
+          {
+             "type":"portmap",
+             "capabilities":{
+                "snat":true,
+                "portMappings":true
+             }
+          }
+       ]
+    }
+---
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  labels:
+    k8s-app: kube-router
+    tier: node
+  name: kube-router
+  namespace: kube-system
+spec:
+  template:
+    metadata:
+      labels:
+        k8s-app: kube-router
+        tier: node
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
+    spec:
+      serviceAccountName: kube-router
+      serviceAccount: kube-router
+      containers:
+      - name: kube-router
+        image: cloudnativelabs/kube-router
+        imagePullPolicy: Always
+        args:
+        - --run-router=true
+        - --run-firewall=true
+        - --run-service-proxy=true
+        - --kubeconfig=/var/lib/kube-router/kubeconfig
+        env:
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: KUBE_ROUTER_CNI_CONF_FILE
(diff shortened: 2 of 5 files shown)


app/controllers/network_routes_controller.go
```diff
@@ -1502,6 +1503,14 @@ func NewNetworkRoutingController(clientset *kubernetes.Clientset,
 	nrc.bgpRRServer = false
 	nrc.bgpServerStarted = false
 
+	nrc.cniConfFile = os.Getenv("KUBE_ROUTER_CNI_CONF_FILE")
+	if nrc.cniConfFile == "" {
+		nrc.cniConfFile = "/etc/cni/net.d/10-kuberouter.conf"
+	}
+	if _, err := os.Stat(nrc.cniConfFile); os.IsNotExist(err) {
+		return nil, errors.New("Invalid CNI conf file " + nrc.cniConfFile)
```
andrewsykim: `s/Invalid CNI conf/CNI conf file does not exist` - the current error sounds like the file exists but it is invalid

daemonset/kubeadm-kuberouter-all-features-hostport.yaml
```diff
@@ -0,0 +1,184 @@
...
+             "bridge":"kube-bridge",
+             "isDefaultGateway":true,
+             "ipam":{
+                "type":"host-local"
+             }
+          },
+          {
+             "type":"portmap",
+             "capabilities":{
+                "snat":true,
```
andrewsykim: is `snat` a requirement to use this?

daemonset/kubeadm-kuberouter-all-features-hostport.yaml
```diff
@@ -0,0 +1,184 @@
...
+metadata:
+  name: kube-router-cfg
+  namespace: kube-system
+  labels:
+    tier: node
+    k8s-app: kube-router
+data:
+  cni-conf.json: |
+    {
+       "cniVersion":"0.3.0",
```
andrewsykim: Isn't `0.6.0` the most recent stable version? Why 0.3.0?

utils/pod_cidr.go
```diff
@@ -41,17 +59,53 @@ func InsertPodCidrInCniSpec(cniConfFilePath string, cidr string) error {
...
+			return fmt.Errorf("Failed to insert subnet cidr into CNI conf file: %s", err.Error())
+		}
+	} else {
+		config := make(map[string]interface{})
+		err = json.Unmarshal(file, &config)
+		if err != nil {
+			return fmt.Errorf("Failed to parse JSON from CNI conf file: %s", err.Error())
+		}
+		config["ipam"].(map[string]interface{})["subnet"] = cidr
+		configJSON, _ := json.Marshal(config)
```
andrewsykim: Seems like these three lines can be added after the if/else 
```go
configJSON, _ := json.Marshal(config)
err = ioutil.WriteFile(cniConfFilePath, configJSON, 0644)
if err != nil {
     return fmt.Errorf("Failed to insert subnet cidr into CNI conf file: %s", err.Error())
}
```

utils/pod_cidr.go
```diff
@@ -41,17 +59,53 @@ func InsertPodCidrInCniSpec(cniConfFilePath string, cidr string) error {
...
 		return fmt.Errorf("Failed to load CNI conf file: %s", err.Error())
 	}
-	config := make(map[string]interface{})
-	err = json.Unmarshal(file, &config)
-	if err != nil {
-		return fmt.Errorf("Failed to parse JSON from CNI conf file: %s", err.Error())
-	}
+	if strings.HasSuffix(cniConfFilePath, ".conflist") {
+		var config interface{}
+		err = json.Unmarshal(file, &config)
```
andrewsykim: Wondering if it would help if we unmarshal this into [libcni.NetworkConfig](https://github.com/containernetworking/cni/blob/master/libcni/api.go#L39-L42).

andrewsykim: or [NetConf](https://github.com/containernetworking/cni/blob/master/pkg/types/types.go#L59-L68)

daemonset/kubeadm-kuberouter-all-features-hostport.yaml
murali-reddy: This is CNI spec version https://github.com/containernetworking/cni/blob/master/SPEC.md Plug-in's vesioning is different. I got to know about it for this PR :)

daemonset/kubeadm-kuberouter-all-features-hostport.yaml
murali-reddy: I think so. Though its not clear from https://github.com/containernetworking/plugins/tree/master/plugins/meta/portmap

app/controllers/network_routes_controller.go
murali-reddy: Will change it.

utils/pod_cidr.go
murali-reddy: Done!

utils/pod_cidr.go
murali-reddy: `NetConf` structure is little wierd. 

https://github.com/containernetworking/cni/blob/master/pkg/types/types.go#L70-L72

It does not have the CIDR, did not help much with unmarshalling.

daemonset/kubeadm-kuberouter-all-features-hostport.yaml
andrewsykim: TIL

utils/pod_cidr.go
```diff
@@ -41,15 +59,46 @@ func InsertPodCidrInCniSpec(cniConfFilePath string, cidr string) error {
...
 		return fmt.Errorf("Failed to load CNI conf file: %s", err.Error())
 	}
-	config := make(map[string]interface{})
-	err = json.Unmarshal(file, &config)
-	if err != nil {
-		return fmt.Errorf("Failed to parse JSON from CNI conf file: %s", err.Error())
-	}
+	var config interface{}
+	if strings.HasSuffix(cniConfFilePath, ".conflist") {
+		err = json.Unmarshal(file, &config)
```
andrewsykim: Same with Unmarshal, seems like we can move it before the if/else

andrewsykim:
lgtm, I think it would be nice to add test cases for this in here https://github.com/murali-reddy/kube-router/blob/master/utils/pod_cidr_test.go#L61

daemonset/kubeadm-kuberouter-all-features-hostport.yaml
andrewsykim: Seems like it sets up SNAT chain, might be good to double check if it conflicts with any SNAT chains kube-router creates

t3hmrman:
A few questions/suggestions:
- is a documentation change required?
- assuming the relevant `Dockerfile`s aren't changing, could we also have a command line option as well?

murali-reddy:
Yes, documentation change would be required. I will update the PR once i have working setup and be able to test portmap functionality. Right now i am struggling to get a setup where I can test. No changes are needed in Dockerfile.

daemonset/kubeadm-kuberouter-all-features-hostport.yaml
murali-reddy: @andrewsykim ok I checked, there is no conflict with any other rules.

```
Chain PREROUTING (policy ACCEPT 0 packets, 0 bytes)
    7   468 CNI-HOSTPORT-DNAT  all  --  *      *       0.0.0.0/0            0.0.0.0/0            ADDRTYPE match dst-type LOCAL

Chain POSTROUTING (policy ACCEPT 24 packets, 1440 bytes)
    0     0 CNI-HOSTPORT-SNAT  all  --  *      *       127.0.0.1           !127.0.0.1

Chain CNI-DN-1d1e1acf7cbbe4ef6121a (1 references)
    1    60 DNAT       tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            tcp dpt:8086 to:10.1.1.32:80

Chain CNI-HOSTPORT-DNAT (2 references)
  135  8148 CNI-DN-1d1e1acf7cbbe4ef6121a  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* dnat name: "mynet" id: "2fd850ee20c7105656a5b5ccd4669c04f5a9ed8035f2f35c41a5aa7e6cf4e81f" */

Chain CNI-HOSTPORT-SNAT (1 references)
    0     0 CNI-SN-1d1e1acf7cbbe4ef6121a  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* snat name: "mynet" id: "2fd850ee20c7105656a5b5ccd4669c04f5a9ed8035f2f35c41a5aa7e6cf4e81f" */

Chain CNI-SN-1d1e1acf7cbbe4ef6121a (1 references)
    0     0 MASQUERADE  tcp  --  *      *       127.0.0.1            10.1.1.32            tcp dpt:80
```

PR #339, opened 2018-03-13
[jjo] support also using status.loadBalancer.ingress IPs via annotation

jjo:


SEJeff:
:+1: nicely done! Tests would be nice though

jjo:
> nicely done! Tests would be nice though
@SEJeff￼: tnx! , added tests at f30795d ;)

murali-reddy:
@jjo Can you please rebase. Apologies for delay in the review.

Wondering do we need an annotation? If the load balancer IP is set would it make sense to interpret it just as external IP.

jjo:
> @jjo Can you please rebase. Apologies for delay in the review.

nw, just done it.

> Wondering do we need an annotation? If the load balancer IP is set would it make sense to interpret it just as external IP.

Afaics the (only?) use case for landing LB IPs locally at nodes would
be indeed baremetal loadbalancing, as cloud-y LBs will obviously
handle those public IPs themselves, that's why I chose to keep the
current behavior unless annotated.
One clear consequence of above would be that a `kube-router` deploy
on cloud would start advertising those LB IPs towards k8s nodes,
possibly interfering with cloud provider net infra.

murali-reddy:
> Afaics the (only?) use case for landing LB IPs locally at nodes would
be indeed baremetal loadbalancing, as cloud-y LBs will obviously
handle those public IPs themselves, that's why I chose to keep the
current behavior unless annotated.
One clear consequence of above would be that a kube-router deploy
on cloud would start advertising those LB IPs towards k8s nodes,
possibly interfering with cloud provider net infra.

So i see two different aspects here. One aspect is setting up service proxy to the LoadBalancer IP as VIP and setting up its endpoints. Second is how the traffic really ends up at a node for load-balncer IP.

In baremetal deployments each node (through kube-router) can advertises the loadbalancer IP so we have ECMP load balancing. Kube-router already does this with `--advertise-cluster-ip` and `--advertise-external-ip`. IMO we could have similar flag for load balancer IP's.

I am not sure what benefit we get enabling selectivley few services with annotation.

SEJeff:
Concur. Perhaps flip it to have an annotation to turn it off (if a user is worried about it interfering with cloud provider networking), or as a kube-router flag. This has the benefit of being a massively better user experience for both on premise clusters (like mine) and cloud clusters running kube-router.

jjo:
Thanks @murali-reddy @SEJeff for the feedback, I still think the user
should be able to selectively dis/en-able which LB services to treat "as"
externalIPs, concur on having for a more useful default behavior,
I like @SEJeff take on having sth like `--advertise-loadbalancer-ip` plus
per-service annotation to disable it, I'll implement it.

murali-reddy:
Thanks @jjo for your patience :) i will test it out this functionality with ELB to see what would be ideal default behaviour.

app/controllers/network_routes_controller.go
```diff
@@ -486,16 +460,43 @@ func (nrc *NetworkRoutingController) getClusterIps() ([]string, error) {
...
+			if svc.Spec.ExternalTrafficPolicy == v1core.ServiceExternalTrafficPolicyTypeLocal {
+				nodeHasEndpoints, err := nrc.nodeHasEndpointsForService(svc)
+				if err != nil {
+					glog.Errorf("error determining if node has endpoints for svc: %q error: %v", svc.Name, err)
+					continue
+				}
+
+				if !nodeHasEndpoints {
+					for _, externalIP := range svc.Spec.ExternalIPs {
+						err := nrc.UnadvertiseClusterIp(externalIP)
```
andrewsykim: I don't think it makes sense to call `UnadvertiseClusterIp` in `getExternalIps()`. WDTY?

app/controllers/network_services_controller.go
```diff
@@ -427,7 +431,13 @@ func (nsc *NetworkServicesController) syncIpvsServices(serviceInfoMap serviceInf
 		// based on FWMARK to enable Direct server return functionality. DSR requires a director
 		// without a VIP http://www.austintek.com/LVS/LVS-HOWTO/HOWTO/LVS-HOWTO.routing_to_VIP-less_director.html
 		// to avoid martian packets
-		for _, externalIP := range svc.externalIPs {
+		extIPSet := sets.NewString(svc.externalIPs...)
+		if svc.uselbips {
+			extIPSet = extIPSet.Union(sets.NewString(svc.loadBalancerIPs...))
+		}
+		glog.V(2).Infof("Service \"%s\" using extIPSet: %v", svc.name, extIPSet.List())
```
andrewsykim: not: you can use `%q` if you want quotes on your strings

Documentation/README.md
```diff
@@ -266,38 +266,21 @@ kubectl annotate service my-service "kube-router.io/service.scheduler=dh"
...
-    - ip: 192.0.2.1
-```
-
-The `kube-router.io/service.uselbips` per service annotation will make
-the ingress IP(s) set by the LoadBalancer, in addition to externalIPs, to:
-* be locally added to nodes (to `kube-dummy-if` network interface, to LVS)
+If you want to also advertise loadbalancer set IPs
+(`status.loadBalancer.ingress` IPs), e.g. when using it with MetalLb,
+add the `--advertise-loadbalancer-ip` flag (`false` by default).
+
```
SEJeff: This is perfect from a user workflow. Thanks for humoring me!

jjo:
Thanks folks for the reviews - I've messed with the branches while
rebasing against master, will reopen another PR from a cleanup tree
(and refer to this one for the discussion)

PR #340, opened 2018-03-15
Dockefile: use ADD instead of COPY and rearrange

jjo:
Couple changes to improve local development, building, etc.

* use ADD instead of COPY to force cache invalidation
* rearrange to cache more expensive network downloads (apk, etc),
  ADDing locally built binaries at last

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,5 @@
 FROM alpine:3.7
 
-COPY kube-router gobgp /usr/local/bin/
-COPY build/image-assets/bashrc /root/.bashrc
-COPY build/image-assets/profile /root/.profile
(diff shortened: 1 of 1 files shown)


murali-reddy:
@jjo Any advantage of using ADD over COPY?

jjo:
I find ADD cache invalidation worth to be 100% sure the
image has the fresh built binaries, from
From https://docs.docker.com/engine/reference/builder/#add
```
Note: The first encountered ADD instruction will invalidate the cache for all
following instructions from the Dockerfile if the contents of <src> have changed.
This includes invalidating the cache for RUN instructions.
```
Anyway, I'm not a Docker(file) expert, so could be wrong there.

PR #344, opened 2018-03-17
ipvs: add --ipvs-ignore-key for manually configured services

dlamotte:
This is simply an escape hatch to tell kube-router not to touch specific
ipvs service's that have been manually configured outside the scope of
kube-router.

### DIFF
--- a/app/controllers/network_services_controller.go
+++ b/app/controllers/network_services_controller.go
@@ -58,6 +58,7 @@ var (
 type NetworkServicesController struct {
 	nodeIP              net.IP
 	nodeHostName        string
+	ipvsIgnoreKey       []string
 	syncPeriod          time.Duration
 	mu                  sync.Mutex
 	serviceMap          serviceInfoMap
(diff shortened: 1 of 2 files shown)


murali-reddy:
I actually happened to [propose solution](https://github.com/kubernetes/kubernetes/issues/59507) to kube-proxy IPVS proxier but dont know it is acceptiable for kube-router though.

I dont have better way out, so changes look good to me as hack.

dlamotte:
Thanks for the link! I like your suggestion to use finalizers as a general solution.

This is for sure a hack in comparison. Appreciate your willingness to move forward regardless.

I'm simply attempting to create an ipvs load balancer in front of the apiserver in a self hosted cluster.

dlamotte:
Actually, now that I think of it, I might be able to convince kube-router to create and manage that ipvs service for me. Was concerned about it not being able to reach the apiserver... Maybe this hack isn't necessary.

dlamotte:
From the issue you linked me, I'm ok closing this for the time being.  I might come back with a new PR but use a CIDR instead of a one off IP.  If/when I get around to it.

Thanks again!

murali-reddy:
@dlamotte Sure. Lets see if a better solution emerges for with kube-router or kube-proxy. Or else we can always re-visit this PR.

Issue #250, opened 2017-12-14
Node podCIDR should be EOL'ed 


murali-reddy:
[Node podCIDR should be EOL'ed ](https://github.com/kubernetes/kubernetes/issues/57130)

This is not good for Kube-Router. Part of simplicty of kube-router comes from fact kube-control manager does subnet allocation for the nodes.

Watch out this issue to see how things will evolve.

aauren:
I'm going to close this as stale. While podCIDR may be EOL'd upstream hasn't done any work to make this happen, probably because too many things rely on it. podCIDR is a core part of kube-router, and it would take a significant re-write of kube-router to divorce ourselves from it.

If we get close to taking on that level of work, we can always re-open this issue or open a new one when we're ready for that step. In the meantime, I don't think that having this in the backlog adds much.

PR #345, opened 2018-03-21
allow pod CIDR to be set using node annotations

andrewsykim:
This allows you to set pod CIDR on a node by setting the annotation `kube-router.io/pod-cidr`. The pod CIDR in the annotation takes precedence over `node.Spec.PodCIDR`. Generally people should still rely on `node.Spec.PodCIDR` (set by the control plane), but this is a nice feature to have in case you need to migrate your cluster into a new IP space.


murali-reddy:
Perhaps comes handy when https://github.com/cloudnativelabs/kube-router/issues/250 becomes serious issue.

murali-reddy:
LGTM

PR #347, opened 2018-03-21
withdraw ClusterIP bgp route for externalTrafficPolicy=Local when there are no local endpoints

TvL2386:
Hi guys,

I noticed that when using a Service object with type: Loadbalancer and externalTrafficPolicy: Local, the routes are not withdrawn. On my external BGP peer I could still see the route which caused for dropped packets when that route was used.

This pull request withdraws routes for this externalTrafficPolicy when there are no endpoints.

### DIFF
--- a/app/controllers/network_routes_controller.go
+++ b/app/controllers/network_routes_controller.go
@@ -392,6 +392,11 @@ func (nrc *NetworkRoutingController) advertiseClusterIPs() {
 				}
 
 				if !nodeHasEndpoints {
+					err := nrc.UnadvertiseClusterIp(svc.Spec.ClusterIP)
+					if err != nil {
+						glog.Errorf("error unadvertising cluster IP: %q error: %v", svc.Spec.ClusterIP, err)
+					}
+
 					continue
 				}
 			}
@@ -421,6 +426,13 @@ func (nrc *NetworkRoutingController) advertiseExternalIPs() {
 				}
 
 				if !nodeHasEndpoints {
+					for _, externalIP := range svc.Spec.ExternalIPs {
+						err := nrc.UnadvertiseClusterIp(externalIP)
+						if err != nil {
+							glog.Errorf("error unadvertising external IP: %q, error: %v", externalIP, err)
+						}
+					}
+
 					continue
 				}
 			}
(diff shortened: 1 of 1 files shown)


SEJeff:
Refs #347

app/controllers/network_routes_controller.go
```diff
@@ -680,6 +692,23 @@ func (nrc *NetworkRoutingController) AdvertiseClusterIp(clusterIp string) error
...
+func (nrc *NetworkRoutingController) UnadvertiseClusterIp(clusterIp string) error {
+	glog.V(2).Infof("Unadvertising route: '%s/%s via %s' to peers", clusterIp, strconv.Itoa(32), nrc.nodeIP.String())
+
+	pathList := []*table.Path{table.NewPath(nil, bgp.NewIPAddrPrefix(uint8(32),
+		clusterIp), true, nil, time.Now(), false)}
+
+	err := nrc.bgpServer.DeletePath([]byte(nil), 0, "", pathList)
+
+	if err != nil {
+		return fmt.Errorf(err.Error())
```
andrewsykim: just 
```go
return err 
```

andrewsykim:
Tested this patch and it works as we expect. However, the convergence time in kube-router is taking a bit longer than I would like once the route is removed.

andrewsykim:
Might be a good time to refactor to use informers so we can get the update events for endpoints right away, as opposed to listing all endpoints on each sync loop

andrewsykim:
cc @murali-reddy

app/controllers/network_routes_controller.go
TvL2386: I agree with you, but the AdvertiseClusterIp method above it should be refactored as well then, since the methods are structured the same

andrewsykim: feel free to fix the method above too :)

TvL2386: alrighty then :)

TvL2386:
I agree that the convergence is not really acceptable. Ideally it should be ran immediately when there are no endpoints left on the node.

andrewsykim:
LGTM, thanks for the PR @TvL2386 

I'll try to do some work in the upcoming week to shorten the convergence time.

PR #351, opened 2018-03-22
add 1s deley between ec2 API calls, and check if node has ec2 IAM

murali-reddy:
- Adding 1s delay between ec2 API calls
- node performs disable src-dst check ONLY if it has necessary IAM


murali-reddy:
Merging as fix worked. https://kubernetes.slack.com/archives/C8DCQGTSB/p1521727392000747

PR #353, opened 2018-03-24
clear state of global service/endpoint watcher in Test_addExportPolicies

andrewsykim:
Calls StartServiceWatcher and StartEndpointsWatcher in Test_addExportPolicies so that fake clientset state is cleared on each run. We should consider refactoring kube-router so we're not using global lister/watchers 

ref: https://github.com/cloudnativelabs/kube-router/pull/352


andrewsykim:
@murali-reddy done

PR #354, opened 2018-03-25
[jjo] support advertising status.loadBalancer.ingress IPs via flag

jjo:
* add `--advertise-loadbalancer-ip` flag, which will make Service's
  Ingress IP(s)set by the LoadBalancer to:
  - be locally added to nodes' `kube-dummy-if` network interface
  - be advertised to BGP peers

* support "kube-router.io/service.skiplbips=true" per Service
  annotation to selectively skip above

* refactor several functions with dupped code to streamline logic as:
  - `getIpsToAdvertise()` which calls ->
    - `getClusterIPs()`
    - `getExternalIPs()`
    - `getLoadBalancerIPs()`
    and contains nodeHasEndpoints logic, returns:
      (ipsToAdvertise, ipsToUnAdvertise)
  - advertiseIPs() which is essentially previous advertiseClusterIPs()
    (which was previously used to advertise _any_ IP actually, ie misnamed),
    with logic to un/advertise based on both passed arguments:
      (ipsToAdvertise, ipsToUnAdvertise)

This is cleanup / rebase from PR #339, find there previous discussion
with @SEJeff @murali-reddy @andrewsykim re: usage and implementation.

### DIFF
--- a/Documentation/README.md
+++ b/Documentation/README.md
@@ -264,6 +264,25 @@ For destination hashing scheduling use:
 kubectl annotate service my-service "kube-router.io/service.scheduler=dh"
 ```
 
+### LoadBalancer IPs
+
+If you want to also advertise loadbalancer set IPs
+(`status.loadBalancer.ingress` IPs), e.g. when using it with MetalLb,
+add the `--advertise-loadbalancer-ip` flag (`false` by default).
+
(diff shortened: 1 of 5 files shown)


murali-reddy:
Thanks @jjo LGTM

Issue #280, opened 2018-01-11
Wrong node ASN number used


murali-reddy:
When full mesh is disabled. A node peers up with only nodes with same ASN. But the ASN number used is default ASN number and not the one configured on the node object annotation.

https://github.com/cloudnativelabs/kube-router/blob/v0.0.20/app/controllers/network_routes_controller.go#L997

cjbrigato:
May it be the same but I was mentioning off-topic on the route-reflector Issue there https://github.com/cloudnativelabs/kube-router/issues/192#comment-348304914 ?

murali-reddy:
Fixed as part of #362

PR #362, opened 2018-04-03
fix bug where default ASN is used for iBGP peering

andrewsykim:
Fixes a bug where the default ASN was used (64512) even though the annotation `kube-router.io/node.asn` was set.


murali-reddy:
@andrewsykim I noted down to fix it https://github.com/cloudnativelabs/kube-router/issues/280, somehow missed it.

@killcity Please use the image `cloudnativelabs/kube-router-git:PR-363` with the fix in this PR to see if this fixes your problem.

SEJeff:
Probably worth adding tests for this as well (good fix regardless @andrewsykim)

murali-reddy:
Its important unit test case @SEJeff One of use will add the unit test.

PR #368, opened 2018-04-04
Fixed the /var/run/docker.sock permissions in the DSR example

farcaller:
I guess the example config was copied wrong. https://github.com/cloudnativelabs/kube-router/blob/master/daemonset/kubeadm-kuberouter-all-features.yaml has `kubeconfig` set to RO but I think you need RW access to talk to docker?

Alternatively both need to be RO


murali-reddy:
> Alternatively both need to be RO

Yes, both need to be read-only. Could you please update the patch?

farcaller:
Done.

murali-reddy:
Thanks you @farcaller

PR #369, opened 2018-04-05
trim end newline in byte[] base64.StdEncoding.DecodeString

murali-reddy:
For BGP kube-router takes base64 encoded string as passoword as described in

https://github.com/cloudnativelabs/kube-router/blob/master/Documentation/bgp.md#base64-encoding-passwords

While decoding to plain text, a newline was part of byte array casuing problems. Fix trims the end new line.

### DIFF
--- a/app/controllers/network_routes_controller.go
+++ b/app/controllers/network_routes_controller.go
@@ -562,11 +562,12 @@ func stringSliceB64Decode(s []string) ([]string, error) {
 	ss := make([]string, 0)
 	for _, b64String := range s {
 		decoded, err := base64.StdEncoding.DecodeString(b64String)
+		decodedStr := strings.TrimSuffix(string(decoded), "\n")
 		if err != nil {
 			return nil, fmt.Errorf("Could not parse \"%s\" as a base64 encoded string",
 				b64String)
 		}
-		ss = append(ss, string(decoded))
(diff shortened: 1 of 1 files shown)


app/controllers/network_routes_controller.go
```diff
@@ -562,11 +562,12 @@ func stringSliceB64Decode(s []string) ([]string, error) {
 	ss := make([]string, 0)
 	for _, b64String := range s {
 		decoded, err := base64.StdEncoding.DecodeString(b64String)
+		decodedStr := strings.TrimSuffix(string(decoded), "\n")
```
tamalsaha: bytes.TrimSpace() ?

murali-reddy: Space is not a problem. base64 encoding adds a new line. May be i will try 

https://golang.org/pkg/encoding/base64/#Encoding.Decode

murali-reddy: Kubernetes [advises](https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually) to use `echo -n "pasword" | base64` where
```
-n     do not output the trailing newline
```

roffe:
Let's not add magic handling of user error in the b64 parser, the docs has been updated using `printf` instead of `echo`

PR #371, opened 2018-04-06
always generate IPIP tunnels regardless of node subnets

andrewsykim:
I'm not sure how graceful this change will be for existing clusters that have the previous routes (no IPIP tunnel). At one point we'll want to account for clusters that are upgrading from a previous version. I think for now it's okay :)


andrewsykim:
@murali-reddy and I had a conversation and decided that we want to be keep the existing behaviour, I'll update this PR soon

andrewsykim:
Closing in favor of https://github.com/cloudnativelabs/kube-router/pull/666

PR #372, opened 2018-04-07
update command line options documentation

phemmer:
This section of the doc was fairly out of date.
Updated it, as well as added a test to ensure any future changes help changes are properly documented.

### DIFF
--- a/kube-router.go
+++ b/kube-router.go
@@ -14,7 +14,14 @@ import (
 )
 
 func main() {
+	if err := Main(); err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err)
+		os.Exit(1)
+	}
+	os.Exit(0)
+}
 
+func Main() error {
 	config := options.NewKubeRouterConfig()
 	config.AddFlags(pflag.CommandLine)
 	pflag.Parse()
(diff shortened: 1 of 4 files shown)


kube-router.go
```diff
@@ -10,10 +10,20 @@ import (
...
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 )
 
 func main() {
+	if err := Main(); err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err)
+		os.Exit(1)
+	}
+	os.Exit(0)
```
phemmer: This change (splitting main into 2 functions) was so that I could call `main` in the test and not have it `os.Exit()` the test.

murali-reddy:
@phemmer thanks. Indeed flags went outdated. Thanks for adding unit test as well, to make future proof.

phemmer:
Fixed for failing build. For some reason `go test ./` generated different output than `go test github.com/cloudnativelabs/kube-router` (one of the items was annotated as `uints` vs `uintSlice`). Don't know why. Only reason I can figure is due to vendoring, but they should both use it.

murali-reddy:
thanks @phemmer

PR #348, opened 2018-03-21
[jjo] added network_services_controller ginkgo testing

jjo:
FYI no new functionality, but just golang interface hooks,
to be able to do BDD network_services_controller testing.

* created a shim/mostly-nil LinuxNetworking go interface,
  (to be able to replace it with mocks for testing)
  hooked all linux networking subsys calls thru it

  `@reviewer`: note that `network_services_controller.go`
  modifications to existing code are indeed e.g.
  replacing:
  ~~~
  ipvsClusterVipSvc, err := ipvsAddService(...)
  ~~~
  by:
  ~~~
  ipvsClusterVipSvc, err := nsc.ln.ipvsAddService(...)
  ~~~

* added `network_services_controller_moq.go` autogenerated
  with `moq`

* using `ginkgo`, created `network_services_controller_test.go`
  and other autogenerated supporting files

* updated `vendor/` deps for
  - github.com/onsi/ginkgo
  - github.com/onsi/gomega

* added `network_services_controller_test.go` testing for:
  - Context("service no endpoints with externalIPs", ...)
  - Context("service no endpoints with loadbalancer IPs with annotation", ...)
  - Context("service no endpoints with loadbalancer IPs without annotation", ...)
  - Context("service no endpoints with loadbalancer without IPs", ...)
  - Context("node has endpoints for service", ...)

PR #373, opened 2018-04-08
refactor to use shared informers

andrewsykim:
* removes `/app/watchers` package and all it's global variables. 
* uses shared informer factory to consolidate event handling into one informer per resource
* uses shared informers `WatchForCacheSync` to consolidate all calls to `HasSynced`
* uses interfaces where possible for more testable code 
* makes a few optimizations since we can do less type assertions


app/controllers/network_policy_controller.go
```diff
@@ -120,16 +130,12 @@ func (npc *NetworkPolicyController) Run(healthChan chan<- *ControllerHeartbeat,
...
-		if watchers.PodWatcher.HasSynced() && watchers.NetworkPolicyWatcher.HasSynced() {
-			glog.V(1).Info("Performing periodic sync of iptables to reflect network policies")
-			err := npc.Sync()
-			if err != nil {
-				glog.Errorf("Error during periodic sync: " + err.Error())
-			} else {
-				sendHeartBeat(healthChan, "NPC")
-			}
+		glog.V(1).Info("Performing periodic sync of iptables to reflect network policies")
+		err := npc.Sync()
```
andrewsykim: I think we can technically get rid of this because it should be covered by the resync period we pass into the informers (this was the case previously too)

andrewsykim: same with service and routes controller

app/server.go
```diff
@@ -157,32 +106,55 @@ func (kr *KubeRouter) Run() error {
...
 	}
 
 	if kr.Config.RunRouter {
-		nrc, err := controllers.NewNetworkRoutingController(kr.Client, kr.Config)
+		nrc, err := controllers.NewNetworkRoutingController(kr.Client, kr.Config, nodeInformer, svcInformer, epInformer)
 		if err != nil {
 			return errors.New("Failed to create network routing controller: " + err.Error())
 		}
 
+		nodeInformer.AddEventHandler(nrc.NodeEventHandler)
```
andrewsykim: I will follow up in another PR so that routes controller has service and endpoints event handlers

app/controllers/network_policy_controller.go
```diff
@@ -1427,9 +1478,14 @@ func NewNetworkPolicyController(clientset *kubernetes.Clientset, config *options
...
 	npc.ipSetHandler = ipset
 
-	watchers.PodWatcher.RegisterHandler(&npc)
-	watchers.NetworkPolicyWatcher.RegisterHandler(&npc)
-	watchers.NamespaceWatcher.RegisterHandler(&npc)
+	npc.podLister = podInformer.GetIndexer()
+	npc.PodEventHandler = npc.newPodEventHandler()
+
+	npc.nsLister = nsInformer.GetIndexer()
+	npc.NamespaceEventHandler = npc.newNetworkPolicyEventHandler()
```
andrewsykim: oops

andrewsykim:
There's some conflicts from https://github.com/cloudnativelabs/kube-router/pull/348 I'll fix it shortly

app/controllers/network_policy_controller.go
andrewsykim: Alternatively keep this but set the resync period to 0 for all informers

murali-reddy: Writing controller that handles all the corner cases is hard :)

Ideally (in the sense of strong consistency) we recive an update and act on it. Additionally informer can be set to a resync period which will take care of any missed updates from API server for what ever reason. 

If we remove periodic sync, in this scenario, there is no control loop to reconcile (eventual consistent) desired state and actual state.

Alternativley if we keep resync period to 0 then if for any reason cache/informer miss an update from API server, then it will have stale data.

AFAIK, earlier design and with your changes, controller (w.r.t resync period, and periodic loop) is similar to kube-proxy.

murali-reddy:
thanks @andrewsykim this is massive clean up.

app/controllers/network_policy_controller.go
andrewsykim: I think the right thing for our case is to set resync period on informers to 0 and keep the main period sync loop we have for each controller (can be in a future PR). Reason being is that if each `OnUpdate` call is just calling `npc.sync()` then we're calling `npc.sync()` per resource update event when we only need to call it once.

andrewsykim: Right now we set resync period and run a separate syncloop which is is unnecessary.

PR #379, opened 2018-04-10
Fix networkpolicies if there are unscheduled pod, log errors (#378)

### LINKED ISSUE #378: networkpolicy bug: ipset not updated if there are unscheduled pods
If there are unscheduled pods with matching networkpolicies, kube-router calls `ipset add` with empty IPs for corresponding networkpolicies and fails to update the IP sets (`ipset swap` is not called).  Therefore, the ipsets
stay frozen and the networkpolicies cannot grant access to/from new pods (with changed IP address).

Possible solution would be to ignore pods in `buildNetworkPoliciesInfo` with empty/unset `matchingPod.Status.PodIP`.

Additionally, there are currently no error handling for ipset.Refresh calls in network_policy_controller.go, such that errors like the described above are not visible in the logs.

xanonid:
Fixes #378

### DIFF
--- a/app/controllers/network_policy_controller.go
+++ b/app/controllers/network_policy_controller.go
@@ -283,8 +283,14 @@ func (npc *NetworkPolicyController) syncNetworkPolicyChains() (map[string]bool, 
 			currnetPodIps = append(currnetPodIps, ip)
 		}
 
-		targetSourcePodIpSet.Refresh(currnetPodIps, utils.OptionTimeout, "0")
-		targetDestPodIpSet.Refresh(currnetPodIps, utils.OptionTimeout, "0")
+		err = targetSourcePodIpSet.Refresh(currnetPodIps, utils.OptionTimeout, "0")
+		if err != nil {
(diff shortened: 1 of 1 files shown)


murali-reddy:
@xanonid thanks for your patch. Please apply same fix in the function `buildNetworkPoliciesInfo()` this is what is relevent. `buildBetaNetworkPoliciesInfo` is function for supporting old clusters with beta network policy semantics. Perhaps i should remove the support for it.

```
	if npc.v1NetworkPolicy {
		npc.networkPoliciesInfo, err = buildNetworkPoliciesInfo()
		if err != nil {
			return errors.New("Aborting sync. Failed to build network policies: " + err.Error())
		}
	} else {
		// TODO remove the Beta support
		npc.networkPoliciesInfo, err = buildBetaNetworkPoliciesInfo()
		if err != nil {
			return errors.New("Aborting sync. Failed to build network policies: " + err.Error())
		}
	}
```

xanonid:
@murali-reddy thanks for the review. I fixed now buildNetworkPoliciesInfo and updated the patch.

Issue #380, opened 2018-04-10
Fine tuning informers resync period 


andrewsykim:
As per discussion here https://github.com/cloudnativelabs/kube-router/pull/373#discussion_r179933074 we should investigate what resync period we should be using for informers. A resync period of 0 might be okay given we run a seperate sync loop in each controller.

murali-reddy:
Part of #399 resync period of shared index informer is set to 0. As per the [discussions](https://groups.google.com/forum/#!topic/kubernetes-sig-api-machinery/PbSCXdLDno0) is sig-api machinery and on my own verfication, it is observed all the objects from cache's are replayed which can get really expensive operation (processin updated). There is no real benefit to set to any value greater than 0, instead periodic full list of objects and ensure if any reconciliation are required is better.

PR #384, opened 2018-04-11
add service/endpoint event handler for routes controller

andrewsykim:


murali-reddy:
Overall changes look good. Please consider below change in this PR.

Its ambiguous to use `advertiseIPs`, `withdrawIPs`, `bgpAdvertiseIP` etc. Since kube-router also advertising the pod CIDR's from the same module, i suggest names like below, so the intent is more explicit.

- advertiseServiceVips()
- withdrawServiceVips()
- bgpAdvertiseServiceVip()

A general comment from this PR and #373. I hae opened https://github.com/cloudnativelabs/kube-router/issues/386 . No need to block this PR on #386.

I belive we need to sort out both #386 and #380, after the number of refactors that went in lately.

andrewsykim:
@murali-reddy used `advertiseVIPs` and `withdrawVIPs` since it's also external/LB IPs, PTAL

murali-reddy:
Thanks @andrewsykim LGTM

PR #390, opened 2018-04-13
ignore update events for endpoints used for leader election

andrewsykim:


app/controllers/network_services_controller.go
```diff
@@ -49,6 +49,8 @@ const (
 	svcHairpinAnnotation   = "kube-router.io/service.hairpin"
 	svcLocalAnnotation     = "kube-router.io/service.local"
 	svcSkipLbIpsAnnotation = "kube-router.io/service.skiplbips"
+
+	LeaderElectionRecordAnnotationKey = "control-plane.alpha.kubernetes.io/leader"
```
andrewsykim: this is a constant copied from client-go, I'll use the one in client-go once the vendor situation is fixed

murali-reddy:
I wold nuke logging completley if its for kubernetes components `kube-scheduler` and `kube-controller-manager`. Also info level is more noise, by default in the log. may be use debug?

andrewsykim:
Done!

PR #395, opened 2018-04-14
Use dep for managing vendor/

andrewsykim:
Some things to note:
* most the file changes are actually deletions since doing `dep init` actually filtered out a lot of dependencies that were actually not needed. 
* It seems that package `github.com/docker/distribution` was using a [fork](https://github.com/cloudnativelabs/kube-router/blob/master/glide.lock#L64) due to dependency issues with `github.com/sirupsen/logrus`. I fixed by updating the package to a version that re-vendored that package. 
* It seems that package `github.com/docker/libnetwork` was also using a [fork](https://github.com/cloudnativelabs/kube-router/blob/master/glide.lock#L99) due to similar issues with import paths, this was fixed by using a version of `docker/libnetwork` that had the updated package (there should be significant difference between the two versions.


murali-reddy:
LGTM

PR #399, opened 2018-04-17
minimize sync() done by the controllers

murali-reddy:
Following changes are done

- use resync period as 0 when creating SharedInformerFactory, this sync wil actually replay all the cached objects, resulting sync() in each controller for each object

- regular full sync will ber performed periodically. Increased defualt value to 5 min.

I will submit a new PR, where work queue will be used to reque the object so that a retry can be performed in case of error. So that full sync interval can be even pushed further.

### DIFF
--- a/pkg/cmd/kube-router.go
+++ b/pkg/cmd/kube-router.go
@@ -115,7 +115,7 @@ func (kr *KubeRouter) Run() error {
 	nsInformer := informerFactory.Core().V1().Namespaces().Informer()
 	npInformer := informerFactory.Networking().V1().NetworkPolicies().Informer()
 
-	go informerFactory.Start(stopCh)
+	informerFactory.Start(stopCh)
 	informerFactory.WaitForCacheSync(stopCh)
 
 	if kr.Config.RunFirewall {
--- a/pkg/controllers/network_policy_controller.go
+++ b/pkg/controllers/network_policy_controller.go
@@ -160,6 +160,8 @@ func (npc *NetworkPolicyController) OnPodUpdate(obj interface{}) {
 
 // OnNetworkPolicyUpdate handles updates to network policy from the kubernetes api server
 func (npc *NetworkPolicyController) OnNetworkPolicyUpdate(obj interface{}) {
+	netpol := obj.(*networking.NetworkPolicy)
+	glog.V(2).Infof("Received network policy update namespace:%s netpol name:%s", netpol.Namespace, netpol.Name)
 	err := npc.Sync()
 	if err != nil {
 		glog.Errorf("Error syncing on network policy update: %s", err)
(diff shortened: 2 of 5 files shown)


pkg/cmd/kube-router.go
```diff
@@ -115,7 +115,7 @@ func (kr *KubeRouter) Run() error {
 	nsInformer := informerFactory.Core().V1().Namespaces().Informer()
 	npInformer := informerFactory.Networking().V1().NetworkPolicies().Informer()
 
-	go informerFactory.Start(stopCh)
```
murali-reddy: removed `go `, as `Start()` method actually launches go routing for each informer an exits

pkg/controllers/network_services_controller.go
```diff
@@ -365,14 +363,19 @@ func (nsc *NetworkServicesController) OnEndpointsUpdate(obj interface{}) {
 		return
 	}
 
-	// build new endpoints map to reflect the change
+	glog.V(1).Infof("Received update to endpoints Name: %v in Namespace %v from watch API", ep.Name, ep.Namespace)
+
+	// build new service and endpoints map to reflect the change
+	newServiceMap := nsc.buildServicesInfo()
```
murali-reddy: this is required, other wise, when kube-router is booting and the `Run()` method of controller is not run yet. In which case we have empty `nsc.serviceMap` and `nsc.endpointsMap` resulting in deletion of endpoints.

pkg/controllers/network_policy_controller.go
```diff
@@ -168,12 +170,12 @@ func (npc *NetworkPolicyController) OnNetworkPolicyUpdate(obj interface{}) {
...
 func (npc *NetworkPolicyController) OnNamespaceUpdate(obj interface{}) {
+	namespace := obj.(*api.Namespace)
+	glog.V(2).Infof("Received namespece: %v update", namespace.Name)
 	// namespace (and annotations on it) has no significance in GA ver of network policy
 	if npc.v1NetworkPolicy {
 		return
 	}
-
-	namespace := obj.(*api.Namespace)
 	glog.V(2).Infof("Received update for namespace: %s", namespace.Name)
```
andrewsykim: remove this line

pkg/controllers/network_services_controller.go
```diff
@@ -381,16 +384,25 @@ func (nsc *NetworkServicesController) OnServiceUpdate(obj interface{}) {
...
 	defer nsc.mu.Unlock()
 
-	glog.V(1).Info("Received service update from watch API")
+	svc, ok := obj.(*api.Service)
+	if !ok {
+		glog.Error("could not convert service update object to *v1.Service")
+		return
+	}
+
+	glog.V(1).Infof("Received update to service Name: %v in Namespace %v from watch API", svc.Name, svc.Namespace)
```
andrewsykim: much cleaner to just do `Received update for service : %s/%s,  svc.Namespace, svc.Name`

andrewsykim:
LGTM, comments are non-blocking

PR #400, opened 2018-04-18
prevent processing updates leading to sync when controller doing full-sync at boot time

murali-reddy:
when full-sync is in progress at controller's boot time, processing updates are not necessary, so skip them. Event in to corner case if full-sync does not cover an update to a resource, next periodic full sync will reconcile the change.


murali-reddy:
Minor changes. Merging the commit.

PR #403, opened 2018-04-21
update gobgp to d31262d for BGP graceful restart fixes

andrewsykim:
Had to update `netlink` as an `override` in `Gopkg.toml` since `gobgp` had constraints that the netlink package is on `master` but we use a tagged release `1.0.0`.  If you try dep ensure without setting the override you get this:
```
$ dep ensure -update github.com/osrg/gobgp
Warning: Gopkg.lock is out of sync with Gopkg.toml or the project's imports.
Solving failure: No versions of github.com/osrg/gobgp met constraints:
	d31262de7d91c81ff979b39950d2d859666dfa3f: Could not introduce github.com/osrg/gobgp@d31262de7d91c81ff979b39950d2d859666dfa3f, as it has a dependency on github.com/vishvananda/netlink with constraint master, which has no overlap with existing constraint ^1.0.0 from (root)
```

Setting `netlink` as an override and running `dep ensure -update github.com/osrg/gobgp` updates gobgp to the version we want (along with some of the constraints it enforces). 

Also had to fix `Test_addExportPolicies`.


murali-reddy:
Thanks @andrewsykim for the fix.  At last rolling update with BGP graceful restart enabled works as expected. There is NO downtime for data path even when control plane (i.e. kube-rotuer pods) is updating.

@phemmer thanks for finding the root cause. Your finding of issue in GoBGP has been super useful.

PR #404, opened 2018-04-22
move health, metrics to sepearate packages

murali-reddy:
no functional change, just moving the health and metrics to sepearate packages


murali-reddy:
Merging. Tested both `/metrics` and `/healthz` with the changes in the PR.

PR #406, opened 2018-04-22
split routing controllers to smaller modules by function

murali-reddy:
No functional change, splitting up `network_routes_controller.go` to smaller components doing specific functionality.


pkg/controllers/routing/aws.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/golang/glog"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// disableSourceDestinationCheck disables src-dst check of all the VM's when cluster
+// is provisioned on AWS. EC2 by default drops any packets originating or destination
+// to a VM with IP other than that of VM's ip. This check needs to be disabled so that
+// cross node pod-to-pod traffic can be sent and recived by a VM.
+func (nrc *NetworkRoutingController) disableSourceDestinationCheck() {
```
andrewsykim: Is there a reason this has to iterate over all nodes and can't just check the current node?

pkg/controllers/routing/bgp_peers.go
```diff
@@ -0,0 +1,310 @@
...
+				State: config.EbgpMultihopState{
+					Enabled:     true,
+					MultihopTtl: peerMultihopTtl,
+				},
+			}
+		}
+		err := server.AddNeighbor(n)
+		if err != nil {
+			return fmt.Errorf("Error peering with peer router "+
+				"\"%s\" due to: %s", peerConfig.NeighborAddress, err)
```
andrewsykim: Can use `%q` to avoid escaping the quotes here

pkg/controllers/routing/ecmp_lb.go
```diff
@@ -0,0 +1,357 @@
+package routing
```
andrewsykim: rename to ecmp_vip.go? or `bgp_vip`?

andrewsykim:
We can probably rename `network_routes_controller.go` -> `controller.go`

pkg/controllers/routing/aws.go
murali-reddy: Agree, it should be improved. I have opened https://github.com/cloudnativelabs/kube-router/issues/407

murali-reddy:
> We can probably rename network_routes_controller.go -> controller.go

FWIW, in general i am mostly keeping the changes closer to the conventions followed in k8s [controllers](https://github.com/kubernetes/kubernetes/tree/master/pkg/controller)

PR #417, opened 2018-05-01
proxy: cleanup stale IPs on kube-dummy-if

### LINKED ISSUE #411: Garbage collect KUBE_DUMMY_IF ip assignments
Deleted or changed service IP's are not removed from the **KUBE_DUMMY_IF** on nodes.

We currently have about 70 active ClusterIP services in our cluster, but kube-dummy-if has *1059* ip-addresses assigned. This actually turned out to be a real problem, because we had another host service that was configured to listen on all interfaces and breached ulimits when trying to open its 1024+ listening socket.

Since no auto-cleanup is running and we don't want to restart kube-router with `--cleanup` every night or so, I believe it should be a reconciliation step to delete unused serviceIP's from the dummy interface, just like unused ipvs rules are purged today.

dlamotte:
Fixes #411

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -689,6 +689,35 @@ func (nsc *NetworkServicesController) syncIpvsServices(serviceInfoMap serviceInf
 		}
 	}
 
+	// cleanup stale IPs on dummy interface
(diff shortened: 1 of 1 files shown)


murali-reddy:
thanks @dlamotte for the PR. LGTM

PR #420, opened 2018-05-03
add kube-router.io/bgp.max.paths annotation

dlamotte:
Add the ability to define a max number of paths out of the cluster.
This is useful for debugging on-premises environments where upstream
routers may have issues with asynchronous routing (due to ECMP VIP).

The state of which nodes have selected themselves to be in the available
paths is available in the kube-router.io/bgp.state.paths annotation.
You should be able to steer which nodes are the L4 director by setting
this as well.

At this time, there is no controller watching for nodes to be deleted or
become NotReady and update the annotation.

Also: wasn't sure how to add in the new vendor'd dependencies so I just copied them in.

### DIFF
--- a/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
+++ b/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
@@ -160,12 +160,20 @@ rules:
     resources:
       - namespaces
       - pods
-      - services
       - nodes
       - endpoints
     verbs:
       - list
       - get
+      - watch
+  - apiGroups:
+      - ""
+    resources:
+      - services
+    verbs:
+      - list
+      - get
+      - update
       - watch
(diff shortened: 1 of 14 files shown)


dlamotte:
This is a bit broken. I'll be pushing up a change ensure the ipvs service is created on the node that is selected to advertise the route.

dlamotte:
Rebased on master.  Still need to fix issue above.

dlamotte:
Attempted to fix it.  Will test deploy internally to test.

dlamotte:
Rolled out this fix.  Not sure if it was this fix or the latest master, but broke quite a bit and I'm not clear as to why.  Rolled it back.  Not sure what could have broken it.

aauren:
Closing as this is too old to merge any longer.

PR #428, opened 2018-05-07
explicilty specify source IP to use when send traffic over tunnels

### LINKED ISSUE #427: routing on tunnel interface not set correctly
When nodes are in different subnets IPIP tunnelinig is used to send the pod-to-pod traffic across the nodes. 

So right now the tunnel interface looks like 
```
100.96.1.0/24 dev tun-1722074240  proto 17
```

While it works fine when physical interface used for the tunnel has single IP. But when there are multiple IP's on physcial interface, then traffic send on the tunnel interface, eventually may endup sending the packets with not necessarily using node IP.

This breaks the PBR routing setup on the node. Please see #421 on role PBR plays when sending the triaffic on the tunnels.

Kube-router needs to control the source IP used when sending the traffic over tunnel interface. Source IP should be Node IP only.

murali-reddy:
Change the routes created to send traffic over tunnel interface from 

```
100.96.0.0/24 dev tun-172205645  proto 17
100.96.1.0/24 dev tun-1722051121  proto 17
```

to 

```
100.96.0.0/24 dev tun-172205645  proto 17  src 172.20.95.219
100.96.1.0/24 dev tun-1722051121  proto 17  src 172.20.95.219
```

Fixes #427

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -400,6 +400,7 @@ func (nrc *NetworkRoutingController) injectRoute(path *table.Path) error {
 
 		route = &netlink.Route{
 			LinkIndex: link.Attrs().Index,
+			Src:       nrc.nodeIP,
 			Dst:       dst,
 			Protocol:  0x11,
 		}


murali-reddy:
Fixes is in `cloudnativelabs/kube-router-git:tunnel-src-ip`

PR #430, opened 2018-05-09
services: correct check for inactive service endpoints

dlamotte:
This fixes issues with ipvs services that have zero endpoints from not being cleaned up.

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -419,6 +419,31 @@ type externalIPService struct {
 	externalIp string
 }
 
+func hasActiveEndpoints(svc *serviceInfo, endpoints []endpointsInfo, nodePodCidrStr string) bool {
+	if svc.local {
+		_, nodePodCidr, err := net.ParseCIDR(nodePodCidrStr)
+		if err != nil {
+			glog.Errorf("Failed to ParseCIDR %s for hasActiveEndpoints on service %s/%s",
+				nodePodCidrStr, svc.namespace, svc.name)
+			return false
+		}
+		for _, endpoint := range endpoints {
+			ip := net.ParseIP(endpoint.ip)
+			if ip == nil {
+				glog.Errorf("Failed to ParseCIDR %s for endpoint in hasActiveEndpoints on service %s/%s",
+					endpoint.ip, svc.namespace, svc.name)
+				continue
+			}
+			if nodePodCidr.Contains(ip) {
+				return true
+			}
+		}
+		return false
+	}
+
+	return len(endpoints) > 0
+}
+
 // sync the ipvs service and server details configured to reflect the desired state of services and endpoint
 // as learned from services and endpoints information from the api server
 func (nsc *NetworkServicesController) syncIpvsServices(serviceInfoMap serviceInfoMap, endpointsInfoMap endpointsInfoMap) error {
@@ -484,6 +509,13 @@ func (nsc *NetworkServicesController) syncIpvsServices(serviceInfoMap serviceInf
 			continue
 		}
 
+		endpoints := endpointsInfoMap[k]
+
+		if !hasActiveEndpoints(svc, endpoints, nsc.podCidr) {
+			glog.V(1).Infof("Skipping service %s/%s as it does not have active endpoints\n", svc.namespace, svc.name)
+			continue
+		}
+
 		// create IPVS service for the service to be exposed through the cluster ip
 		ipvsClusterVipSvc, err := nsc.ln.ipvsAddService(ipvsSvcs, svc.clusterIP, protocol, uint16(svc.port), svc.sessionAffinity, svc.scheduler)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


murali-reddy:
@dlamotte I dont think `len(endpoints) == 0` is a necessary condition to clean up IPVS service. Clearly we could have a Kubernetes service with no endpoints active at any point. The way it is (without the fix in this PR) reflects the desired state as per the state of Kubernetes servieces and endpoints. I would rather keep the way it is unless you are seeing any issues.

murali-reddy:
@dlamotte sorry i am bit confused here. Let me understand issue here.

So the problem this PR fixes is only in the case of services with `service.local=true`/ `externalTrafficPolicy: Local`? or in general any service with 0 endpoints?

IMO if your pod's want to access the in-cluster service then I believe using `service.local=true`/ `externalTrafficPolicy: Local` is not correct. This mechanism is a _way_ primarily meant for ingress traffic to preserve source IP.

If you are using external IP and DSR, then it does not make sense to mark service with `service.local=true`/ `externalTrafficPolicy: Local`. Reason being it achives same benfit (preserving source IP) with more benefits (scale out load balancing).

dlamotte:
This fixes the behavior for pods running in the cluster that are clients to a service with `service.local=true` or `externalTrafficPolicy: Local`.  It's odd to me that, even though its an optimization for ingress traffic that pods within the cluster would be affected and unable to access the endpoint.

In our case, we're currently running an rsyncd server.  The service selects all rsyncd pods for port 873 and uses `service.local=true`.  We've setup an external DNS name for the service pointing to the external IP on the service.  We'd like to be able to say "use this name to rsync".  But currently, without this fix, someone has to decide "well, if I'm running this _in_ the cluster, I should use the internal name vs the external DNS name for external cluster".

This external name pointing to the external IP now works with this fix.

That said, I can understand an ipvs service representing the service running in a cluster that has zero endpoints.  I'm unclear on the utility of that, but I definitely can see that it would be representing the same thing whereas this fix would delete that ipvs service and so the ipvs service would be missing even though there is a service in the cluster for it (just no endpoints).

I think the value here is that services with `service.local=true` will function like they would anywhere else whereas without this fix they wont.

SEJeff:
We've ran into this exact same issue fwiw, and agree it makes sense conceptually.

dlamotte:
@murali-reddy and I chatted about this issue on slack this morning.  We were thinking it was only an issue if the IP was attached to an interface locally.  I confirmed just now that even if the IP isn't on a local interface, the ipvs service affects the traffic (like iptables).

I replicated this by creating a `kube-router.io/service.local=true` service with an external IP 10.1.1.4.  The following is how I verified on a node that did _not_ have a service endpoint running on the local node:

```
% ip addr | grep 10.1.1.4
# no records
% ip route show table all | grep 10.1.1.4
# no records
% ipvsadm -ln -t 10.1.1.4:873 # I happen to be working with an rsync service on port 873
Memory allocation problem
# standard error when service doesn't exist
% nc 10.1.1.4 873
@RSYNCD: 31.0
^C
# as you can see, I can connect to rsyncd as it responds and works which is not located on the local node
% ipvsadm -A -t 10.1.1.4:873
% ipvsadm -ln -t 10.1.1.4:873
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  10.1.1.4:873 wlc
% nc 10.1.1.4 873
(UNKNOWN) [10.1.1.4] 873 (rsync) : Connection refused
```

So it appears ipvs services play a role even when the IP is not local (like iptables) and services with zero endpoints cause traffic connecting from that node to have issues like above.

murali-reddy:
Ok, lets remove services with empty endpoints. 

But deleting the IPVS service in the cleanup phase is not sufficient as periodic sync will ends up creating them, though gets deleted in the end of sync.

So right now `syncIPVSServices` does below
- assign service VIP to the dummy interface
- setup any routing, PBR for DSR etc
- create IPVS service
- add endpoints as IPVS servers
- perform cleanup of any stale IPVS servers, services, PBR rules, vip's on dummy interface etc

so if service is marked with `service.local=true or externalTrafficPolicy: Local` and does not have any local endpoints, then perhaps we should skip in the very first step.

dlamotte:
Good point @murali-reddy.  I'll look to update this PR with that soon.

dlamotte:
I'm not sure I'm grokking everything going on in this loop that I'm skipping with this latest commit.  But I believe its what you're after, right @murali-reddy ?  I haven't tested this yet, but will do so after you give your OK.

murali-reddy:
>  I believe its what you're after, right @murali-reddy ? 

@dlamotte right. What should be the behaviour for services with external IP and load balancer IP associated, but service itself is marked with either `service.local=true or externalTrafficPolicy: Local`.

dlamotte:
Not sure.  ClusterIP, ExternalIPs, LoadBalancerIP and NodePorts should probably all not be created if the service is marked `local=true` and the node doesn't have endpoints running on it?

I will try testing this current fix on our cluster today.

dlamotte:
@murali-reddy been running this in the cluster today and appears to be ok.

PR #432, opened 2018-05-10
service: limit external IP routing to when the endpoint is considered active on the node

dlamotte:
Not sure about the `moq` and unittesting stuff.  I got them to pass, but I'm pretty new to that stuff.

PS: I am verifying these PRs in our production cluster before I PR them. 😄


dlamotte:
Forgot, this depends on #420.  That's why it fails.

aauren:
Closing as this is too old to merge any longer.

PR #442, opened 2018-05-16
adds explicit check for existing tunnel to avoid partial matches

rmenn:
The current stings.Contains passes checks when there is are similar IPs

Ex: 
the two ips exist `10.1.73.7` &  `10.1.73.72` where in the latter the tunnel will pass the condition for the former. This issue was encountered multiple times when there are similar IPs with the last octets sharing values. This has fixed the issue in our current environment.


murali-reddy:
@rmenn LGTM. Thanks for catching this issue and fixing this. Really appreciate for being persistent and fixing this.

rmenn:
if someone wants to test this, an image is available at `rmenn/kube-router:v0.1.3` please pardon the versioning

murali-reddy:
I just made a release just for this issue https://github.com/cloudnativelabs/kube-router/releases/tag/v0.2.0-beta.6 :)

PR #463, opened 2018-06-08
Health 1

roffe:
PR does:

Health controller starts first, every controller has it's own sync period as part of heartbeat timeout. since all controllers fire full sync the first time they start we measure how long it takes to get the first heartbeat and add that as a "buffer" time to the heartbeat timeout for each controller

the following are the default sync times

      --iptables-sync-period duration    The delay between iptables rule synchronizations (e.g. '5s', '1m'). Must be greater than 0. (default 5m0s)
      --ipvs-sync-period duration        The delay between ipvs config synchronizations (e.g. '5s', '1m', '2h22m'). Must be greater than 0. (default 5m0s)
      --routes-sync-period duration      The delay between route updates and advertisements (e.g. '5s', '1m', '2h22m'). Must be greater than 0. (default 5m0s)

a additional 1500 ms static time is added to this to this to allow for load variation or increase in services

so the controllers heartbeat timeout would be syncinterval + buffer + extra 1500 ms

on start the controllers would have syncintervall + extra 1500 ms to complete before being reported as unhealthy

### DIFF
--- a/pkg/cmd/kube-router.go
+++ b/pkg/cmd/kube-router.go
@@ -82,6 +82,13 @@ func (kr *KubeRouter) Run() error {
 
 	stopCh := make(chan struct{})
 
+	hc, err := healthcheck.NewHealthController(kr.Config)
+	if err != nil {
+		return errors.New("Failed to create health controller: " + err.Error())
+	}
+	wg.Add(1)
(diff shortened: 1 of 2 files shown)


roffe:
looks like make test flaked on the travis instance. weird

PR #465, opened 2018-06-13
Report delay metrics as seconds, not nanos

### LINKED ISSUE #459: Prometheus metrics should use base units
Prometheus (rightfully) recommends using base units for metrics instead of random units, as documented [here](https://prometheus.io/docs/practices/naming/#base-units). This simplifies consumption of the metrics as you no longer need to convert between all kinds of random units.

To point at one of the metrics [here](https://github.com/cloudnativelabs/kube-router/blob/1c7866cd91874670971c2944924dfa063ec038ce/pkg/controllers/netpol/network_policy_controller.go#L207-L211) is `kube_router_controller_iptables_sync_time` which seems to currently get reported as nanoseconds.

andor44:
Fixes #459

### DIFF
--- a/dashboard/kube-router.json
+++ b/dashboard/kube-router.json
@@ -480,7 +480,7 @@
           },
           "yaxes": [
             {
-              "format": "ns",
+              "format": "s",
               "label": "Time",
               "logBase": 1,
               "max": null,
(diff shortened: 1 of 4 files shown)


roffe:
Hi @andor44 and thanks for the PR, Could you please also include a update to the dashboard.

https://github.com/cloudnativelabs/kube-router/blob/master/dashboard/kube-router.json#L483

There should be 4 places where the `"format": "ns"` can be changed to `"format": "s"` to match your patch

besides that LGTM

Issue #461, opened 2018-06-07
Sporadic Connection Refused for services on Network Policy sync


johanot:
Some time ago I informally raised on slack that there might be an issue with the NetworkPolicyController causing intermittent connection failures while reconciling. I can now confirm that this happens, although it can be hard to re-create.

I think I narrowed it down to this chunk of code in `network_policy_controller.go`:

```golang
// TODO use iptables-restore to better implement the logic, than flush and add rules
err = iptablesCmdHandler.ClearChain("filter", policyChainName)
if err != nil && err.(*iptables.Error).ExitStatus() != 1 {
	return nil, nil, fmt.Errorf("Failed to run iptables command: %s", err.Error())
}

err = npc.processIngressRules(policy, targetDestPodIpSetName, activePolicyIpSets)
if err != nil {
	return nil, nil, err
}

err = npc.processEgressRules(policy, targetSourcePodIpSetName, activePolicyIpSets)
if err != nil {
	return nil, nil, err
}
```

I tried commenting out the ClearChain line and made sure that ingress/egress rules were only processed once (on kube-router startup) and now I couldn't re-produce the problem.

My guess is, that in some cases, incoming packets are dropped due to the time gap between chain clearing and chain re-building, although I'm not entirely sure. 
For sure is though, that I get intermittent Connection Refused issues when talking to services and they always occur at the exact time of network policy Sync().

## Reproduce

To reproduce the error, I run something like:
```sh
time while true ; do curl -s -o /dev/null http://myservice.default.svc.cluster.local/api || break ; sleep 0.1 ; done
```
**hint:** set `--iptables-sync-period` **low**, or apply some continuous changes to the cluster (triggering sync) in order not to wait forever to see the error. Perhaps you need to have a certain amount of network policies as well, in order for the gap between chain flush and rule creation to be noticeable.

## Suggestions
```golang
// TODO use iptables-restore to better implement the logic, than flush and add rules
```
a fix is already suggested in the above comment ^^. 
Although I don't know if this would provide better atomicity: 
1. Building a new chain
2. Changing the chain reference
3. (defer) Deleting the old chain

For the sake of atomicity/consistency between different NW policies. E.g. When service A talks to service B, A having an egress rule and B the corresponding ingress rule; perhaps it's better to rebuild _all_ chains in one loop and afterwards changing all references in a second loop? 

Of course best of all would be to not at all touch what haven't been changed. Perhaps iptables-restore can help on that.

[...]

PR #471, opened 2018-06-21
Fix for network policy connection refused issue (#461)

johanot:
* Instead of clearing the iptables firewall chains for each resync, new chains are now generated side-by-side with the existing ones.

* Chain naming now has an addition component, version, which ensures chain name uniqueness.

* Existing cleanup procedure for stale iptables rules will handle garbage collection of unused chains.


murali-reddy:
thanks @johanot 

LGTM

Issue #385, opened 2018-04-11
Allow binding specific IP addresses


andrenth:
Currently kube-router listens on all IP addresses for BGP, gRPC and health check ports:

```
# netstat -lnp | grep kube-router
tcp        0      0 0.0.0.0:179             0.0.0.0:*               LISTEN      4535/kube-router
tcp6       0      0 :::50051                :::*                    LISTEN      4535/kube-router
tcp6       0      0 :::179                  :::*                    LISTEN      4535/kube-router
tcp6       0      0 :::20244                :::*                    LISTEN      4535/kube-router
```

In multi-homed nodes, it can useful to be able to select a specific bind IP address, so that, for example, a private cluster becomes inaccessible from the public network by design, without the need for extra iptables rules.

murali-reddy:
@andrenth Good point, i guess specific IP that BGP server should listen on would node IP.

PR #473, opened 2018-06-21
issue-385: make it optional on which ip address BGP server listens

murali-reddy:
Modifying the fix done for https://github.com/cloudnativelabs/kube-router/issues/385 in commit https://github.com/cloudnativelabs/kube-router/commit/8c746b2cbc8026717d534236358033007fd7c9b0

Adding an option `bgp-listen-all-ip` to control on which IP GoBGP server listens on. By default value is false, and GoBGP server will just listen on node IP. When set to true GoBGP server will listen on `0.0.0.0`

This flexibility is desinrable when node is  multi-homed and GoBGP server can learn routes advertised from different networks.

cc @Thoro

### DIFF
--- a/docs/bgp.md
+++ b/docs/bgp.md
@@ -142,3 +142,13 @@ kubectl annotate node <kube-node> "kube-router.io/peer.ips=192.168.1.99,192.168.
 kubectl annotate node <kube-node> "kube-router.io/peer.asns=65000,65000"
 kubectl annotate node <kube-node> "kube-router.io/peer.passwords=U2VjdXJlUGFzc3dvcmQK,"
 ```
+
+## BGP listen address list 
+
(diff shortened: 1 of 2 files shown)


Thoro:
Thanks!

roffe:
i suggest to introduce bgp-listen-address instead that by default use node ip but allows the users to override with the interface of their chosing to allow for further config in corner case scenarios

murali-reddy:
@roffe makes sense. Will change the PR

murali-reddy:
I updated the PR to introudce new per node annotation `kube-router.io/bgp-local-addresses` which lets you to configure local IP addresses on which BGP server should listen.

By default (i.e with no node anntation configures) kube-rotuer continue to configure the GoBGP to listen on the node IP address only.

Issue #469, opened 2018-06-15
Support BGP Path Prepending


andrewsykim:
gobgp allows you to configure path prepends as part of the [policy definition](https://github.com/osrg/gobgp/blob/master/docs/sources/policy.md). Supporting path prepending can be useful since you can let cluster admins have better control of networking traffic by prepending paths (making the AS path longer, i.e. don't route to it unless all other nodes of shorter path fail).

PR #476, opened 2018-06-29
Add Support for BGP Prepend

es:
Fixes https://github.com/cloudnativelabs/kube-router/issues/469

For traffic shaping purposes, you may want to prepend the AS path announced to peers.
This PR enables that on a per-node basis with annotations:
- `kube-router.io/path-prepend.as`: AS number to be prepended
- `kube-router.io/path-prepend.repeat-n`: number of times to prepend ASN

### DIFF
--- a/docs/bgp.md
+++ b/docs/bgp.md
@@ -37,10 +37,10 @@ get peered.
 
 ### Route-Reflector setup  Without Full Mesh
 
-This model support the common scheme of using Route Reflector Server node to concentrate 
+This model support the common scheme of using Route Reflector Server node to concentrate
 peering from Client Peer. This has the big advantage of not needing full mesh, and
-scale better. In this mode kube-router expects each node is configured either in 
-Route Reflector server mode or in Route Reflector client mode. This is done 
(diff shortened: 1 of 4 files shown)


pkg/options/options.go
```diff
@@ -129,4 +130,5 @@ func (s *KubeRouterConfig) AddFlags(fs *pflag.FlagSet) {
 	// 	"Password that cluster-node BGP servers will use to authenticate one another when \"--nodes-full-mesh\" is set.")
 	fs.StringVarP(&s.VLevel, "v", "v", "0", "log level for V logs")
 	fs.Uint16Var(&s.HealthPort, "health-port", 20244, "Health check port, 0 = Disabled")
+	fs.BoolVar(&s.PathPrepend, "path-prepend", false, "tktk")
```
andrewsykim: Not sure we need this, we should just turn on path prepend if it's found in the annotations.

es:
PR ready for 👀

andrewsykim:
LGTM

Issue #477, opened 2018-07-01
New chains are not generated, irrespective of the added syncVersion 


johanot:
The fix for #461 introduced a syncVersion which is appended to the components that make up the unique firewall and policy chain names. Unfortunately, a int64 to string conversion has been implemented incorrectly such that chain names are always the same.

Since the ClearChain step has been removed, this mistake causes old chain rules never to get cleaned up, and thus changes to existing network policies never get reconciled correctly.

aauren:
This appears to have been fixed in #477 please re-open if this is not the case.

PR #478, opened 2018-07-01
Fix nwplcy re-sync issue (#477)

johanot:
Ref: #461 #477 

* use strconv for converting int64 to string

* change order of pod-fw sync, chain items has to be added before jumping to the chain starts

* added logging of syncversion, decreased logging verbosity+severity for planned chain cleanups

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -206,7 +206,7 @@ func (npc *NetworkPolicyController) Sync() error {
 	defer npc.mu.Unlock()
 
 	start := time.Now()
-	syncVersion := string(start.UnixNano())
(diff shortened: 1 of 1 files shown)


murali-reddy:
@johanot could you please give quick gist of why changing order is needed?

johanot:
@murali-reddy sure! The intent behind my previous fix was to generate new unique chain names at each sync.. But by chance I figured that it didn't work at all. Once I changed the string convertion, new chains were finally generated, but my original problem with sporadic connection refuse came back.

After some time I realized that the jump-rule to the firewall chains was inserted while the the chains were still empty. So I reorganized such that the chains get populated *before* they are "linked". Thus my conn refused problem disappeared again.

murali-reddy:
LGTM.

Thanks.

Issue #475, opened 2018-06-29
NetworkPolicy: Support including both podSelector and namespaceSelector in a NetworkPolicyPeer


voidengineer:
Hi,

Kubernetes 1.11 allows including both podSelector and namespaceSelector in a NetworkPolicyPeer (see PR https://github.com/kubernetes/kubernetes/pull/60452 and issue https://github.com/kubernetes/kubernetes/issues/58637).

Are there any changes required in kube-router to support this?

johanot:
Yes, changes are required.

The network_policy_controller currently assumes that the policy peers have _either_ namespaceSelector or podSelector. I'm working on a PR.

johanot:
@murali-reddy @voidengineer When you have time, let me know what you think of: https://github.com/cloudnativelabs/kube-router/pull/479

voidengineer:
LGTM

murali-reddy:
Fixed in https://github.com/cloudnativelabs/kube-router/pull/479

PR #479, opened 2018-07-02
Implemented the use of both namespaceSelector and podSelector

johanot:
Ref: #475 

* Moved code for evaluation of policy peers into separate func to avoid code duplication

* Ensured fallback to policy namespace, if namespaceSelector is not set

- Tested on kubernetes v1.11.0 with combinations of
namespaceSelector alone
podSelector alone 
namespaceSelector + podSelector


murali-reddy:
LGTM

murali-reddy:
Thanks @johanot

Issue #492, opened 2018-07-20
Allow for use of non-standard BGP-port internally and externally


johanot:
### Usage Scenario

I have a router box that runs bird for peering with other parties outside the cluster. I want the same server to peer with my k8s cluster using kube-router. Thus, I get a socket ip:port clash. 

Bird and other BGP clients/servers support specifying a custom remote and listening port for BGP. 

### Request

It should be possible to select a custom listening port for the BGP-server as well as a custom remote port for external BGP peers.

### Implementation

I will soon have a PR ready for discussion.

aauren:
Closed through #493

PR #493, opened 2018-07-20
Added support for custom BGP ports with 179 still being default (#492)

johanot:
* Introduced new cmdline flag --bgp-port, which controls BGP Server listening port and remote port of in-cluster node peers

* Introduced new cmdline flag --peer-router-ports, which controls remote BGP port for external peers

* Introduced new node annotation kube-router.io/peer.ports with same effect as --peer-router-ports

Note: Although it is now possible to specify a custom remote port for external peers, still only **one** BGPServer will be started locally, listening on the port specified with the `--bgp-port` flag. 
Hence; in case you are using a custom `--bgp-port` for in-cluster peering and different `--peer-router-ports` for external peering, the external peering connection becomes one-way (passive).

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -31,6 +31,7 @@ Usage of kube-router:
       --advertise-loadbalancer-ip        Add LoadbBalancer IP of service status as set by the LB provider to the RIB so that it gets advertised to the BGP peers.
       --advertise-pod-cidr               Add Node's POD cidr to the RIB so that it gets advertised to the BGP peers. (default true)
       --bgp-graceful-restart             Enables the BGP Graceful Restart capability so that routes are preserved on unexpected restarts
+      --bgp-port uint16                  The port open for incoming BGP connections and to use for connecting with other BGP peers. (default 179)
       --cache-sync-timeout duration      The timeout for cache synchronization (e.g. '5s', '1m'). Must be greater than 0. (default 1m0s)
       --cleanup-config                   Cleanup iptables rules, ipvs, ipset configuration and exit.
       --cluster-asn uint                 ASN number under which cluster nodes will run iBGP.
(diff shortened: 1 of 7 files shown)


murali-reddy:
Thanks @johanot .

Overall changes look good to me. 

Does the delete of peer work if there is mismatch in the config here https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/routing/bgp_peers.go#L163?

johanot:
@murali-reddy That's a good question. Haven't tested that. Will try to do some more testing later.

murali-reddy:
@johanot can you please rebase to latest master. Also would you please confirm if delete works fin

johanot:
@murali-reddy Rebase done. Will test deletion tomorrow.

johanot:
@murali-reddy I had to redo my rebase, as I accidently nuked **/pkg/controllers/proxy/network_services_controller_moq.go** as part of the first rebase. Didn't think auto-generated files was supposed to be committed. Anyway, that's fixed now.

I tested removal of BGP-neighbours on our prod cluster today. It works. 

One could actually argue, that this is actually a bug in GoBGP, and that it shouldn't work?
https://github.com/osrg/gobgp/blob/840807a520732ab09e83709382f7da1dc4a22ed0/pkg/server/peer.go#L55
^^ the deletion happens based on the `NeighborAddress` only, the port is not taken into account. However I won't jump into a GoBGP discussion here :-)

murali-reddy:
thanks @johanot 

>One could actually argue, that this is actually a bug in GoBGP, and that it shouldn't work?
https://github.com/osrg/gobgp/blob/840807a520732ab09e83709382f7da1dc4a22ed0/pkg/server/peer.go#L55
^^ the deletion happens based on the NeighborAddress only, the port is not taken into account. However I won't jump into a GoBGP discussion here :-)

Logically its address what matters I guess, but just wanted to check if delete would fail if port is taken into consideration.

PR #498, opened 2018-07-31
Improve health check for cache synchronization

### LINKED ISSUE #489: WaitForCacheSync method is too slow to lead to unhealthy judgment
In some cases, when the WaitForCacheSync method is too slow and the --iptables-sync-period value is too small,the CheckHealth method will return false, which will eventually cause k8s to restart kube-router circularly. 
In fact,the CheckHealth method should take into account the initialization process including WaitForCacheSync.

jimmy-zh:
fix #489
Improve health check for cache synchronization. See the issue above for more details.

### DIFF
--- a/pkg/cmd/kube-router.go
+++ b/pkg/cmd/kube-router.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"time"
 )
 
 // These get set at build time via -ldflags magic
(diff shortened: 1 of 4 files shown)


pkg/cmd/kube-router.go
```diff
@@ -177,6 +182,22 @@ func (kr *KubeRouter) Run() error {
 	return nil
 }
 
+// CacheSync performs cache synchronization under timeout limit
+func (kr *KubeRouter) CacheSync(informerFactory informers.SharedInformerFactory, stopCh <-chan struct{}) error {
```
andrewsykim: prefer `CacheSyncOrTimeout`

pkg/cmd/kube-router.go
```diff
@@ -177,6 +182,22 @@ func (kr *KubeRouter) Run() error {
...
+	syncOverCh := make(chan struct{})
+	go func() {
+		informerFactory.WaitForCacheSync(stopCh)
+		close(syncOverCh)
+	}()
+
+	select {
+	case <-time.After(kr.Config.CacheSyncTimeout):
+		return errors.New("timeout")
```
andrewsykim: Can this error include the timout time we waited for?

pkg/healthcheck/health_controller.go
```diff
@@ -107,7 +107,6 @@ func (hc *HealthController) CheckHealth() bool {
 	graceTime := time.Duration(1500 * time.Millisecond)
 
 	if hc.Config.RunFirewall {
-
```
andrewsykim: 👍

pkg/healthcheck/health_controller.go
```diff
@@ -182,17 +181,72 @@ func (hc *HealthController) Run(healthChan <-chan *ControllerHeartbeat, stopCh <
 
 }
 
+//RunServer starts the HealthController's server
+func (hc *HealthController) RunServer(stopCh <-chan struct{}, wg *sync.WaitGroup) error {
```
andrewsykim: Is this suppose to replace `Run`?

jimmy-zh: Yes. Actually I split Run into RunServer and RunCheck. Put the former before CacheSyncOrTimeout and the latter after CacheSyncOrTimeout.

andrewsykim: I see, should we delete `Run` then? I still see it in this file

jimmy-zh: I have deleted it

andrewsykim:
@roffe mind taking a look at this PR, a few changes with health check controller.

pkg/options/options.go
```diff
@@ -50,6 +51,7 @@ type KubeRouterConfig struct {
 
 func NewKubeRouterConfig() *KubeRouterConfig {
 	return &KubeRouterConfig{
+		CacheSyncTimeout:   5 * time.Minute,
```
roffe: I think we should have a lower value here something like 1 minute

pkg/cmd/kube-router.go
```diff
@@ -76,24 +77,40 @@ func CleanupConfigAndExit() {
...
+	podInformer := informerFactory.Core().V1().Pods().Informer()
+	nodeInformer := informerFactory.Core().V1().Nodes().Informer()
+	nsInformer := informerFactory.Core().V1().Namespaces().Informer()
+	npInformer := informerFactory.Networking().V1().NetworkPolicies().Informer()
+	informerFactory.Start(stopCh)
+
+	err = kr.CacheSyncOrTimeout(informerFactory, stopCh)
+	if err != nil {
+		return errors.New("Failed to synchronize cache: " + err.Error())
 	}
```
roffe: i think we can move 
```
	hc, err := healthcheck.NewHealthController(kr.Config)
	if err != nil {		if err != nil {
		return errors.New("Failed to create health controller: " + err.Error())			return errors.New("Failed to create health controller: " + err.Error())
	}		}
	wg.Add(1)		wg.Add(1)
	go hc.Run(healthChan, stopCh, &wg)		go hc.RunServer(stopCh, &wg)
```

to after the pod cache has finished

roffe: I think we can move 
```
	hc, err := healthcheck.NewHealthController(kr.Config)
	if err != nil {
		return errors.New("Failed to create health controller: " + err.Error())
	}
	wg.Add(1)
	go hc.RunServer(stopCh, &wg)
```

so it doesn't start the health controller until after the pod cache is synced since failing to sync the podcache will halt kube-router and force a restart

pkg/options/options.go
jimmy-zh: Well, as far as I can see, 1 minute is really enough.

roffe:
LGTM

andrewsykim:
Thanks @jimmy-zh!

PR #502, opened 2018-08-05
Introduces the option --override-nexthop, to override the next hop used in advertised routes

### LINKED ISSUE #480: Allow setting next-hop when multiple peer routers are defined
As discussed on [#kube-router slack.](https://kubernetes.slack.com/archives/C8DCQGTSB/p1530560213000411)  We should be able to define the next-hop for a given peer router in the `--peer-router-ips` list or `kube-router.io/peer.ips` annotation.  

One possible use-case for this would be for resiliency:
```
NodeA → Interface1 (10.0.1.1) → Router (10.1.0.1)
    ↓                            ↑                   
     →  Interface2 (10.0.2.1) →  ↑
```
\* The above assumes the same ASN per interface with all other attributes the same except next-hop

murali-reddy:
Introduces the option --override-nexthop, setting it to true will make advertised next hop in the router to the BGP peers will be automatically selected to be appropriate reachable local IP. This will be override any next-hop set for the routes in the RIB. Kube-router by default sets the next-hop to `node IP` which is not correct in case of nodes with multiple interfaces and use different interaces for different external peers.

Fixes #480

### DIFF
--- a/docs/bgp.md
+++ b/docs/bgp.md
@@ -152,3 +152,7 @@ Here is sample example to make GoBGP server to listen on multiple IP address
 kubectl annotate node ip-172-20-46-87.us-west-2.compute.internal "kube-router.io/bgp-local-addresses=172.20.56.25,192.168.1.99"
 ```
 
+## Overriding the next hop
+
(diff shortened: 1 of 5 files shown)


murali-reddy:
Merging based on my testing and testing done by @zohar

PR #504, opened 2018-08-06
Add the ability to announce additional static CIDRs

trevex:
We have several static CIDRs that are announce for each machine in our 3 layer CLOS topology. Because bird and kube-router have to peer with the same switches, we needed kube-router to announce additional CIDRs for us. 

This PR adds the ability to either specify additional CIDRs via the new `--static-cidrs` flag or by annotating nodes with `kube-router.io/static-cidrs=...`. Both take a comma separated list of CIDRs.

This PR adds:
* Flag and annotations to advertise static CIDRs
* Additional tests and documentation to accommodate the new feature

### DIFF
--- a/docs/bgp.md
+++ b/docs/bgp.md
@@ -142,3 +142,17 @@ kubectl annotate node <kube-node> "kube-router.io/peer.ips=192.168.1.99,192.168.
 kubectl annotate node <kube-node> "kube-router.io/peer.asns=65000,65000"
 kubectl annotate node <kube-node> "kube-router.io/peer.passwords=U2VjdXJlUGFzc3dvcmQK,"
 ```
+
+## Additional BGP features
+
+### Advertising additional static CIDRs to routers
+
+If kube-router should announce additional routes and CIDRs, that a local bird instance might take care of in other use-cases. It is possible to specify these either by using the command line flag:
+```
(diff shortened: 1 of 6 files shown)


murali-reddy:
@trevex thanks for you PR. Changes look good to me. Thanks for writing tests as well.

Just for general understanding, what could be the use-cases where a node would like announce specific CIDR's to the external peers?

trevex:
@murali-reddy In our case we have a 3 layer CLOS network topology and need to take care of a lot of network grunt work on the nodes. For this we use bird, but bird and gobgp clash for obvious reasons. Therefore we thought, well lets just make kube-router take care of everything.

That being said we are currently investigating a different approach, that might be even more feasible since less changes to kube-router are necessary and therefore this PR should probably be on hold for now until the second option was investigated.

(The alternative option is to setup a "localhost AS", so kube-router in fact connects to its local machine instead [see [this](https://metallb.universe.tf/configuration/romana/#configure-romana) for a similar setup, this `--bgp-port` PR is required to accomplish this though]).

As always I'll get back to you once I know more.

trevex:
(Repeating what I stated on slack) The alternative to use a "localhost AS" similar to the setup described [here](https://metallb.universe.tf/configuration/romana/) proved to be more stable, because the pivot from bird to kube-router was unreliable.

PR #508, opened 2018-08-08
update BGP export policies on endpoints add event

andrewsykim:
There's an edge case for services using `externalTrafficPolicy: Local` where the export policies aren't updated for up to 5m (default routes sync period) because the first Service update event fails because the corresponding endpoint was not created. This is a race condition between kube-router & the apiserver.
```
E0808 18:11:30.708175       1 ecmp_vip.go:92] error getting routes for service: nginx, err: endpoint resource doesn't exist for service: "nginx"
```
^ that log message is from [here](https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/routing/ecmp_vip.go#L92) right before the call to `nrc.addExportPolicies()`. 

This PR adds a `OnEndpointsAdd` event handler to call `addExportPolicies` in addition to `OnEndpointsUpdate` when endpoint resources are created.


andrewsykim:
Still testing this, do not merge :)

murali-reddy:
LGTM

PR #514, opened 2018-08-10
[jjo] ipAddrDel(): also delete VIP local rt addition

### LINKED ISSUE #513: local rt addition not undone when VIP is removed from node
As a consequence of #433, when a VIP is removed from the node,
this previously added VIP to the `local` rt sticks there, voiding
that node and its hosted Pods from reaching the VIP elsewhere.

This is pretty noticeable if using `externalTrafficPolicy: Local`, as
a service's Pod moving off a node will void other (local) service
consumers from contacting it, this is indeed my use case (together
with latest `metallb` in `L2` mode).

Work-around: after the svc Pod leaves the previously hosting node:
```
ip r del <extn_svc_IP> dev kube-dummy-if ta local
```

jjo:
Fixes #513


jjo:
(travis-ci build failed on transit error, triggering a rebuild)

murali-reddy:
LGTM

PR #518, opened 2018-08-15
prevent IPIP tunnel creation when --override-nexthop=true

murali-reddy:
When --override-nexthop=ture next hop that will be advertised depends on the local IP used and the peer. So it does not necessarily be `node IP`. Hence the check 
```go
if !nrc.nodeSubnet.Contains(nexthop) && !nrc.overrideNextHop {
    if !nrc.enableOverlays {
```


murali-reddy:
Merging based on testing. There is no change to default behaviour i.e) create IPIP tunnels when nodes are in different subnets and overlay is enabled.

Only in case of `--override-nexthop=true` tunnels wont be created.

PR #522, opened 2018-08-29
Healthcheck & Dashboard update

roffe:
* added configurable static timeout for the controllers, defaults to 30 sec
* added metric `controller_errors` with label "controller"
* changed metric `controller_bgp_advertisements_received` to `controller_bgp_advertisements` with label "type" `sent` or `advertised`


docs/user-guide.md
```diff
@@ -46,7 +46,9 @@ Usage of kube-router:
   -h, --help                             Print usage information.
       --hostname-override string         Overrides the NodeName of the node. Set this if kube-router is unable to determine your NodeName automatically.
       --iptables-sync-period duration    The delay between iptables rule synchronizations (e.g. '5s', '1m'). Must be greater than 0. (default 5m0s)
+      --iptables-sync-timeout duration   The timeout for iptables rule synchronizations (e.g. '5s', '1m'). Must be greater than 0. (default 30s)
```
andrewsykim: we should specifically mention that it will cause health endpoint to return 500

pkg/cmd/kube-router.go
```diff
@@ -17,11 +17,12 @@ import (
 	"github.com/cloudnativelabs/kube-router/pkg/options"
 	"github.com/golang/glog"
 
+	"time"
```
andrewsykim: This should go to the very top with the rest of stdlib

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -19,14 +19,15 @@ import (
 	"github.com/golang/glog"
 	"github.com/prometheus/client_golang/prometheus"
 
+	"regexp"
```
andrewsykim: same here, group with other stdlib

pkg/healthcheck/health_controller.go
```diff
@@ -170,7 +153,7 @@ func (hc *HealthController) RunServer(stopCh <-chan struct{}, wg *sync.WaitGroup
 
 //RunCheck starts the HealthController's check
 func (hc *HealthController) RunCheck(healthChan <-chan *ControllerHeartbeat, stopCh <-chan struct{}, wg *sync.WaitGroup) error {
-	t := time.NewTicker(5000 * time.Millisecond)
+	t := time.NewTicker(1000 * time.Millisecond)
```
andrewsykim: With the new timeout, does it still make sense to lower this value?

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -118,36 +121,37 @@ type protocolAndPort struct {
...
-		err := npc.Sync()
-		if err != nil {
-			glog.Errorf("Error during periodic sync of network policies in network policy controller. Error: " + err.Error())
-			glog.Errorf("Skipping sending heartbeat from network policy controller as periodic sync failed.")
-		} else {
-			healthcheck.SendHeartBeat(healthChan, "NPC")
-		}
+	err := npc.Sync()
+	if err != nil {
+		metrics.ControllerErrors.WithLabelValues("NPC").Inc()
```
andrewsykim: Can we make "NPC" a constant?

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -928,7 +941,7 @@ func cleanupStaleRules(activePolicyChains, activePodFwChains, activePolicyIPSets
 		for podFwChain := range activePodFwChains {
 			podFwChainRules, err := iptablesCmdHandler.List("filter", podFwChain)
 			if err != nil {
-
+				glog.Errorf("Failed to list iptable filters")
```
andrewsykim:
```go
glog.Errorf("Failed to list iptable filters: %s", err)
```

andrewsykim: should this have `metrics.ControllerErrors.WithLabelValues(networkPolicyControllerLabel).Inc()`?

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -296,7 +300,7 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
 			glog.Errorf("Error during periodic ipvs sync in network service controller. Error: " + err.Error())
 			glog.Errorf("Skipping sending heartbeat from network service controller as periodic sync failed.")
 		} else {
-			healthcheck.SendHeartBeat(healthChan, "NSC")
+			healthcheck.SendHeartBeat(nsc.healthChan, networkServiceControllerLabel)
```
andrewsykim: IPVS default sync period is greater than the default sync timeout here. Is that okay? Maybe I'm missing something

andrewsykim: Ahh I see, we do sync period + timeout for health checks

pkg/metrics/metrics_controller.go
```diff
@@ -101,16 +101,21 @@ var (
 		Name:      "controller_bgp_internal_peers_sync_time",
 		Help:      "Time it took to sync internal bgp peers",
 	}, []string{})
-	ControllerBGPadvertisementsReceived = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+	ControllerBGPadvertisements = prometheus.NewCounterVec(prometheus.CounterOpts{
```
andrewsykim: nice!

pkg/metrics/metrics_controller.go
```diff
@@ -101,16 +101,21 @@ var (
...
 		Namespace: namespace,
-		Name:      "controller_bgp_advertisements_received",
-		Help:      "Time it took to sync internal bgp peers",
-	}, []string{})
+		Name:      "controller_bgp_advertisements",
+		Help:      "Number of BGP advertisements received and sent",
```
andrewsykim: small nit: can we make this `received and advertised` to be consistent with code?

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -354,7 +377,7 @@ func (nrc *NetworkRoutingController) advertisePodRoute() error {
 		subnet), false, attrs, time.Now(), false)}); err != nil {
 		return fmt.Errorf(err.Error())
 	}
-
+	metrics.ControllerBGPadvertisements.WithLabelValues("sent").Inc()
```
andrewsykim: Does this need a `if nrc.MetricsEnabled` check?

pkg/controllers/routing/ecmp_vip.go
```diff
@@ -25,7 +26,7 @@ func (nrc *NetworkRoutingController) bgpAdvertiseVIP(vip string) error {
 
 	_, err := nrc.bgpServer.AddPath("", []*table.Path{table.NewPath(nil, bgp.NewIPAddrPrefix(uint8(32),
 		vip), false, attrs, time.Now(), false)})
-
+	metrics.ControllerBGPadvertisements.WithLabelValues("sent").Inc()
```
andrewsykim: does this need to check for `nrc.MetricsEnabled`?

pkg/controllers/routing/ecmp_vip.go
```diff
@@ -47,6 +48,7 @@ func (nrc *NetworkRoutingController) advertiseVIPs(vips []string) {
 		if err != nil {
 			glog.Errorf("error advertising IP: %q, error: %v", vip, err)
 		}
+		metrics.ControllerBGPadvertisements.WithLabelValues("sent").Inc()
```
andrewsykim: does this need to check for `nrc.MetricsEnabled`?

pkg/metrics/metrics_controller.go
roffe: ofc

pkg/controllers/routing/ecmp_vip.go
roffe: as we discussed on Slack i think we can just have these as they are as the metrics are instanciated even if metrics controller is disabled so no errors are triggered and we don't need to have 2 additional lines of code per metric checking if metrics are enabled

roffe: i procceded and removed the extra if statements and config variable where it didn't make sense to have it any more. i left the one in the service controller since there is a function there that uses quiet alot of ipvs calls when doing sync just to get the metrics

pkg/controllers/proxy/network_services_controller.go
roffe: Would it make sense to implement a context for the sync() functions that is based of the timeout to enforce it as well? could prevent bugs in the syncs that prevents it from exiting

murali-reddy:
@roffe could you please rebase to latest master?

jrthrawny:
We have encountered some cases where our nodes have not properly reloaded their iptable rules and this went silently undetected for some time. We looked through the logs and found errors such as:

 `E1206 22:10:16.636454       1 network_routes_controller.go:335] Failed to inject routes due to: no such process`

I am curious if this PR could help us detect this and self heal as it would cause that effected router pod to fail a health check and get restarted. If it does, I am happy to help get this PR merged in.

roffe:
@jrthrawny no this pr would not adress that, it was mostly metrics and preventing a large rate of NSC.SYNC() to occur that was the motivation behind it but to much has changed in the codebase to salvage it now

PR #527, opened 2018-08-31
Make the comments of the iptables rules in NWPLCY chains more accurate and reasonable

jimmy-zh:
Make the comments of the iptables rules in NWPLCY chains more accurate and reasonable.


jimmy-zh:
@murali-reddy OK!

I don't think there will be duplicate iptables rules in the case of an upgrade, because we will have different network policy chain names, the new rules including the changed comments will belong to the new chains, and the old chains will be removed, right?

murali-reddy:
verified no redundent rules. Merging.

PR #529, opened 2018-09-04
Use ipset to manage multiple CIDRs in a network policy rule

jimmy-zh:
There are some advantages of using ipset to manage multiple CIDRs in a network policy rule:
1. Fewer iptables rules and better performance.
2. Lay the foundation for supporting the semantics of `except`. I will contribute another PR later to implement this feature.

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -31,6 +31,10 @@ import (
 
 const (
 	networkPolicyAnnotation = "net.beta.kubernetes.io/network-policy"
+	kubePodFirewallChainPrefix = "KUBE-POD-FW-"
+	kubeNetworkPolicyChainPrefix = "KUBE-NWPLCY-"
+	kubeSourceIpSetPrefix = "KUBE-SRC-"
+	kubeDestinationIpSetPrefix = "KUBE-DST-"
 )
 
 // Network policy controller provides both ingress and egress filtering for the pods as per the defined network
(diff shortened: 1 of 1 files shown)


roffe:
Hi @jimmy-zh and thank you for the PR!

How would these changes affect a existing running cluster with kube-router installed when performing a inline upgrade?

jimmy-zh:
I think it's just a change in the number of iptables rules, it should have no effect on functionality, or what else I didn't think about? Also trouble you check it out.

jimmy-zh:
I have verified these changes in my own cluster.

roffe:
Thanks for the feedback @jimmy-zh I will try to fit in testing this code in my existing staging cluster during the week!

jimmy-zh:
OK, thanks and wait for the result.

roffe:
So i built current master with this PR + https://github.com/cloudnativelabs/kube-router/pull/522 

( https://github.com/roffe/kube-router/tree/pr-529-health-pr-1 ) 

Image @ roffe/kube-router-git:v0.2.0-beta.4-64-gccd35b3c

Inline upgrade seems to have gone fine and the first hour has been trouble free so far 👍

roffe:
Might have talked to soon, it seems my services are starting to fail one by one now

roffe:
So it seems for some reason all of a sudden kubelet couldn't reach the kube-router pod anymore.

Have to investigate further if this is a combination of your + my PR or whats going on as my PR i have been running in staging a couple of days

roffe:
Ok, so what happened was that the iptables sync took over 30 seconds on the first start on my master node and that caused it to become unhealthy
never seen this happend before on any of my nodes

roffe:
it seems it only happened once though. if i delete the pods now it doesn't seem to happend

roffe:
I had to rollback my tests and is currently trying to recover my staging cluster

jimmy-zh:
That might seem odd. Your PR added extra timeout control for iptables synchronization, so now a more than 30 seconds synchronization time will cause a timeout, but the problem is that the timeout happens only once, not every time you redeploy kube-router?

I will also test further to see if that will recur.

roffe:
Did a new test upgrade in the weekend, Unfortunately the cluster goes into a defunct state with this PR, My current staging can't be down but will accept interuptions so i i can't test this PR properly

roffe:
1. Yes, i have current 0.2.0 running
2. I have no network policies whatsoever in the cluster
3. I updated the DS image to point to a new one i built of your PR
4 not needed
5. Old pods gets replaced one by one in a rolling update. kube-router comes up and reports healthy but overlay network breaks down and i can't access services etc

murali-reddy:
AFAICS changes are very contained to modifying iptables rules in network policy chains only. But lets wait for @roffe to confirm if there is any side-affect.

murali-reddy:
@roffe if you have no objection, can we merge this?

roffe:
I won't be able to test until next week when i'm back in to the office. My initial tests broke my cluster so from my side it's a no go until someone else verifies it actually works with inline updating kube-router with this PR without having to reboot or clearing the iptables

sohel2020:
@roffe could you please check this and report. This is a vital feature for us, so we're waiting to release.

murali-reddy:
> My initial tests broke my cluster so from my side it's a no go until someone else verifies it actually works with inline updating kube-router 

@roffe I did upgrade testing which went fine for me. But lets wait till you get a chance to confirm.

roffe:
I've done some tests and it seems to be ok. Might have been something funk lingering on all the nodes since previous versions or similar... Either way I can't reproduce the troubles that happened last time so go ahead

jimmy-zh:
@murali-reddy I have reset the branch, please check, thanks.

jimmy-zh:
@sohel2020  It should be completed within this week.

sohel2020:
@jimmy-zh  awesome. :)

roffe:
> Any timeline to submit PR to except ipBlock? we badly needed this feature.

You are free to contribute as well if there is features you would like to see in kube-router

sohel2020:
@roffe I wish I could. But I'm not a go developer. :(

PR #541, opened 2018-09-22
Add CLI option to toggle disabling of source-dest-check in EC2

rjosephwright:
When the number of nodes in a cluster is high enough, the
`disableSourceDestinationCheck()` logic creates a high number
of requests to EC2, resulting in throttling and subsequent
problems, such as the inability to attach EBS volumes. This is
not necessarily mitigated by the `ec2IamAuthorized` attribute
which was added to overcome this issue, as the number of
requests can still be high enough to reach Amazon's request
limits. In addition, it is not necessary to run this multiple
times in a loop for all the nodes in a cluster, as it is
sufficient to set it once when an instance boots.

This CLI option allows an administrator to turn off this
feature for kube-router so they can use some other means of
setting the attribute.


murali-reddy:
thanks for the pr @rjosephwright 

Make sense for kube-router not do disable src-dst check forcefully

rjosephwright:
Thanks @murali-reddy

PR #545, opened 2018-09-28
Ipv6; BGP peering

uablrek:
## BGP peering in a ipv6-only cluster

This PR takes ipv6 support as far as bgp peering for nodes inside the cluster;

```
# gobgp neighbor
Peer                AS  Up/Down State       |#Received  Accepted
1000::1:c0a8:101 64512 00:00:37 Establ      |        0         0
1000::1:c0a8:102 64512 00:00:37 Establ      |        0         0
1000::1:c0a8:103 64512 00:00:40 Establ      |        0         0
```

**NOTE**; with this PR ipv6 code is actually used if the nodeIP is recognised as an ipv6 address. Before ipv4 operation was hard-coded.

The CIDR handling does not work yet so the veth devices does not get any (ipv6) addresses. The `kube-bridge` device is howerver assigned an ipv6 address;

```
# ip -6 addr show dev kube-bridge
11: kube-bridge: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 state UP qlen 1000
    inet6 1000::2:b00:301/120 scope global 
       valid_lft forever preferred_lft forever
    inet6 fe80::6409:c0ff:fe65:9419/64 scope link 
       valid_lft forever preferred_lft forever
```

### Design notes

The `iptablesCmdHandler` is now created with protocol;

```
iptables.NewWithProtocol(iptables.ProtocolIPv6)
```

A function for creating the right iptablesCmdHandler is provided to avoid code duplication. The `PodEgressRules` funtions are now methods in `NetworkRoutingController` to be able to access ipv6 stuff.

#### The ipset 31-caracter name limitation

This is a problem. The `Refresh` function added "-temp" to the name which together with the "inet6:" prefix made the temp names too long. As a temporary solution just a "-" is added.
IMO some hash string like "tmp_d2d362cdc6579390f1c0617d7" shall be used for temporary ipset's to make the temp name independent from the original set name.

Please comment on this.

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -60,6 +60,7 @@ type NetworkRoutingController struct {
 	nodeName                string
 	nodeSubnet              net.IPNet
 	nodeInterface           string
+	isIpv6                  bool
 	activeNodes             map[string]bool
 	mu                      sync.Mutex
 	clientset               kubernetes.Interface
@@ -151,7 +152,7 @@ func (nrc *NetworkRoutingController) Run(healthChan chan<- *healthcheck.Controll
 	}
 
 	glog.V(1).Info("Performing cleanup of depreciated rules/ipsets (if needed).")
-	err = deleteBadPodEgressRules()
+	err = nrc.deleteBadPodEgressRules()
 	if err != nil {
 		glog.Errorf("Error cleaning up old/bad Pod egress rules: %s", err.Error())
 	}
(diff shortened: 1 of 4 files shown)


uablrek:
Took a closer look and the CIDR allocation *does* work.

```
# jq . < /etc/cni/net.d/10-kuberouter.conf  | cat
{
  "bridge": "kube-bridge",
  "ipam": {
    "subnet": "1000::2:b00:100/120",
    "type": "host-local"
  },
  "isDefaultGateway": true,
  "isGateway": true,
  "name": "ekvm",
  "type": "bridge"
}
```

It is routing to the node cids that fails. Calls to a service that happens to be local works.

murali-reddy:
LGTM.

Verified with https://github.com/Nordix/xcluster/blob/master/ovl/kube-router-ipv6/README.md

thanks @uablrek for the patch and helping out with testing

PR #555, opened 2018-10-18
Improved detect in ipv6IsEnabled()

uablrek:
The ipv6IsEnabled() function only checks if the kernel has ipv6 support, not if it is disabled with;

```
sysctl -w net.ipv6.conf.all.disable_ipv6=1
```

By trying to open a port on the standard looback address for ipv6 `[::1]:0` a disabled ipv6 is also detected.

This caused #155 to resurface.


murali-reddy:
LGTM

PR #558, opened 2018-10-22
add toleration for nodes with taint node.kubernetes.io/not-ready

### LINKED ISSUE #547: kube-router should tolerate node.kubernetes.io/not-ready
When spinning up a kubeadm cluster the master is tainted with node.kubernetes.io/not-ready:

`KubeletNotReady              runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni config load failed: no network config found in /etc/cni/net.d: cni plugin not initialized: failed to load cni config`

because kube-router cni isn't available yet. That prevents kube-router from installing a cni config in the first place; so it should tolerate the presence of `node.kubernetes.io/not-ready:NoSchedule`.

murali-reddy:
Fixes #547

### DIFF
--- a/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
+++ b/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
@@ -127,6 +127,9 @@ spec:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
         operator: Exists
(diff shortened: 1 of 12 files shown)


murali-reddy:
Manually verified. this is manifest only change.

PR #559, opened 2018-10-23
Standalone mode

raffaelespazzoli:
This patch is aimed to enable running kube-router outside of a kubernetes node - `standalone` mode.
In this mode only the service-proxy feature can be enabled.
This mode assumes that an external host, on which kube-router standalone runs, exposes kubernetes service endpoints to be load balanced to pods. We assume that routability between this host and the kuberneters pod's network is provided.
This mode also supports the case where multiple kube-router standalone instances run on the same host. This is to support the case where this host is the gateway to multiple kubernetes clusters. In this case each kube-router instance will own a specific service CIDR.
To enable this mode the following parameters have been added:
`--service-cidr string`: Defines the service cidr for which this instance of kube-router is responsible. Useful if running multiple instance of kube-router on the same host.
 `--standalone`: Whether kube-router is running on a kubernetes node or in standalone mode. Default is false. When this option is true only the service proxy can be run and standlone-hostname and stanalone-ip must be specified.
 `--standalone-hostname string`: Defines the hostname to be used in standalone mode, mandatory when standalone is true
 `--standalone-iface string`: Defines the interface to be used in standalone, mandatory when standalone is true
 `--standalone-ip string`: Defines the ip to be used in standalone mode in octects format, mandatory when standalone is true

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -65,6 +65,11 @@ Usage of kube-router:
       --run-firewall                     Enables Network Policy -- sets up iptables to provide ingress firewall for pods. (default true)
       --run-router                       Enables Pod Networking -- Advertises and learns the routes to Pods via iBGP. (default true)
       --run-service-proxy                Enables Service Proxy -- sets up IPVS for Kubernetes Services. (default true)
+      --service-cidr string              Defines the service cidr for which this instance of kube-router is responsible. Useful if running multiple instance of kube-router on the same host.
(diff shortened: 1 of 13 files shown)


aauren:
These changes are substantial for supporting a use-case that is not the primary focus of this project. While it does seem like it would be useful we're currently focusing on trying to better support our core mission as a kubernetes network provider.

Closing for now.

PR #560, opened 2018-10-23
infer endpoint is local from endpoints "subset.addresses.nodeName"

### LINKED ISSUE #557: `kube-router.io/service.local=true` breaks services that run on the nodes network stack
When using `kube-router.io/service.local=true` for a service that points to pods that run on the nodes network stack neither ipvs rules nor kube-dummy-if addresses are added. It seems like kube-router lacks logic to handle pod ips that are not in the nodePodCidr https://github.com/cloudnativelabs/kube-router/blob/v0.2.1/pkg/controllers/proxy/network_services_controller.go#L472

murali-reddy:
Fixes #557

this changes make service proxy in-line with routing controller 

https://github.com/cloudnativelabs/kube-router/blob/v0.2.1/pkg/controllers/routing/ecmp_vip.go#L381

and kube-proxy

https://github.com/kubernetes/kubernetes/blob/v1.12.0/pkg/proxy/endpoints.go#L221

in the way and endpoint is inferred if its local or not.

this works for both pod in podCIDR range and in host network as well

`cloudnativelabs/kube-router-git:issue-557` image with the fix if any one wishes to try


xnaveira:
We'll give this a try and let you know

msoderberg:
We tried with the new image but ran into the same problem. Now we get this in the kube-router logs: 
```
E1023 13:40:59.522649       1 ecmp_vip.go:92] error getting routes for service: dummy, err: endpoint resource doesn't exist for service: "dummy"
```

We do have an endpoint resource for the dummy service though:
```
kubectl get endpoints -n default                                                                                                                                                                                          
NAME         ENDPOINTS                                               AGE
dummy        192.168.51.11:80,192.168.59.10:80                       52m
```

murali-reddy:
thanks for testing it out @msoderberg 

> error getting routes for service: dummy, err: endpoint resource doesn't exist for service: "dummy"

Is it a transient error? You consistently see this? This error has nothing do with this patch in the PR.

You still see that pods that are in hostNetwork are not treated as local endpoints?

murali-reddy:
@msoderberg Thanks for testing out. I updated the patch. My earlier fix was not covering all cases.

Please try the updated image.

murali-reddy:
Meging based on self-testing  and testing done by @msoderberg

Issue #544, opened 2018-09-28
Apache Bench can fill up ipvs service proxy in seconds


neeseius:
I am not sure if I have something configured wrong but here is my Centos7 physical node and kube-router agent setup:

[ipvsadm package]
$ rpm -q ipvsadm
ipvsadm-1.27-7.el7.x86_64

[kube router process and options]
$ ps -ocommand= -C kube-router
/usr/local/bin/kube-router --run-router=true --run-firewall=true --run-service-proxy=true --kubeconfig=/etc/kubernetes/kube-router.kubeconfig --hostname-override=node6 --enable-overlay=true

[service]
kubectl get svc
NAME        TYPE           CLUSTER-IP       EXTERNAL-IP                            PORT(S)        AGE
test-svc    NodePort       172.30.176.114   <none>                               80:30530/TCP   7h

[ipvs]
$ ipvsadm -ln | head -n 1
IP Virtual Server version 1.2.1 (size=4096)

[ipvs service]
$ ipvsadm -ln | grep -A1 30530
TCP  10.200.1.146:30530 rr
  -> 172.32.9.68:80               Masq    1      0          0

If I use apache bench with tcp keepalive all is swell and absurdly fast, posting over 10,000 requests per second and ipvsadm will show stats like below during such a test:
$ ipvsadm -ln | grep -A1 30530
TCP  10.200.1.146:30530 rr
  -> 172.32.9.68:80               Masq    1      0          757

However if I run the same test without keep-alive then "InActConn" jumps up to 14000 within a few seconds and up until that point things are very fast, but after that point the virtual server just completely hangs up and stops responding to requests until "InActConn" drops back below 14000. This happens if I run apache bench on the node itself and hit the clusterIp, or if I run it from a random server and hit the nodeport.

---ipvs
$ ipvsadm -ln | grep -A1 30530
TCP  10.200.1.146:30530 rr
  -> 172.32.9.68:80               Masq    1      0          14115

--- apache bench output
ab -c 100 -n 20000 http://node6:30530
This is ApacheBench, Version 2.3 <$Revision: 1826891 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking node6 (be patient)
Completed 2000 requests
Completed 4000 requests
Completed 6000 requests
Completed 8000 requests
Completed 10000 requests
Completed 12000 requests
[...]

PR #562, opened 2018-10-23
Allow changing default behavior of the service.local configuration

msoderberg:
We use kube-routers routing and proxy features and have run into the problem that is described in detail in this issue: https://github.com/moby/moby/issues/35082#issuecomment-385433755

After a lot of experimentation and testing, we've come to the conclusion described by my colleague here: https://github.com/cloudnativelabs/kube-router/issues/544#issuecomment-431394829

To reduce complexity for us and for our users we'd like be able to set the default behavior of the service.local configuration, hence this PR. We understand that this might be a corner case but we want to avoid rolling our own version of kube-router if possible.


andrewsykim:
@msoderberg can you run a controller in your cluster that just adds the kube-router service.local annotation to all new services?

msoderberg:
@andrewsykim that's definitely an option, we'll look into that!

andrewsykim:
Recommend using initializers for this https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#initializers

msoderberg:
We tested with an initializer last Friday and that will work for us, closing this. Thanks for the help @andrewsykim!

PR #563, opened 2018-10-23
Added "--router-id=" parameter.

uablrek:
For ipv6 it is not possible to use the ip address as router-id
and this parameter is required.

There is discussions ongoing on how to handle router-id in an ipv6-only cluster.  For now at least the router-id must be specified using this command line parameter. This keeps the procedure to obtain a router-id outside the `kube-router` binary and allows ipv6 testing to continue.


murali-reddy:
LGTM

Issue #554, opened 2018-10-15
Support to set weight and reliable scheduler


inju-song:
Hi kube-router team.

We use the kube-router for networking in k8s cluster and external load balancing(I think this is the best network solution for k8s, thanks). However, we need to set weight to 0 or non-zero per k8s service and reliable scheduler to serve production k8s cluster using kube-router.

1. kube-router can be provided as several external load balancer but there is no reliable option between each load balancer as far as I know. If a node fails, maybe all connections inside the node are lost.
    - [MH](https://github.com/torvalds/linux/blob/master/net/netfilter/ipvs/Kconfig#L228) is a good solution I think. There is no need to setup extra high availability solution such as keepalived for reliability.
2. To serve service deployment strategies such as blue-green or canary, we need to set weight to zero or non-zero, but can't setup the weight per service. For example,
   - A older service(`blue`) gets 0 weight and A newer service(`green`) gets greater than or equal to 1 weight.

Already I developed features to set ipvs scheduler as MH through k8s svc annotation without modifying `libnetwork/ipvs` and have used for our k8s cluster. 

So, I'm wondering if the kube-router have plans to support to setup various weight and reliable scheduler such as MH scheduler.

murali-reddy:
@inju-song thanks for the report.

AFAIK, `maglev` is recent addition to Linux kernel? Kube-router would like to integrate a `consistent hashing` solution. This is something we wanted for current implementation of DSR (direct server return) in kube-rotuer as well. I have not fully explored `IP_VS_MH` yet.

Please see https://cloudnativelabs.github.io/post/2017-11-01-kube-high-available-ingress/

But kube-router does have support for older schedulers 

https://github.com/cloudnativelabs/kube-router/blob/master/docs/user-guide.md#load-balancing-scheduling-algorithms

If you have any thoughts on how `IP_VS_MH`  can be integrated into kube-router please share.

inju-song:
@murali-reddy 
Thanks for your reply.

Kube-router uses `libnetwork/ipvs`, but there is no `MaglevHashing` const variable. But we can just use `libnetwork/ipvs` functions to add or update `MH` virtual service without the const variable.

[...]

PR #564, opened 2018-10-25
Integrate ip_vs_mh scheduler into kube-router

inju-song:
Implement #554 

kube-router supports setting `ip_vs_mh` scheduler.
  -  usage: `kube-router.io/service.scheduler: mh`

`ip_vs_mh` provides two scheduler flags.
   - flag-1: destiantion fallback
   - flag-2: source hashing with address and **port**

(Please see [here](https://github.com/torvalds/linux/blob/master/net/netfilter/ipvs/ip_vs_mh.c) for `ip_vs_mh` details)

So I add a service annotation feature to set `ip_vs_mh` with the flags through kube-router.
   - usage: `kube-router.io/service.schedflags: flag-1` or `flag-1, flag-2`
   - Any combination of these flags are available .

If the `ip_vs_mh` is set with scheduler flags, you can see the ipvs service like below. 
```shell
TCP  172.24.3.177:80 mh (flag-1,flag-2)
```


murali-reddy:
thanks @inju-song for your work on kernel support for Maglev hashing and integration into kube-router.

Could you please point to me details on the kernel requirements and also how to test the hashing?

inju-song:
@murali-reddy 
`ip_vs_mh` is integrated into kernel since v4.18, so the required kernel version is v4.18 or later for using `ip_vs_mh`.

You can test `ip_vs_mh` by setting some annotations in kubernetes service like below.
```
For maglev hashing scheduling use:
kubectl annotate service my-service "kube-router.io/service.scheduler=mh"

For maglev hashing scheduling flags use:
kubectl annotate service my-service "kube-router.io/service.schedflags=flag-1, flag-2"
```

The flags can be set only when `mh` scheduler is set. If the scheduler is not set, setting flags will be ignored. Of course, only one flag can be set also.

It also supports tunnel forwarding mode, so we can use `ip_vs_mh` in dsr mode. However `ipvsAddFWMarkService()` [only can set `rr`](https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/proxy/network_services_controller.go#L1547) when dsr mode service created at first time, you should update `mh` schedule after setting `extrenalIP` and `dsr` annotation for using `mh` in tunnel mode.

PR #576, opened 2018-11-15
No reason to restrict Peer ASN's to private only. 

rillion:
This change is to restrict to public and private ranges. As per https://en.wikipedia.org/wiki/Autonomous_system_(Internet)#ASN_Table


murali-reddy:
thanks @rillion for your PR. LGTM, I dont see any reason to reject public ASN.

PR #577, opened 2018-11-18
make NSC set net.ipv4.vs.conn_reuse_mode=0

roffe:
will fix the IPVS low throughput issues

https://github.com/kubernetes/kubernetes/pull/71114

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -296,7 +296,7 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
 
 	err = ensureIpvsConnReuseMode()
 	if err != nil {
-		return errors.New("Failed to do sysctl net.ipv4.vs.conn_reuse_mode=0 due to: %s" + err.Error())
+		return fmt.Errorf("failed to set net.ipv4.vs.conn_reuse_mode=0: %s", err)
 	}
 
 	// loop forever unitl notified to stop on stopCh
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -1410,7 +1410,16 @@ func ensureIpvsConntrack() error {
 }
 
 func ensureIpvsConnReuseMode() error {
-	return ioutil.WriteFile("/proc/sys/net/ipv4/vs/conn_reuse_mode", []byte(strconv.Itoa(0)), 0640)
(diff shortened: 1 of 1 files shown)


pkg/controllers/proxy/network_services_controller.go
```diff
@@ -295,6 +294,11 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
 		return errors.New("Failed to do sysctl net.ipv4.vs.expire_quiescent_template=1 due to: %s" + err.Error())
 	}
 
+	err = ensureIpvsConnReuseMode()
+	if err != nil {
+		return errors.New("Failed to do sysctl net.ipv4.vs.conn_reuse_mode=0 due to: %s" + err.Error())
```
andrewsykim: nit:
```go
return fmt.Errorf("failed to set net.ipv4.vs.conn_reuse_mode=0: %s", err)
```

andrewsykim:
Do we need to check if `conn_reuse_mode` is available first? It seems like it was introduced in linux 4.1.

roffe:
@andrewsykim glansing over the code from K8S i can't rly say they do:

https://github.com/Lion-Wei/kubernetes/blob/f39060c6047b8dce5af197a97d5f18422d93c9d1/pkg/proxy/ipvs/proxier.go

roffe:
now we check and inform that it's a non fatal error if not there
if there is a error statting the file we also inform and continue as it's non-fatal

murali-reddy:
thanks @roffe

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -1405,6 +1409,19 @@ func ensureIpvsConntrack() error {
...
+	sysctlPath := "/proc/sys/net/ipv4/vs/conn_reuse_mode"
+	if _, err := os.Stat(sysctlPath); err != nil {
+		if os.IsNotExist(err) {
+			glog.Infof("%s not found, skipping setting net.ipv4.vs.conn_reuse_mode=0 (non fatal error, feature introduced into kernel in 4.1)", sysctlPath)
+			return nil
+		}
+		glog.Errorf("skipping setting net.ipv4.vs.conn_reuse_mode=0, error stating: %s : %s", sysctlPath, err.Error())
+		return nil
+	}
+	return ioutil.WriteFile("sysctlPath", []byte(strconv.Itoa(0)), 0640)
```
andrewsykim: `"sysctlPath"` -> `sysctlPath`

roffe:
i'm fixing it with a pr

PR #580, opened 2018-11-21
unified function to set sysctl values 

roffe:
and add https://github.com/kubernetes/kubernetes/pull/70530/files

### DIFF
--- a/pkg/utils/sysctl.go
+++ b/pkg/utils/sysctl.go
@@ -9,7 +9,7 @@ import (
 	"github.com/golang/glog"
 )
 
-func setSysctl(path string, value int) error {
+func SetSysctl(path string, value int) error {
 
 	sysctlPath := fmt.Sprintf("/proc/sys/%s", path)
 	if _, err := os.Stat(sysctlPath); err != nil {
--- a/pkg/utils/sysctl.go
+++ b/pkg/utils/sysctl.go
@@ -14,10 +14,10 @@ func SetSysctl(path string, value int) error {
 	sysctlPath := fmt.Sprintf("/proc/sys/%s", path)
 	if _, err := os.Stat(sysctlPath); err != nil {
 		if os.IsNotExist(err) {
-			glog.Infof("%s not found, could not set value: %s", sysctlPath)
+			glog.Infof("%s not found, could not set value: %s, Does your kernel support this feature?", sysctlPath)
 			return nil
 		}
-		glog.Errorf("skipping setting %s, error stating: %s", sysctlPath, err.Error())
+		glog.Errorf("error stating: %s : %s", sysctlPath, err.Error())
 		return nil
 	}
 	return ioutil.WriteFile(sysctlPath, []byte(strconv.Itoa(value)), 0640)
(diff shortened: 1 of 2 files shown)


pkg/utils/sysctl.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/golang/glog"
+)
+
+func SetSysctl(path string, value int) error {
+
+	sysctlPath := fmt.Sprintf("/proc/sys/%s", path)
+	if _, err := os.Stat(sysctlPath); err != nil {
+		if os.IsNotExist(err) {
+			glog.Infof("%s not found, could not set value: %s, Does your kernel support this feature?", sysctlPath)
```
andrewsykim: should we be returning err here instead of just logging?

pkg/utils/sysctl.go
```diff
@@ -0,0 +1,24 @@
...
+			return nil
+		}
+		glog.Errorf("error stating: %s : %s", sysctlPath, err.Error())
```
andrewsykim: same here, return err instead of log?

pkg/utils/sysctl.go
```diff
@@ -0,0 +1,24 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strconv"
+
+	"github.com/golang/glog"
+)
+
+func SetSysctl(path string, value int) error {
+
```
andrewsykim: nit: remove extra line here

pkg/utils/sysctl.go
andrewsykim: I feel like the caller of this method should be logging if anything. Maybe a new error type is needed if we don't want to "exit" on error?

pkg/utils/sysctl.go
roffe: some of these are non-fatal. f.x `conn_reuse_mode` where it does simply not exist in older kernels.

If i return a error this would exit nsc.Run

i'll add a third flag so we can say log or fail when calling SetSysctl for those values

andrewsykim: I was thinking more of having a new error type for non-fatal errors. The caller can check for that error and log or return err accordingly.

pkg/utils/sysctl.go
roffe: I don't understand what you are suggesting. could you provide some example of how this new error type would work?

pkg/utils/sysctl.go
roffe: how does one do a error-type for non-fatal errors? errors.New just gives me a error?

roffe:
@andrewsykim is this in line with what you was thinking?

andrewsykim:
Was just thinking we just create specific variables for those cases, e.g.
```go
var SysctlNotFoundErr := errors.New("sysctl not found")
...
...
```

But I'm realizing now that in Kubernetes core, they just ignore the error 🤷‍♂️

pkg/utils/sysctl.go
roffe: i got it and have done what you asked for :)

ticpu:
I don't know why no one seems to be affected by this, but since arp_ignore is set to 1, all our externalIPs stopped working until we echo 0 > arp_ignore.

PR #587, opened 2018-11-28
Updating deployment files for kubeadm

ventris:
Adding tolerations for key kubeadmNode.

This taint is added in kubeadm and makes the daemonset unable to install on the master nodes in a cluster installed by kubeadm. Without it the demonset will install on all nodes except the masters.

Tested on a cluster with Kubernetes version 1.12.2

### DIFF
--- a/daemonset/kube-router-firewall-daemonset.yaml
+++ b/daemonset/kube-router-firewall-daemonset.yaml
@@ -88,6 +88,9 @@ spec:
       - effect: NoSchedule
         key: node.kubernetes.io/not-ready
         operator: Exists
+      - effect: NoSchedule
+        key: kubeadmNode
+        operator: Exists
       volumes:
       - name: lib-modules
         hostPath:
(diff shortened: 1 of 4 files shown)


murali-reddy:
@ventris thanks for your PR. Can you please point me the documentation on this taint. I can not find proper documentation for this. We recently made a fix to support 1.12 https://github.com/cloudnativelabs/kube-router/commit/d9570c587263eb869f10139c380dffb30fd0d05b I did not face any issue.

ventris:
There is no real documentation for this change in kubeadm but it can be found in both [v1alpha3 ](https://github.com/kubernetes/kubernetes/blob/aae438e823e94c7b9a9e00504790da0ed97617d7/cmd/kubeadm/app/apis/kubeadm/v1alpha3/doc.go#L176) of the kubeadm api and [v1beta1](https://github.com/kubernetes/kubernetes/blob/25c9ac65447e8e53e7854d354653e79a925c3cd2/cmd/kubeadm/app/apis/kubeadm/v1beta1/doc.go#L179).

I cant find any information about this taint but it seems like only new installations of 1.12 get this taint on the master nodes.

ventris:
@murali-reddy After some more troubleshooting I have found that the issue is on our side and not in the application, the taint was included from the documentation and has no bearing on kubeadm.

Closing this issue as the change is no needed for kube-router to work.

murali-reddy:
@ventris good to know. thanks

Issue #584, opened 2018-11-27
Default route from eBGP peer, pulls kernel default upon withdraw/shutdown


sanjmonkey:
If an eBGP neighbor, say an upstream ToR, sends routes to kube-router which already exist in the kernel routing table, such as 0.0.0.0/0, and then the ToR either shuts down the session or withdraws 0/0, the node has its default route removed from kernel routing table, and is unreachable.

## Expected Behaviour
Kernel routing table to keep its configured default route, if it had one configured, when peering with external neighbors and the same route is received from them.

## Current Behaviour
Before eBGP session is established:
```
sandy.breeze@kube-node-uk02:~$ netstat -rn
Kernel IP routing table
Destination     Gateway         Genmask         Flags   MSS Window  irtt Iface
0.0.0.0         80.80.80.1      0.0.0.0         UG        0 0          0 ens160
10.10.2.0       0.0.0.0         255.255.255.0   U         0 0          0 kube-bridge
80.80.80.0      0.0.0.0         255.255.255.248 U         0 0          0 ens160
172.17.0.0      0.0.0.0         255.255.0.0     U         0 0          0 docker0

sandy.breeze@kube-node-uk02:~$ ip route list default
default via 80.80.80.1 dev ens160 proto static
```

Whilst eBGP session established and receiving 0.0.0.0/0 from neighbor:
```
sandy.breeze@kube-node-uk02:~$ netstat -rn
Kernel IP routing table
Destination     Gateway         Genmask         Flags   MSS Window  irtt Iface
0.0.0.0         80.80.80.1      0.0.0.0         UG        0 0          0 ens160
10.10.2.0       0.0.0.0         255.255.255.0   U         0 0          0 kube-bridge
80.80.80.0      0.0.0.0         255.255.255.248 U         0 0          0 ens160
172.17.0.0      0.0.0.0         255.255.0.0     U         0 0          0 docker0

sandy.breeze@kube-node-uk02:~$ ip route list default
default via 80.80.80.1 dev ens160 proto 17
````

After eBGP session is shutdown:
```
sandy.breeze@kube-node-uk02:~$ netstat -rn
Kernel IP routing table
Destination     Gateway         Genmask         Flags   MSS Window  irtt Iface
10.10.2.0       0.0.0.0         255.255.255.0   U         0 0          0 kube-bridge
80.80.80.0      0.0.0.0         255.255.255.248 U         0 0          0 ens160
[...]

PR #590, opened 2018-11-30
netlink.RouteReplace matches the kernels static and replaces it. 

sanjmonkey:
netlink.RouteAdd doesn't disrupt the static. netlink.RouteDel still handles the withdraw correctly for those inserted with RouteAdd, but keeps the kernels statics in tact

Fixes https://github.com/cloudnativelabs/kube-router/issues/584

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -454,7 +454,7 @@ func (nrc *NetworkRoutingController) injectRoute(path *table.Path) error {
 		return netlink.RouteDel(route)
 	}
 	glog.V(2).Infof("Inject route: '%s via %s' from peer to routing table", dst, nexthop)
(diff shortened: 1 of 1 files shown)


roffe:
LGTM

murali-reddy:
thanks @sanjmonkey for the PR. I agree route-replace is disruptive for data path.
But wonder netlink.RouteAdd is idempotent operation. `ip route add` fails if route exists. Best would be check if route exist and add only if its not present.

```
# ip route
default via 192.168.1.1 dev enp0s3 onlink 
10.244.0.0/24 dev cni0  proto kernel  scope link  src 10.244.0.1 
```
```
# ip route add 10.244.0.0/24 dev cni0
RTNETLINK answers: File exists

```

aauren:
Closing as author never responded to maintainer suggestions.

PR #594, opened 2018-12-06
docs/bgp.md: change example to use printf

aauren:
Using echo places a new-line character at the end of the base64 generated string which will cause the peering to fail for most use-cases as it is unlikely to be present in the configuration of the other peer.


roffe:
Hi @aauren `echo -n` has the same effect but `printf` works as well 👍

aauren:
@roffe Thanks for the quick merge! Yeah, I figured `printf` was a bit safer as `echo -n` isn't posix compliant and may not work on all systems.

PR #596, opened 2018-12-07
Change append to insert for iptables rules

eric:
Updates how iptables FORWARD rules are applied to accommodate an existing final DROP rule for the chain.

I am running CentOS 7.6 with firewalld enabled and found that all of the iptables entries that kube-router were adding were not taking effect due to where they were located.

This follows the convention of [docker's libnetwork](https://github.com/docker/libnetwork/blob/49627167f0585504fd78ed8827529aec57a9618d/iptables/iptables.go#L183-L226) and uses `-I` instead of `-A` for adding rules to the `FORWARD` chain.

For reference, the output from `iptables`:

```
# iptables -t filter -L FORWARD -v
Chain FORWARD (policy DROP 0 packets, 0 bytes)
    0     0 FORWARD_direct  all  --  any    any     anywhere             anywhere            
    0     0 FORWARD_IN_ZONES_SOURCE  all  --  any    any     anywhere             anywhere            
    0     0 FORWARD_IN_ZONES  all  --  any    any     anywhere             anywhere            
    0     0 FORWARD_OUT_ZONES_SOURCE  all  --  any    any     anywhere             anywhere            
    0     0 FORWARD_OUT_ZONES  all  --  any    any     anywhere             anywhere            
    0     0 DROP       all  --  any    any     anywhere             anywhere             ctstate INVALID
    0     0 REJECT     all  --  any    any     anywhere             anywhere             reject-with icmp-host-prohibited
    0     0 ACCEPT     all  --  kube-bridge any     anywhere             anywhere             /* allow outbound traffic from pods */
    0     0 ACCEPT     all  --  any    kube-bridge  anywhere             anywhere             /* allow inbound traffic to pods */
    0     0 ACCEPT     all  --  any    eno1    anywhere             anywhere             /* allow outbound node port traffic on node interface with which node ip is associated */
```

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -559,7 +559,7 @@ func (nrc *NetworkRoutingController) enableForwarding() error {
 		return fmt.Errorf("Failed to run iptables command: %s", err.Error())
 	}
 	if !exists {
-		err := iptablesCmdHandler.Insert("filter", "FORWARD", 0, args...)
+		err := iptablesCmdHandler.Insert("filter", "FORWARD", 1, args...)
 		if err != nil {
 			return fmt.Errorf("Failed to run iptables command: %s", err.Error())
 		}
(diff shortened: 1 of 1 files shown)


murali-reddy:
@eric Thanks for your PR. Which component is adding below rule?

```
 0     0 REJECT     all  --  any    any     anywhere             anywhere             reject-with icmp-host-prohibited
```
It overrides default policy `Chain FORWARD (policy DROP 0 packets, 0 bytes)`

Reason it was `Append` instead of `Insert` was to give network policies rules to give precedence. So there is conflict here.

eric:
It is added by firewalld on CENTOS by default.

PR #597, opened 2018-12-07
Update firewall rules via firewalld if it is available

eric:
Use the k8s iptables library to set iptables rules via firewalld if it is available (on CentOS for example).


asteven:
I think this PR is badly named. Sure, it enables the use of firewalld if it's available.
But IMHO the more important part is: re-use the upstream iptables module.

I don't care about firewalld - but reusing existing, well maintained code, is a plus.

@murali-reddy what say you?

aauren:
Closing this as it's no longer merge-able, but tagging @murali-reddy on it as I believe that using the upstream kubernetes iptables libraries may be worth looking into for the 1.2 release efforts.

PR #604, opened 2018-12-12
Add iptables input rules for ipvs services

### LINKED ISSUE #282: ExternalIP allows access to node
Since the external IPs are also added to the kube-dummy-if they now allow access to the node from outside of the cluster network, if the iptables rules don't forbid this.

I would suggest to add a DROP rule into the INPUT chain for all traffic going in on kube-dummy-if

### LINKED ISSUE #602: CentOS with firewalld creates a REJECT rule for INPUT chain
I've found that on CentOS 7.6.1810 with firewalld enabled, kube-router doesn't work right without manually adding an ACCEPT rule for any packets coming from the `kube-bridge` interface.

Here is my INPUT chain:

```
# iptables -v -L INPUT
Chain INPUT (policy ACCEPT 0 packets, 0 bytes)
  41M   12G KUBE-FIREWALL  all  --  any    any     anywhere             anywhere            
  40M   12G ACCEPT     all  --  any    any     anywhere             anywhere             ctstate RELATED,ESTABLISHED
 295K   18M ACCEPT     all  --  lo     any     anywhere             anywhere            
 738K  216M INPUT_direct  all  --  any    any     anywhere             anywhere            
 738K  216M INPUT_ZONES_SOURCE  all  --  any    any     anywhere             anywhere            
 738K  216M INPUT_ZONES  all  --  any    any     anywhere             anywhere            
 1006 40713 LOG        all  --  any    any     anywhere             anywhere             ctstate INVALID PKTTYPE = unicast LOG level warning prefix "STATE_INVALID_DROP: "
 1006 40713 DROP       all  --  any    any     anywhere             anywhere             ctstate INVALID
   24  1576 LOG        all  --  any    any     anywhere             anywhere             PKTTYPE = unicast LOG level warning prefix "FINAL_REJECT: "
 177K   60M REJECT     all  --  any    any     anywhere             anywhere             reject-with icmp-host-prohibited
```

Is there something that should be adding this rule automatically?

bazuchan:
Traffic coming to IPVS service first goes throught iptables INPUT chain. If iptables policy set to DROP and no INPUT rules added manually to access services they will be inaccessible. This commit adds iptables rule to access IPVS services.

Should fix #602 and somewhat fixes #282

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -56,6 +56,7 @@ const (
 	svcSchedFlagsAnnotation = "kube-router.io/service.schedflags"
 
 	LeaderElectionRecordAnnotationKey = "control-plane.alpha.kubernetes.io/leader"
+	svcIpSetName                      = "KUBE-SVC-ALL"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


murali-reddy:
LGTM.

asteven:
@bazuchan awesome! Many thanks!

Issue #605, opened 2018-12-12
gobgp complains about NextHop Attribute containing IPV6 address


Arvinderpal:
For IPv6 route advertisements, MP-BGP extensions should be used. Recent PR https://github.com/cloudnativelabs/kube-router/pull/578 does this; however it also inserts the IPV6 address in the NEXTHOP field. This causes gobgp to produce the following error:

I1212 05:06:46.172157   22013 network_routes_controller.go:332] Processing bgp route advertisement from peer
WARN[0019] the received Update message was treated as withdraw  Key="fd00::9" State=BGP_FSM_ESTABLISHED Topic=Peer error="invalid nexthop address"

A recent PR in gobgp (https://github.com/osrg/gobgp/issues/1893) does not produce he above error/warning for IPv6 addresses in NEXTHOP field; however, according to rfc4760, NEXTHOP is only for IPV4 addresses. https://tools.ietf.org/html/rfc4760 

Suggested fix: remove the IPV6 address insertion into NEXTHOP.

Arvinderpal:
@uablrek I will issue a small PR for this fix.

gedOHub:
For IPv4 NextHopAttribute from API works. What methor I should use for setting nexthop for Ipv6?

aauren:
Closed via #606. Thanks again @Arvinderpal !

PR #606, opened 2018-12-12
Removes IPv6 address insertion from BGP IPv4-only NEXTHOP field

Arvinderpal:
Fixes Issue: https://github.com/cloudnativelabs/kube-router/issues/605


murali-reddy:
thanks @Arvinderpal LGTM.

PR #619, opened 2018-12-24
Periodicaly sync iptables MASQUERADE rules

bazuchan:
Periodicaly sync iptables MASQUERADE rules, so they will be added back if deleted manually.

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -350,6 +350,12 @@ func (nsc *NetworkServicesController) sync() error {
 	nsc.mu.Lock()
 	defer nsc.mu.Unlock()
 
+	// enable masquerad rule
+	err = ensureMasqueradeIptablesRule(nsc.masqueradeAll, nsc.podCidr)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


murali-reddy:
@bazuchan thanks for the PR.

calling `ensureMasqueradeIptablesRule` is sync loop make sense. 

Though eventually I think, kube-router needs periodic full-sync and delta-sync (that is done per service/endpoint update). Anything that we add in sync() will add more delay now to processing updates to the api objects.

[AppendUnique](https://github.com/coreos/go-iptables/blob/master/iptables/iptables.go#L144) already does the exist check, so feel its redundant, did you see any issue?

bazuchan:
> [AppendUnique](https://github.com/coreos/go-iptables/blob/master/iptables/iptables.go#L144) already does the exist check, so feel its redundant, did you see any issue?

Yes, I haven't checked that, it is indeed redundant. I will remove this part of pr. Also will changing log level to 2 for `glog.V(1).Info("Successfully added iptables masquerad rule")` be ok?



> Anything that we add in sync() will add more delay now to processing updates to the api objects.

It's true, but either there is a way to manage all host firewall with kube-router or we need a full sync.

PR #627, opened 2019-01-07
Fixed typo in `Global External BGP Peers` example

zerkms:
`peer-router-asns` does not need quotes around


murali-reddy:
thanks for catching this @zerkms

Issue #574, opened 2018-11-14
Kube-router, GoBGP and Kernel Routing Table, Service VIP Persists


pichanaquiMantaTer:
A service VIP is advertised from several k8s worker nodes to a route-reflector (GoBGP, external to cluster) with distinct next-hops all in subnets sometimes different from a selected worker node WN_0 not originating the VIP. In case one of the advertising worker nodes does belong to the same subnet as WN_0, as long as the BGP best next-hop is not in the subnet of WN_0, the VIP with next-hop does not get installed into the kernel routing table (KRT) of WN_0. However, if the BGP best route does have next-hop belonging to the same subnet as WN_0, the VIP with next-hop is installed into the KRT(WN_0). When the worker node no longer advertises the BGP best route for the VIP, the VIP with this next-hop should be withdrawn from the KRT. It is, on the other hand, withdrawn from the associated kube-router GoBGP table.

The VIP persists in the KRT(WN_0), which appears to be a bug in kube-router interaction with the KRT.

This also occurs when WN_0 originates the same VIP. In this case, since WN_0 has a pod advertising the VIP, a traceroute -I from the host of WN_0 to the VIP resolves to the VIP on the pod, despite the KRT only reporting the (/32 route) VIP to a next-hop on a different host on the same subnet.
Finally, does kube-router bypass the KRT altogether?

Reference: 
kube1 ~ # kube-router -V
Running kube-router version , built on 2018-09-12T20:15:54+0000, go1.10.3

All worker nodes only iBGP peer with an external route-reflector.

Thanks.

aauren:
Closed via #629

PR #629, opened 2019-01-09
cleanup local routes if nexthop moves outside host subnet

aauren:
This is a bug fix for: https://github.com/cloudnativelabs/kube-router/issues/574

We were seeing cases where the route would stick in the kernel routing table after the NextHop for a given service VIP had changed to a node outside of the host's subnet. This happens because there is a gap in the logic where if the NextHop for a given service is not withdrawn but rather transitions to a Kubernetes node outside the subnet of the current host nothing is done with the previously inserted route.

This PR makes it so that kube-router will cleanup lingering routes at the same point that it also cleans up tunnels when the next hop has transitioned to a node outside the subnet of the current host.

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -26,6 +26,7 @@ import (
 	"github.com/osrg/gobgp/table"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netlink/nl"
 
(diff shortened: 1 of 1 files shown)


aauren:
On further comment, without this fix the existing functionality has the potential to send service traffic to hosts that are no longer hosting the service and cause service disruptions. This is especially true if the service.local annotation is enabled for the given service:
```
kube-router.io/service.local: "true"
```

aauren:
@murali-reddy @roffe @andrewsykim - Whenever you can, do you think you could give this a review? This is a bug that we're actively experiencing in our environment.

murali-reddy:
LGTM

PR #634, opened 2019-01-11
Avoid duplicate peer pods in npc rules variables

jimmy-zh:
Fix the `buildNetworkPoliciesInfo` method in NPC to avoid duplicate peer pods in npc rules variables. 

This bug doesn't affect the final iptables rules, because`ipset` ensures that there is no duplicate pod IP in a set. However, we still need to fix it.


murali-reddy:
LGTM

PR #645, opened 2019-01-22
Specify docker.io as registry for multiple registries environment

Slach:
when use multiple registries for pulling images in container runtime 
we need specify which registry will use exacly

because https://quay.io/repository/cloudnativelabs/kube-router already exists and contains old kube-router image


murali-reddy:
thanks for the PR. LGTM

Issue #652, opened 2019-01-31
Services with 0 endpoints are created and deleted at each sync


renchap:
If a service has 0 target pods, each sync creates an IPVS service, and then deletes it (and its IP):

```
$ kubectl logs -f kube-router-vmg92 --namespace kube-system|grep 172.18.75.139
I0131 14:22:25.416456       1 network_services_controller.go:1855] Successfully added service: TCP:172.18.75.139:8080 (Flags: )
I0131 14:22:25.417174       1 network_services_controller.go:1855] Successfully added service: TCP:172.18.75.139:7070 (Flags: )
I0131 14:22:25.620621       1 network_services_controller.go:1053] Found an IP 172.18.75.139 which is no longer needed so cleaning up
I0131 14:22:25.649744       1 network_services_controller.go:1088] Found a IPVS service TCP:172.18.75.139:8080 (Flags: [hashed entry]) which is no longer needed so cleaning up
I0131 14:22:25.649826       1 network_services_controller.go:1088] Found a IPVS service TCP:172.18.75.139:7070 (Flags: [hashed entry]) which is no longer needed so cleaning up
I0131 14:22:26.478995       1 network_services_controller.go:1855] Successfully added service: TCP:172.18.75.139:8080 (Flags: )
I0131 14:22:26.484770       1 network_services_controller.go:1855] Successfully added service: TCP:172.18.75.139:7070 (Flags: )
I0131 14:22:26.649749       1 network_services_controller.go:1053] Found an IP 172.18.75.139 which is no longer needed so cleaning up
I0131 14:22:26.668463       1 network_services_controller.go:1088] Found a IPVS service TCP:172.18.75.139:8080 (Flags: [hashed entry]) which is no longer needed so cleaning up
I0131 14:22:26.668695       1 network_services_controller.go:1088] Found a IPVS service TCP:172.18.75.139:7070 (Flags: [hashed entry]) which is no longer needed so cleaning up
```

We have many services with no targets (replica set to 0 for example), so each time there is a pod targeted by a service starting or stopping, hundred of IPVS services are created and immediately deleted.

When you lose some nodes (or drain them), or some pods are stuck in a restart loop, this creates a *lot* of IPVS activity on our nodes, causing production issues (200+ IPVS updates per second).

Here is one endpoint:
```
apiVersion: v1
kind: Endpoints
metadata:
  creationTimestamp: "2018-12-05T00:57:43Z"
  labels:
    app: xxx
    metrics: jmx
[...]

PR #654, opened 2019-02-04
If no endpoint is available, don't create the ipvs rule

menardorama:
Had to close initial PR #653 sorry for that

This PR propose a temporary fix to issue #652 by just skipping ipvs rule creation when no endpoint is available.

I have removed the condition to filter for svc.local when filtering the ipvs rules creation.

Now ipvs is skipped when no endpoint is available.

There may be a better way to handle this but it stop create and delete ipvs rules

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -841,7 +841,7 @@ func (nsc *NetworkServicesController) syncIpvsServices(serviceInfoMap serviceInf
 
 		endpoints := endpointsInfoMap[k]
 
-		if svc.local && !hasActiveEndpoints(svc, endpoints) {
+		if !hasActiveEndpoints(svc, endpoints) {
 			glog.V(1).Infof("Skipping service %s/%s as it does not have active endpoints\n", svc.namespace, svc.name)
 			continue
 		}


murali-reddy:
@menardorama Sorry i don't see how this would work for both services that marked for `externaTrafficPolicy=Local` and the services that are not marked.

taking out `svc.local` means you are excluding any service which does not have local endpoint

menardorama:
I don't understand, I juste remove the condition for svc.local which is less restrictive.

Now externaTrafficPolicy=Local and the others are skipped when there is no endpoint.

It's just to prevent creation/deletion process in the same sync

Am I wrong ?

menardorama:
HI
Any news ?

murali-reddy:
@menardorama sorry for late response.

so there are two parts to the network_service_controller.

- build service and endpoints info by  listing the services and endpoints from the API
- get current servers, services from the IPVS and the add missing once and delete the not needed once

```
-		if svc.local && !hasActiveEndpoints(svc, endpoints) {
+		if !hasActiveEndpoints(svc, endpoints) {
```
your changes affect second part. So even if (svc.local == false) we will skip configuring IPVS services if the service does not have any endpoint on the local node. This is not desirable.

menardorama:
Ok thanks for the feedback I will have a look.

murali-reddy:
@menardorama do you plan to or have time to update the PR? If not i can merge the PR and address the feedback.

aauren:
Closing as parent issue is already addressed in the current master and releases.

PR #666, opened 2019-02-10
Introduces the option --overlay-type={subnet,full}, to be able to always generate IPIP tunnels regardless of node subnets

lucasmundim:
This PR implements PR #371 ideas, keeping the existing behaviour but allowing users to opt in for a full overlay regardless of the subnet the nodes are in

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -49,9 +49,10 @@ Usage of kube-router:
       --disable-source-dest-check        Disable the source-dest-check attribute for AWS EC2 instances. When this option is false, it must be set some other way. (default true)
       --enable-cni                       Enable CNI plugin. Disable if you want to use kube-router features alongside another CNI plugin. (default true)
       --enable-ibgp                      Enables peering with nodes with the same ASN, if disabled will only peer with external BGP peers (default true)
-      --enable-overlay                   When enable-overlay set to true, IP-in-IP tunneling is used for pod-to-pod networking across nodes in different subnets. When set to false no tunneling is used and routing infrastrcture is expected to route traffic for pod-to-pod networking across nodes in different subnets (default true)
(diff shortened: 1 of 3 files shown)


docs/user-guide.md
```diff
@@ -49,9 +49,10 @@ Usage of kube-router:
...
+      --enable-overlay                   When enable-overlay is set to true, IP-in-IP tunneling is used for pod-to-pod networking across nodes in different subnets. When set to false no tunneling is used and routing infrastructure is expected to route traffic for pod-to-pod networking across nodes in different subnets (default true)
       --enable-pod-egress                SNAT traffic from Pods to destinations outside the cluster. (default true)
       --enable-pprof                     Enables pprof for debugging performance and memory leak issues.
+      --full-overlay                     When full-overlay is set to true, it changes "--enable-overlay=true" default behavior so that IP-in-IP tunneling is used for pod-to-pod networking across nodes regardless of the subnet the nodes are in. When set to false, the default, default "--enable-overlay=true" behavior is used
```
andrewsykim: What do you think about using `--overlay-type` instead, with a default value `regional` or `subnet`. And the other option is `type`?

lucasmundim: Sounds like a good idea to me. I don't know if I understood your question correctly ("And the other option is type?"). Could you elaborate a bit more?

andrewsykim: Sorry, typo. I meant "And the other option is full". 

So user needs to set both
```
--enable-overlay=true
--overlay-type={subnet,full}
```

andrewsykim: cc @murali-reddy

lucasmundim: > Sorry, typo. I meant "And the other option is full".

Now I got it :)

> So user needs to set both
> 
> ```
> --enable-overlay=true
> --overlay-type={subnet,full}
> ```

Yes, it makes sense. I've updated the code to reflect that.

murali-reddy:
Did basic testing switching flag `--overlay-type={subnet,full}`. Code changes look good..

PR #669, opened 2019-02-15
fix #639

### LINKED ISSUE #639: Removing ExternalIP Fails to Withdraw Route
I am using externalIPs to advertise out anycast BGP routes from kube-router. Advertising works as you'd expect: I add an externalIP and it is advertised. Unfortunately when removing the externalIP, the route is never withdrawn from my peers. 

Running kube-router version v0.2.0-beta.7, built on 2018-06-08T09:34:34+0000, go1.8.7

mk01:
currently are withdrawals calculated from actual setup of externalIPs only - real previous state is not considered.

this commit keeps list of exported externalIPs.
on service change it it provides the actual difference between advertised extIPs from previous and new state.


mk01:
closing in favour of #670

PR #670, opened 2019-02-15
fix #639

### LINKED ISSUE #639: Removing ExternalIP Fails to Withdraw Route

mk01:
- get external IPs to withdraw as diff against previous generation of service

### DIFF
--- a/pkg/controllers/routing/ecmp_vip.go
+++ b/pkg/controllers/routing/ecmp_vip.go
@@ -62,7 +62,7 @@ func (nrc *NetworkRoutingController) withdrawVIPs(vips []string) {
 func (nrc *NetworkRoutingController) newServiceEventHandler() cache.ResourceEventHandler {
 	return cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			nrc.OnServiceUpdate(obj)
+			nrc.OnServiceCreate(obj)
 		},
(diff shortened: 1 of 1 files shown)


pkg/controllers/routing/ecmp_vip.go
```diff
@@ -65,7 +66,7 @@ func (nrc *NetworkRoutingController) newServiceEventHandler() cache.ResourceEven
 			nrc.OnServiceUpdate(obj)
 		},
 		UpdateFunc: func(oldObj, newObj interface{}) {
-			nrc.OnServiceUpdate(newObj)
+			nrc.OnServiceUpdate(newObj, oldObj)
```
murali-reddy: @mk01 Please add new function `OnServiceCreate` and use it to handle `AddFunc`. And modify the signature of `OnServiceUpdate` to take the new and old objects explicitly (instead of objOld ...interface{})

There should have been separate handler for just Add to begin with. Nevertheless its good to modify it now.

murali-reddy:
LGTM.

PR #680, opened 2019-02-25
NFTables handler for the network policy controller

johanot:
We're currently in the middle of a stabilization phase with this feature, which is why I mark this as a draft for now. But.. I think it's time to create a PR to get the discussion going here. I'd like to hear anyone's comments and thoughts on this.

## Background
For quite some time, we've experienced performance problems with the current network policy implementation. I'll quote *xanonus* from kube-router Slack here:
> Hi, we have the problem, that iptables settings are applied too delayed after pods are started. 
> Sometimes it seems to take up 30-40s until connections are possible. The
> kube_router_controller_iptables_sync_time average over the cluster is about 10s,
> but on some nodes is has spikes up to 40s.
> The cluster is not very large (single master, ~50 nodes & ~50 NPCs
> whereas only very few NPCs per Pod). 
> (...)

This is exactly what we're seeing, and we think the issues are due to the great amount of individual iptables commands being executed for-each pod+netpol on each node. One could try and fix this problem by implementing partial rule sync instead of doing full resync for-each iteration of the controller, but that is difficult and it potentially opens for all sorts of *out-of-sync*-issues.

We thought we could just solve the problem by creating a file with rules for `iptables-restore` to consume. That would reduce the number of `os.exec`'s and copy'ing of data to and from user-space to just one pr. sync. But alas.. iptables has a namespacing issue. Particularly due to it using global chains for FORWARD, INPUT, OUTPUT, MANGLE etc. This means that if a kube-router host runs anything else that manipulates iptables rules, `iptables-restore` might accidentally flush those rules while reconciling the network policy rules. 

nftables solves the namespacing issue by supporting custom tables and chain hooks.

For the record: The're plenty of other reasons for choosing nftables over iptables :).

## Implementation

Basically this change involves:

1) Move current iptables command execution to a separate file.
2) Make the current network policy struct representation slightly more generic.
3) Add new nftables handler alongside the existing iptables handler.
4) Add a flag to choose between either iptables or nftables as network policy handler/backend and let iptables be the default.

---

### 1. Move current iptables command execution to a separate file.

This is what's going to give me the most amount of merge-conflicts until this PR gets either merged or abandoned. I felt however, that it was necessary to not just logically, but also physically separate the actual implementation from the controller. This makes it far simpler to implement other alternative network policy handlers, while still sharing components like the network policy elevator.

iptables functions have been moved to `pkg/controllers/netpol/iptables.go`.

### 2. Make the current network policy struct representation slightly more generic.

Structs+fields have (most of them) become public. This makes it easier to share data between abstraction layers. It also makes it easier to inject data from unit tests. Without forcing everything into the same package.

In some cases, structs have been slightly changed. This is very the current data abstraction had a tight coupling to the iptables / ipset implementation. Ipblocks for example, was `[][]string` where the first dimension holds the CIDR and the second dimension holds implementation specific ipset-options.

### 3. Add new nftables handler alongside the existing iptables handler.

nftables implementation has been added at `pkg/controllers/netpol/nftables.go`.

Basically this is nothing more than a go-template that creates a file for `nft -f <file>` to read and execute. The `nft -f <file>` operation is executed in an atomic fashion, and; if parsing of the rules file fail, the netfilter ruleset is unchanged. See: https://wiki.nftables.org/wiki-nftables/index.php/Atomic_rule_replacement

### 4. Add a flag to choose between either iptables or nftables as network policy handler/backend ...

This is what makes this PR backward-compatible. Unless you opt-in for the nftables firewall implementation, you won't get it. At least for now.


murali-reddy:
@johanot thanks for your work on this. I will play around with this. Is there a docker image that I can use?

johanot:
@murali-reddy Great! I've just pushed an extra commit to the branch now adding nftables to the docker image PATH. That should be sufficient to test, as long as you start kube-router with `--network-policy-handler=nftables`.

If you don't want to `make container` yourself, I've pushed the resulting image here: `johanot/kube-router-git:nftables-second-attempt`

johanot:
@murali-reddy Did you get a chance to try this out? :-) I've been busy with other things myself, but we will take this piece to production in two weeks time.

murali-reddy:
@johanot apologies for reverting back so late

I seem to be running into some issue, and network policies does not seem to work.

```
I0419 08:35:08.096670       1 network_policy_controller.go:310] Found pod name: frontend-69859f6796-hlzzr namespace: default for which network policies need to be applied.
I0419 08:35:08.096733       1 network_policy_controller.go:310] Found pod name: frontend-69859f6796-p5grr namespace: default for which network policies need to be applied.
I0419 08:35:08.096786       1 nftables.go:166] Flushing nftables configuration to file: /tmp/kube-router.nft
I0419 08:35:08.098503       1 nftables.go:198] Execute nftables command: -f /tmp/kube-router.nft
I0419 08:35:08.100620       1 network_policy_controller.go:218] sync network policies took 4.242719ms
E0419 08:35:08.100727       1 network_policy_controller.go:143] Error during periodic sync of network policies in network policy controller. Error: exit status 1
E0419 08:35:08.100775       1 network_policy_controller.go:144] Skipping sending heartbeat from network policy controller as periodic sync failed.
I0419 08:35:08.102882       1 network_policy_controller.go:161] Received update to pod: kube-system/kube-controller-manager-weave-master
I0419 08:35:08.102965       1 network_policy_controller.go:221] Starting sync of network policies
I0419 08:35:08.103059       1 network_policy_controller.go:310] Found pod name: frontend-69859f6796-p5grr namespace: default for which network policies need to be applied.
I0419 08:35:08.103109       1 network_policy_controller.go:310] Found pod name: frontend-69859f6796-hlzzr namespace: default for which network policies need to be applied.
I0419 08:35:08.103153       1 nftables.go:166] Flushing nftables configuration to file: /tmp/kube-router.nft
I0419 08:35:08.103540       1 nftables.go:198] Execute nftables command: -f /tmp/kube-router.nft
I0419 08:35:08.107352       1 network_policy_controller.go:218] sync network policies took 4.383439ms
E0419 08:35:08.107432       1 network_policy_controller.go:170] Error syncing network policy for the update to pod: kube-system/kube-controller-manager-weave-master Error: exit status 1

```

aauren:
This is a good bit of work and this comment isn't meant to diminish what has been done here at all, but just as maybe another piece of the conversation.

It seems like nftables is already beginning to be outmoded by bpfilter (https://lwn.net/Articles/747551/), if we were going to change the firewall implementation for kube-router, should we consider bpfilter instead?

johanot:
@aauren Before writing the first line code for this PR, I read that article. :-) I also read: https://cilium.io/blog/2018/04/17/why-is-the-kernel-community-replacing-iptables/ and looked into the Cilium Kubernetes firewall implementation. Admitted, it looks good and it's probably the academically correct way to go, but in the end, the perfect solution was not what I was looking for.

I was looking for a concrete and simple way of solving some very present problems with the current firewall implementation. _Simple_ is the keyword here, because I actually investigated and tested bpf frameworks for golang, but it seems that the only way you can get full-featured bpf interaction from Go is to pull in the C-bindings, like Cilium does. Also, it seems like nftables have much better support for things like tracing and debugging, which I've actually added on my company's internal branch for this impl., though it is not part of this PR yet.

One solution could of course be to vendor the Cilium implementation directly, but I don't know about licensing there, and also; sportsmanship.

Overall nftables just seem much more mature, and we also have to think about boring things like kernel support. Even though some never features of nftables have been added in 4.X, most basic features are available since 3.18.

Sidenote: nftables is far from dead and it's in active development. There have even been proposals for nftables to use BPF as backend. If that becomes reality, I think the argument for plain BPF bytecode is no longer there, since nftables syntax is just much more readable and hence debugable.

johanot:
@murali-reddy Thanks for testing.

Can you perhaps send me the policies you have on your test cluster? Also, I will push some small fixes to this branch soon.

aauren:
@johanot Thanks for your very thorough and in-depth reply.

I can see that you have considered the angles. I was also not aware of the implementation limitations that would be present for a GO application in it's usage of bpfilters. :smile:

ticpu:
After more testing we discovered that the iptables handler will always allow traffic to the cluster IPs in our test environment. I'd say that the nftables handler is currently doing the right thing by disallowing access to services, however, there's no way to tell the network policy controller that we want to access a specific service.

I see 2 ways to fix this issue, one would be to use the podSelector to trace back to the endpoint of the pod and then to the service and get the clusterIP to allow it. The other way would be to have a serviceSelector (svcSelector?) in networkpolicies that would allow to match against both the clusterIP and externalIPs of the service.

Right now, I was able to make this work by allowing the whole clusterIP range for the port to replicate the iptables handler behavior. But right now, I really like how clean the new code is and how fast it reacts, nftables is also much easier to read. We will continue to test this branch and will move to production soon if it continue working as it does.

Great job on this!

johanot:
@ticpu Thanks for trying this out! It's been a long time since I posted this draft PR, but since then we've moved to production with this in my org. The only issues we've had so far, is that people had to get used to being explicit about egress policies as you mention, and yes - we do the same as you - i.e. allow the entire serviceCIDR in a netpol. I agree that a svcSelector on netpols would be nice, but I guess that requires changes to the k8s api.

ticpu:
@johanot, would you mind me opening a new pull request for the rebased branch with the service selector including your updated commits?  We've moved up to our test environment from dev without any issue. Still amazed on how fast it is.

aauren:
Closing as this PR is superseded by #745

PR #682, opened 2019-02-27
Add iptables rules for accessing tunneled services from nodes

bazuchan:


murali-reddy:
@bazuchan Sorry I am not sure what the problem is. Could you please open an issue and link to this PR or add some description?

Why do we need that rule in `OUTPUT` chain, the one in `PREROUTING` should apply to both the cases?

bazuchan:
When a local process on a node (or host-networked pod) originates connection to tunneled service, it's packets doesn't go through PREROUTING chain. You can see it on this chart [https://en.wikipedia.org/wiki/Netfilter#/media/File:Netfilter-packet-flow.svg](url). So it doesn't get fwmark-ed and doesn't get routed to IPVS.

murali-reddy:
got it. LGTM. thanks.

Issue #684, opened 2019-03-04
Use conflist format by default for cni config


MarkDeckert:
## Issue:
I'm using kube-router with all features installed, completely replacing kube-proxy. When using any additional CNI plugins alongside kube-router, you would have to use the list format of the cni config.  Right now, kube-router by default uses the non-list format.

## How to duplicate (example):
Deploy linkerd-cni.  This actually detects and renames the .conf to .conflist and reformats the file to list format prior to adding itself in as a plugin.  Since kube-router is hardcoded to non-list format it will break on next restart.

## Suggested fix:
In order to add more CNI plugins right now it takes a number of very manual steps that require modifying the config in the CM and 2 places in the daemonset yaml.  Instead, kube-router could always use conflist by default, which means deploying CNI plugins (like linkerd-cni) would have no additional steps required by kube-router.  Only the following changes would be required in the yaml manifests to make this the default.

1.  cni-conf.json entry in configmap becomes:
```
  cni-conf.json: |
    {
       "cniVersion":"0.3.0",
       "name":"mynet",
       "plugins":[
          {
             "name":"kubernetes",
             "type":"bridge",
             "bridge":"kube-bridge",
             "isDefaultGateway":true,
             "ipam":{
                "type":"host-local"
             }
          }
       ]
    }
```
2.  Environment variable in kube-router container is set by default:
```
        - name: KUBE_ROUTER_CNI_CONF_FILE
          value: /etc/cni/net.d/10-kuberouter.conflist
```
3. Init container command is set by default:
```
      - command:
        - /bin/sh
        - -c
        - set -e -x;if [ ! -f /etc/cni/net.d/10-kuberouter.conflist ]; then TMP=/etc/cni/net.d/.tmp-kuberouter-cfg;
          cp /etc/kube-router/cni-conf.json ${TMP}; mv ${TMP} /etc/cni/net.d/10-kuberouter.conflist;
          fi
```

## Note:
I submitted an issue on the linkerd project to request an update to install documentation to warn about this issue with auto-renaming the config, but it would be nice if kube-router worked seamlessly either way.

MarkDeckert:
Please see this PR https://github.com/cloudnativelabs/kube-router/pull/688

PR #688, opened 2019-03-12
use conflist format for all deploys

MarkDeckert:
use conflist format for all deploys

### DIFF
--- a/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
+++ b/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
@@ -9,13 +9,19 @@ metadata:
 data:
   cni-conf.json: |
     {
-      "name":"kubernetes",
-      "type":"bridge",
-      "bridge":"kube-bridge",
-      "isDefaultGateway":true,
-      "ipam": {
-        "type":"host-local"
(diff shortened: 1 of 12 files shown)


MarkDeckert:
PR for https://github.com/cloudnativelabs/kube-router/issues/684

I realized while making the changes this is essentially identical to what is already done in the hostport example manifest.  I just ticked up the cni version number to 0.6.0.

murali-reddy:
@MarkDeckert thanks for the PR. Please rebase.

Does it also make sense to remove previous file (if it exists) with `.conf` extension. I believe if there exists more than one config file, first one in lexicographic order will be chosen by kubelet when kube-router might use one with `.conflist` extension.

MarkDeckert:
I've messed this one up massively trying to correct a few things.  I will do a new PR

MarkDeckert:
@murali-reddy thank you for mentioning the kubelet behavior.  Current kubelet documentation proves that to be correct.  It does make sense to remove previous .conf files.  I will be doing a new PR though for other reasons.

I also considered deleting/handling any previous .conflist files as well, but that gets into a bunch of other issues so in that case might be better to just let it break.

PR #690, opened 2019-03-14
default cni config to list format

MarkDeckert:
Updated PR - this one also deletes existing *.conf files in cni config directory.  These are some thoughts:

1. This deletes existing *.conf file so it can safely be applied over existing kube-router and it will update the conf format.

2. Although the kubelet only reads the first config in lexicographic order, there doesn't appear to be any limitation in the cni spec itself against having multiple configs.  Therefore, the init container only deletes *.conf files if there was no pre-existing kube-router-installed 10-kuberouter.conflist file.  The assumption is the first install is in a prepared environment, but after that all bets are off.

2. Existing *.conflist or other files are not deleted.  If kube-router is deployed into an environment like this the assumption is that the environment is complex and carefully crafted.  But, if it gets deployed without that extra care, it may break things but at least it doesn't irretrievably destroy the previous configs.

3. Perhaps some intelligence could be added where the init/install process parses existing conf/conflist files and reworks/adds itself in seamlessly, but this seems out of scope for a simple deployment config change like this.

### DIFF
--- a/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
+++ b/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
@@ -9,7 +9,7 @@ metadata:
 data:
   cni-conf.json: |
     {
-       "cniVersion":"0.6.0",
+       "cniVersion":"0.3.0",
        "name":"mynet",
        "plugins":[
           {
--- a/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
+++ b/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
@@ -9,13 +9,19 @@ metadata:
 data:
   cni-conf.json: |
     {
-      "name":"kubernetes",
-      "type":"bridge",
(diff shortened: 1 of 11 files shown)


MarkDeckert:
https://github.com/cloudnativelabs/kube-router/issues/684

murali-reddy:
@MarkDeckert thanks for reworking on the PR. Sorry i should have caught this earlier. But ` "cniVersion":"0.6.0"` is in valid, (you may be getting confused from CNI version with plugin versions)

See https://github.com/containernetworking/cni/blob/master/SPEC.md

Please leave CNI version to be 0.3.0 as its in https://github.com/cloudnativelabs/kube-router/blob/master/daemonset/kubeadm-kuberouter-all-features-hostport.yaml#L12

MarkDeckert:
@murali-reddy Thank you I shouldn't have added that change at the last minute.  I have fixed the version at 0.3.0 now.

Issue #692, opened 2019-03-18
enableForwarding breaks forwarded traffic security for dual-networks nodes


RaveNoX:
Hello.
I have following cluster setup:
![kube-router-break](https://user-images.githubusercontent.com/1229948/54511802-2140de00-4963-11e9-989f-db6a87681b17.png)

In this issue I will describe problem on "k8s Node-Ext" node.
This node is connected to both internet and LAN networks, as you can see.
For internet connection used interface "ens1" and for LAN connection interface "ens3" is used.

Kube-router was installed on all nodes.
There are no workload are currently running on the cluster.

Now if I looks at rules on "k8s Node-Ext" then I will see  following:
```
#iptables-save -t filter
# Generated by iptables-save v1.6.1 on Mon Mar 18 09:48:11 2019
*filter
:INPUT ACCEPT [644:111554]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [667:124014]
:DOCKER-USER - [0:0]
:KUBE-FIREWALL - [0:0]
:KUBE-ROUTER-SERVICES - [0:0]
-A INPUT -m comment --comment "handle traffic to IPVS service IPs in custom chain" -m set --match-set kube-router-service-ips dst -j KUBE-ROUTER-SERVICES
-A INPUT -j KUBE-FIREWALL
-A FORWARD -o ens3 -m comment --comment "allow outbound node port traffic on node interface with which node ip is associated" -j ACCEPT
-A FORWARD -o kube-bridge -m comment --comment "allow inbound traffic to pods" -j ACCEPT
-A FORWARD -i kube-bridge -m comment --comment "allow outbound traffic from pods" -j ACCEPT
-A FORWARD -j DOCKER-USER
-A OUTPUT -j KUBE-FIREWALL
-A DOCKER-USER -j RETURN
-A KUBE-FIREWALL -m comment --comment "kubernetes firewall for dropping marked packets" -m mark --mark 0x8000/0x8000 -j DROP
-A KUBE-ROUTER-SERVICES -m comment --comment "allow input traffic to ipvs services" -m set --match-set kube-router-ipvs-services dst,dst -j ACCEPT
-A KUBE-ROUTER-SERVICES -p icmp -m comment --comment "allow icmp echo requests to service IPs" -m icmp --icmp-type 8 -j ACCEPT
[...]

Issue #120, opened 2017-08-11
add filter table FORWARD chain rule to forward traffic from "kube-bridge"


murali-reddy:
~~By default Linux has `filter` table `FORWARD` chain has default policy as `DROP`.~~

Its actually policy by Docker 1.13 https://github.com/moby/moby/pull/28257

Some possible options for CNI plugins https://github.com/kubernetes/kubernetes/issues/40182

I guess Kube-router should add exception atleast to permit the traffic from `kube-bridge` interface

```
iptables -A FORWARD -o kube-bridge -j ACCEPT
iptables -A FORWARD -i kube-bridge -j ACCEPT
```

murali-reddy:
Testing the patch #127 figured that we need one more rule in case service proxy to permit node-port traffic.

`iptables -t filter -A FORWARD -o ens33 -j ACCEPT`

PR #695, opened 2019-03-21
Change outbound nodeport rule to accept RELATED and ESTABLISHED traffic

RaveNoX:
This pull request fixes issue #692

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -639,8 +639,8 @@ func (nrc *NetworkRoutingController) enableForwarding() error {
 		}
 	}
 
-	comment = "allow outbound node port traffic on node interface with which node ip is associated"
-	args = []string{"-m", "comment", "--comment", comment, "-o", nrc.nodeInterface, "-j", "ACCEPT"}
+	comment = "allow outbound node port traffic"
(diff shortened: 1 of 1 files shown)


murali-reddy:
@RaveNoX thanks for the PR and detailed bug report #692 

I don't recall why (but very likely due docker changing default FORWRD policy as described in https://github.com/kubernetes/kubernetes/issues/39823) that rule was added as part of #120. Current rule is a bad solution. So it must be changed.

I need to test it to understand better. But here are the questions that are going through my mind

- Is ACCEPT for RELATED and ESTABLISHED  is enough? Does IPVS<-->real-server traffic is considered RELATED to the client<-->IPVS
- Does the packet hit FORWARD chain, from http://www.austintek.com/LVS/LVS-HOWTO/HOWTO/LVS-HOWTO.filter_rules.html it indicates packet may go through the POSTROUTING chain

Let me try the fix and revert back

RaveNoX:
Here a good guide about IPVS & IPTABLES interaction:
http://www.austintek.com/LVS/LVS-HOWTO/HOWTO/LVS-HOWTO.filter_rules.html

RaveNoX:
Hello!
Is there any news on this issue?

trnl:
@RaveNoX, thanks for the fix! 

@murali-reddy  I wonder if there are any updates from your side? It looks like we are hitting the same problem at the end.

RaveNoX:
@murali-reddy any news on this?

Issue #572, opened 2018-11-08
Add TCP graceful termination


MarkDeckert:
Feature request:

When a pod is deleted, make an attempt to wait for TCP sessions to drain before killing the pod. Perhaps  by watching active connection count in ipvs and adding an additional wait period.  I would assume at some point you would want to just kill any long-running or stuck sessions regardless.

As a reference, see related functionality in kube-proxy: https://github.com/kubernetes/kubernetes/pull/66012

Also it might be a nice feature for there to be an option to disable or change the behavior slightly on a service by service basis via annotation.  If for example an app is configured in a way where it's intentional that clients never drop a session, the user might want to just always immediately force drop, or conversely, extend the wait period, on pod delete.

roffe:
The default period is 5 minutes and can be set using `--ipvs-graceful-period` cli-opt 

The timeout can be overriden per service by setting the annotation `kube-router.io/service.gracefulterminationperiod` to a duration string

example:
    kube-router.io/service.gracefulterminationperiod: 30s

If the duration string is invalid the default period set by `--ipvs-graceful-period` will be used

I'm trying to get Active & Inactiveconn into the equation but i need to upgrade the vendored netlink & ipvs package for that to happend

aauren:
Closed via #706. Thanks again @roffe!

PR #706, opened 2019-04-09
Issue 572 - Graceful termination + Update to go-1.10.8, alpine-3.9

roffe:
Add graceful termination for IPVS destinations
Cleanup main loop of Network Service Controller
Stop removing services with 0 endpoints
Upgrade vendored libnetwork & netlink library ( to get statistics for IPVS destinations )
Update to go-1.10.8 & alpine 3.9


mrueg:
Not sure if it will help to run the `go test` with `-v` for more verbosity?

roffe:
@mrueg i've spent the whole day trying to get the tests working on 1.11 but it seems the routing controller ones are the ones failing, specificly it seems it's related to gobgp somehow deadlocking in the tests.

roffe:
Closes https://github.com/cloudnativelabs/kube-router/issues/572

.travis.yml
```diff
@@ -3,7 +3,7 @@ services:
 
 language: go
 go:
-  - 1.10.x
+  - 1.10.8
```
mrueg: 1.10.x is technically 1.10.8, not sure if this is needed

mrueg: would suggest to add a comment that tests are stuck with 1.11.x and later here instead

roffe: The tests for the router controller does not work with Go 1.11, i suspect it being due to the old vendored goBGP server

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -1059,9 +1100,9 @@ func (nsc *NetworkServicesController) syncIpvsServices(serviceInfoMap serviceInf
 	// cleanup stale IPs on dummy interface
 	glog.V(1).Info("Cleaning up if any, old service IPs on dummy interface")
 	addrActive := make(map[string]bool)
-	for k, endpoints := range activeServiceEndpointMap {
+	for k := range activeServiceEndpointMap {
```
roffe: @murali-reddy The problem I had is that services that does not have any endpoints are flapping, making many unnecessary LN calls to add and remove the service on every sync. also if the only endpoint in the service is gracefully terminating we can't yank the service out under it.

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -330,34 +336,74 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
...
+				glog.Errorf("Skipping sending heartbeat from network service controller as periodic sync failed.")
+			} else {
+				healthcheck.SendHeartBeat(healthChan, "NSC")
+			}
 		}
 	}
 }
 
-func (nsc *NetworkServicesController) sync() error {
+func (nsc *NetworkServicesController) sync(syncType int) {
```
roffe: The new behaviour is instead of having a blocking sync() on every update we have a overflowing channel, if there is a request to sync while we are syncing we will still do it but we won't have X-XX amount of them hanging in the air my the mutex in it

murali-reddy:
thanks @roffe LGTM

Issue #712, opened 2019-04-14
kube-router pods crash when api server restarts/goes down


MarkDeckert:
# Kube-router version
0.2.5

# kubernetes version
1.13.5 bare metal

# Features
All features enabled

# Problem
If API server is down/unavailable, kube-router pods often crash. This causes any incoming cluster traffic that depends on the proxy to fail.

# Error message:

> client-go/informers/factory.go:73: Unexpected watch close - watch lasted less than a second and no items received                                                                                                                       [5/1248]
W0402 16:53:17.623971       1 reflector.go:334] github.com/cloudnativelabs/kube-router/vendor/k8s.io/client-go/informers/factory.go:73: watch of *v1.NetworkPolicy ended with: very short watch: github.com/cloudnativelabs/kube-router/vendor/k
8s.io/client-go/informers/factory.go:73: Unexpected watch close - watch lasted less than a second and no items received
W0402 16:53:17.624008       1 reflector.go:334] github.com/cloudnativelabs/kube-router/vendor/k8s.io/client-go/informers/factory.go:73: watch of *v1.Pod ended with: very short watch: github.com/cloudnativelabs/kube-router/vendor/k8s.io/clie
nt-go/informers/factory.go:73: Unexpected watch close - watch lasted less than a second and no items received
W0402 16:53:17.624033       1 reflector.go:334] github.com/cloudnativelabs/kube-router/vendor/k8s.io/client-go/informers/factory.go:73: watch of *v1.Endpoints ended with: very short watch: github.com/cloudnativelabs/kube-router/vendor/k8s.i
o/client-go/informers/factory.go:73: Unexpected watch close - watch lasted less than a second and no items received
W0402 16:53:17.624060       1 reflector.go:334] github.com/cloudnativelabs/kube-router/vendor/k8s.io/client-go/informers/factory.go:73: watch of *v1.Namespace ended with: very short watch: github.com/cloudnativelabs/kube-router/vendor/k8s.i
o/client-go/informers/factory.go:73: Unexpected watch close - watch lasted less than a second and no items received
W0402 16:53:17.624083       1 reflector.go:334] github.com/cloudnativelabs/kube-router/vendor/k8s.io/client-go/informers/factory.go:73: watch of *v1.Node ended with: very short watch: github.com/cloudnativelabs/kube-router/vendor/k8s.io/cli
ent-go/informers/factory.go:73: Unexpected watch close - watch lasted less than a second and no items received
[...]

PR #714, opened 2019-04-15
issue-712 handle panic when non-pod is passed to OnPodUpdate

MarkDeckert:
This will handle a panic in the NetworkPolicy controller when trying to assert pod on what sometimes turns out to be a non-pod object

Used as a reference from similar situation:
https://github.com/kubernetes/kubernetes/blob/56817f254391d8ac06969ae4265d36b8bd64af0e/pkg/cloudprovider/providers/azure/azure.go#L544-L552


MarkDeckert:
https://github.com/cloudnativelabs/kube-router/issues/712

aauren:
Closing as the parent issue has been resolved and tested here: https://github.com/cloudnativelabs/kube-router/issues/712#issuecomment-614570483

PR #719, opened 2019-04-25
add --nodeport-bindon-all-ip to generic kuberouter manifests.

bserdar:
I have a cluster on linode where worker nodes have a public IP and a private IP. Without the --nodeport-bindon-all-ip, kuberouter binds the nodeport to the public IP only, but linode load balancers require a private IP, so the ingress doesn't work without this flag.

### DIFF
--- a/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
+++ b/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
@@ -70,6 +70,7 @@ spec:
         - "--run-router=true"
         - "--run-firewall=true"
         - "--run-service-proxy=true"
+        - "--nodeport-bindon-all-ip"
         - "--kubeconfig=/var/lib/kube-router/kubeconfig"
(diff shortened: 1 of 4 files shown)


murali-reddy:
@bserdar thanks for your PR.

If `--nodeport-bindon-all-ip=false`, its hard-coded to listen *only* on the node IP. Is it not the behaviour your seeing?

> Without the --nodeport-bindon-all-ip, kuberouter binds the nodeport to the public IP only, but linode load balancers require a private IP, so the ingress doesn't work without this flag.

Which one (public or private?) is your node IP in your case?

bserdar:
What do you mean by "node ip"?

The node has two addresses. How does it pick which address to listen on? Without the flag, it listens on the public address only.

aauren:
Closing as I believe the default of not binding on all interfaces is more desirable for most users and any users that do desire this setting are able to easily update their chart before deployment.

PR #720, opened 2019-04-25
Prefer node PodCIDR from an annotation

dparalen:
Current implementation never considers the "kube-router.io/pod-cidr"
annotation when creating an ipset for the node pod network CIDR.
The Node.Spec.PodCIDR is always used instead.

This patch prefers the annotation PodCIDR over the Node.Spec.PodCIDR

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -549,7 +549,15 @@ func (nrc *NetworkRoutingController) syncNodeIPSets() error {
 	currentPodCidrs := make([]string, 0)
 	currentNodeIPs := make([]string, 0)
 	for _, node := range nodes.Items {
-		currentPodCidrs = append(currentPodCidrs, node.Spec.PodCIDR)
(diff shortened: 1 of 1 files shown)


dparalen:
Hi all,
just to add a bit more context, I've run into an [K8s node pod CIDR API issue](https://github.com/kubernetes/kubernetes/issues/76761) the other day, that manifests in kube-router as the following error log:
```
E0425 13:34:14.081520    3889 network_routes_controller.go:262] Error synchronizing ipsets: Failed to sync Pod Subnets ipset: ipset v6.34: Syntax error: cannot parse : resolving to IPv4 address failed
```
After a bit of debugging, it turns out the issue is the empty string from the undefined value of `Node.Spec.PodCIDR`. This happens inspite of the node being annotaded with `kube-router.io/pod-cidr`.

murali-reddy:
thanks @dparalen LGTM

dparalen:
@murali-reddy thanks for a very quick response!

PR #722, opened 2019-04-25
Correct ipip Always On

aauren:
@murali-reddy @lucasmundim 

In the wake of #666 we removed some returns from the route/tunnel cleanup process of the `injectRoute` function in order to facilitate the `--overlay-type=full` option. However, these returns were what prevented routes and tunnels from being created when `--enable-overlay=false` was specified.

As such, kube-router will now make IPIP tunnels and route traffic to them regardless of whether `--enable-overlay=true` or not. This PR facilitates both functionalities by choosing to return ONLY if overlays have been disabled.

In other words:
* `--enable-overlay=false` - means that tunnels and routes to those tunnels will be cleaned and NOT created no matter how `--overlay-type` is set
* `--enable-overlay=true` - IPIP tunnels and routes to those tunnels will be created based upon the value of the `--overlay-type` setting

I've tested this in my own cluster and can confirm that it works correctly.

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -440,7 +440,8 @@ func (nrc *NetworkRoutingController) injectRoute(path *table.Path) error {
 
 	// create IPIP tunnels only when node is not in same subnet or overlay-type is set to 'full'
 	// prevent creation when --override-nexthop=true as well
-	if (!sameSubnet || nrc.overlayType == "full") && !nrc.overrideNextHop {
+	// if the user has disabled overlays, don't create tunnels
+	if (!sameSubnet || nrc.overlayType == "full") && !nrc.overrideNextHop && nrc.enableOverlays {
 		// create ip-in-ip tunnel and inject route as overlay is enabled
 		var link netlink.Link
 		var err error
@@ -488,7 +489,7 @@ func (nrc *NetworkRoutingController) injectRoute(path *table.Path) error {
 			Dst:       dst,
 			Protocol:  0x11,
 		}
-	} else {
+	} else if sameSubnet {
 		route = &netlink.Route{
 			Dst:      dst,
 			Gw:       nexthop,


aauren:
@murali-reddy I saw that you released 0.3.0 earlier today. Might I suggest that this be merged and tagged in an upcoming 0.3.1 release? 0.3.0 broke the `--enable-overlay` option by disregarding it through a logic omission.

As such in the current 0.3.0 release ALL pod network traffic is sent via tunnel no matter what options are passed to kube-router. This breaks our use case and likely several other people's use cases who want to do pod routing directly via BGP without the extra overhead incurred by IPIP.

aauren:
Changed requested merge branch to `0.3` as requested in slack.

murali-reddy:
@aauren I looked at the code, yes there is problem(a regression from 0.2.5), but does not seem this fix will completely address the issue.

So there are four models

- direct routing across the nodes with in same L2 subnet (this should work irrespective of `--enable-overlay` is true/false and `--overlay-type` is subnet/full)
- if `--enable-overlay` is false **only** direct routing across the nodes in same L2 subnet works
- if `--enable-overlay` is true *and*   `--overlay-type` is subnet direct routing across the nodes in same L2 subnet works and IPIP tunnels are used across the nodes in different L2 subnets
- if `--enable-overlay` is true *and*   `--overlay-type` is full then full IPIP overlay used

With your fix, #2 will be broken as we skip adding routes so direct routing does not work.

aauren:
@murali-reddy I'm a little bit confused about your point with `#2`. Could you explain what you meant by `direct routing`?

I guess my assumption here is that anyone that would choose `--enable-overlay=false` would be relying on BGP to route their pod and service traffic. If they don't have the benefit of BGP then they would need to use tunnelling in order to directly route anything that is not L2 right?

The previous functionality before #666 was: if `--enable-overlay` was set to false (https://github.com/cloudnativelabs/kube-router/pull/666/files#diff-1fd6985938391e6490ed65ff7371ebefL420) then the logic would ALWAYS return before setting up tunnels or routes: https://github.com/cloudnativelabs/kube-router/pull/666/files#diff-1fd6985938391e6490ed65ff7371ebefL448

Was the goal of #666 to introduce some type of direct routing that didn't involve either BGP or tunneling?

murali-reddy:
> Could you explain what you meant by direct routing?

Its not a standard term, but a node be able to send the traffic to another node directly with out going through any gateway or routing infrastructure.

This should not be a problem when both nodes are in same subnet **and** if you are fine with underlay network to see pod-to-pod traffic. Across the nodes in different L2 subnets we need some tunneling to send the traffic across the nodes. This is where kube-router uses IPIP tunnelling.

> Was the goal of #666 to introduce some type of direct routing that didn't involve either BGP or tunnelling?

In many environments (primarily cloud provides) as anti-spoofing measure they drop the packets if the source and destination does not match the one of the IP's of the nodes. So we can not expose pod-to-pod traffic to the underlay networks. You can overcome in some clouds, for e.g. in AWS by disabling source-destination checks. 

What #666 achieves is full overlay network (with IPIP tunnels even for nodes in same L2 subnet) for pod-to-pod traffic so underlay network will never see that traffic.

> I guess my assumption here is that anyone that would choose --enable-overlay=false would be relying on BGP to route their pod and service traffic.

Not really. 

I want to cater to case, when you have simple cluster deployed in a single L2 subnet (like a single AZ of AWS) when hosts send traffic directly. This is what we have before 0.2.5, things would work for this case even if   `--enable-overlay=false`

I wonder if we need to make this conceptual model simple with better terms and naming of options. These combinations have organically grown over last two years.

aauren:
@murali-reddy I think I'm finally beginning to understand what you're intending. I think that my trouble in understanding is that I thought that the way the code was in 0.25 was the way that you wanted it and then reading the code from 0.25 was causing a problem in understanding the use-case you wanted to support.

So I have force-pushed to my branch to resolve the issue. This will continue to place routes in the local routing table when the destination is within the same subnet regardless of the enableOverlay option. However, it will not create the IPIP tunnels when the enableOverlay option is disabled.

I think that it would be a good idea to revisit both the options and their names as well as the logic in this function as there are a couple of things flow wise that could be improved, but for now this should fix the regression without voiding the use-case you defined.

Let me know what you think.

murali-reddy:
> I think that it would be a good idea to revisit both the options and their names as well as the logic in this function as there are a couple of things flow wise that could be improved, but for now this should fix the regression without voiding the use-case you defined.

Sure, we need to do that. Changes look good to me.

PR #732, opened 2019-05-12
handle null *route

murali-reddy:
handle case when `route *netlink.Route` is not set in `injectRoute`


murali-reddy:
verified on multi AZ cluster on AWS with all combinations (--enable-overlay set to true/false, --overlay-type set to subnet/full)

PR #733, opened 2019-05-13
Fix PMTU discovery for service IPs

### LINKED ISSUE #685: kube-router iptables rule for ICMP breaks PMTU
The following rule the kube-router adds by default:

`ACCEPT     icmp --  0.0.0.0/0            0.0.0.0/0            /* allow icmp echo requests to service IPs */ icmptype 8`

breaks ICMP unreachable - must fragment:

```02:09:46.233125 IP 10.104.10.66.4506 > 10.100.27.10.37742: Flags [.], seq 108:1556, ack 1161, win 244, options [nop,nop,TS val 90047081 ecr 302705102], length 1448
02:09:46.238984 IP 10.104.27.1 > 10.104.10.66: ICMP 10.100.27.10 unreachable - need to frag (mtu 1476), length 36
02:09:46.769167 IP 10.104.10.66.4506 > 10.100.27.10.37742: Flags [.], seq 108:1556, ack 1161, win 244, options [nop,nop,TS val 90047215 ecr 302705102], length 1448
02:09:46.769416 IP 10.104.27.1 > 10.104.10.66: ICMP 10.100.27.10 unreachable - need to frag (mtu 1476), length 36
```

Proved this by adding the following rule before in chain KUBE-ROUTER-SERVICES, it works again:

`ACCEPT     icmp --  0.0.0.0/0            0.0.0.0/0`

Should change the above rule to allow unreachable as well as echo.

bazuchan:
Fix #685 for non-DSR services


asteven:
This is exactly _not_ the kind of patch I wanted.
Did you miss the comment in https://github.com/cloudnativelabs/kube-router/issues/685#issuecomment-476909931 or is there some other reason why you implemented it like this?

bazuchan:
> Did you miss the comment in #685 (comment) or is there some other reason why you implemented it like this?

IMHO either you leave ALL the rest to host firewall, either you reject ALL the rest. Personally I would prefer to have no REJECTs and leave ability for local customization. But currently all other traffic to service IPs is rejected, so it would be consistent to reject all other icmp traffic too.

asteven:
We have to reject traffic for service IP's because we open up that hole ourself in the first place.
IMHO ICMP handling is different. I typically want that per host, not per IP.

Problem is that iptables is not really meant to be controlled from multiple sources :-(

aauren:
This is reasonable and only extends the ICMP use-case that we are already providing for by allowing some additional types that are extremely valuable for basic networking.

Issue #622, opened 2018-12-27
armhf support?


cjbottaro:
Trying to get this to run on Rasbian Stretch which is `armhf`. I saw this commit https://github.com/cloudnativelabs/kube-router/commit/c7ce39836788d66b634b3c7c4f4c935468394bad, but that only looks to do `arm` and `arm64`?

Thanks for the help!

cjbottaro:
Here's what I found out...

There is support for `arm` and `arm64` but cloudnativelabs doesn't host those images on Docker Hub. This guy does though: https://hub.docker.com/r/xjjo/kube-router/tags

I have a non-homogenous Kubernetes cluster; some nodes are `arm` and some are `amd64`. So taking inspiration from the flannel installation manifests, I had to make multiple kube-router DaemonSet manifests: `kube-router-arm` and `kube-router-amd64` using nodeSelector to put each one nodes of the the right architecture.

It would be really nice if cloudnativelabs hosted images of each supported arch on Docker Hub and altered the install manifests to include them. I can take care of that latter part if there is interest.

cjbottaro:
Here ya go: https://gist.github.com/cjbottaro/5dfaef75a212a798c73b865c168ef19f

Couple of things to note:
* There are multiple DaemonSet manifests in that file
* hostPort is enabled

hakan458:
@cjbottaro Can you provide a little more detail? What controls which one gets run? I see you have a line like 
```
      nodeSelector:
        beta.kubernetes.io/arch: arm
```
But not sure how it gets routed to one or the other? Is there one daemonset file which then points to the arm/amd64 one?

Issue #591, opened 2018-11-30
arm64 support


runningman84:
It looks like kube-router supports arm64:
https://github.com/cloudnativelabs/kube-router/commit/c7ce39836788d66b634b3c7c4f4c935468394bad

But the image and or the daemonset kubeadm-kuberouter-all-features.yaml does not work on arm64:

```
root@lepotato:~# kubectl describe pod kube-router-7tvjz -n kube-system
Name:               kube-router-7tvjz
Namespace:          kube-system
Priority:           0
PriorityClassName:  <none>
Node:               lepotato/192.168.178.14
Start Time:         Fri, 30 Nov 2018 22:49:04 +0000
Labels:             controller-revision-hash=c5fb78f6f
                    k8s-app=kube-router
                    pod-template-generation=1
                    tier=node
Annotations:        scheduler.alpha.kubernetes.io/critical-pod: 
Status:             Running
IP:                 192.168.178.14
Controlled By:      DaemonSet/kube-router
Init Containers:
  install-cni:
    Container ID:  docker://be2647b83d22657f7a35bfe87cddbc18bdd0ca29d0e6749e889672b822e64058
    Image:         busybox
    Image ID:      docker-pullable://busybox@sha256:2a03a6059f21e150ae84b0973863609494aad70f0a80eaeb64bddd8d92465812
    Port:          <none>
    Host Port:     <none>
    Command:
      /bin/sh
      -c
      set -e -x; if [ ! -f /etc/cni/net.d/10-kuberouter.conf ]; then TMP=/etc/cni/net.d/.tmp-kuberouter-cfg; cp /etc/kube-router/cni-conf.json ${TMP}; mv ${TMP} /etc/cni/net.d/10-kuberouter.conf; fi
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Fri, 30 Nov 2018 22:49:08 +0000
      Finished:     Fri, 30 Nov 2018 22:49:09 +0000
    Ready:          True
    Restart Count:  0
    Environment:    <none>
    Mounts:
      /etc/cni/net.d from cni-conf-dir (rw)
      /etc/kube-router from kube-router-cfg (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from kube-router-token-mv6vl (ro)
Containers:
  kube-router:
    Container ID:  docker://e8655061df1f71fb2bfaba3a432c96e75dea2faa48cf7ff24c2197b5fb680728
[...]

PR #735, opened 2019-05-15
First stab at pushing multiarch releases

wilmardo:
Closes: #622 #591

I tried to understand the build and push process and adapted it so it should push and build multiarch.

Added the `push-manifest` in the makefile, it uses the [estesp/manifest-tool](https://github.com/estesp/manifest-tool) to push the manifest for the release and latest tag.

It should work but I haven't test it :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ go:
   - 1.10.8
 
 branches:
-  only: 
+  only:
     - master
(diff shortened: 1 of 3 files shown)


murali-reddy:
thanks @wilmardo 

I can not test it now. If i face an issue during release will address issues if any at that point.

PR #740, opened 2019-05-28
New Feature: CLI option to automatically configure Route Reflector servers and clients

bincyber:
#### What type of PR is this?

New feature

kube-router currently supports enabling BGP Route Reflection by annotating Kubernetes nodes to be Route Reflector servers or clients. The problem is cluster provisioning tools such as `kubeadm` and `kops` do not support adding annotations to nodes during bootstrapping.

What if this could be done in a more Kubernetes native way? Kubernetes master nodes can be Route Reflector servers as they have a more permanent presence in the cluster. And Kubernetes worker nodes should only be Route Reflector clients due to their ephemeral nature.

This PR adds a new command-line option: `--route-reflector-mode.` This can be set to `manual` or `auto`.
* `manual` mode is the default and keeps the existing method of configuring Route-Reflector servers and clients by checking annotations on node objects.
* `auto` mode checks if a node is a Kubernetes master node by verifying it has the label `node-role.kubernetes.io/master`. If it has this label, the node is configured as as a Route Reflector server, otherwise it will be configured as a Route Reflector client.

The label `node-role.kubernetes.io/master` is automatically set on Kubernetes master nodes by all cluster provisioning tools.

### DIFF
--- a/docs/bgp.md
+++ b/docs/bgp.md
@@ -37,14 +37,19 @@ get peered.
 
 ### Route-Reflector setup  Without Full Mesh
 
-This model support the common scheme of using Route Reflector Server node to concentrate
-peering from Client Peer. This has the big advantage of not needing full mesh, and
-scale better. In this mode kube-router expects each node is configured either in
-Route Reflector server mode or in Route Reflector client mode. This is done
-with node `kube-router.io/rr.server=ClusterID`, `kube-router.io/rr.client=ClusterId`
-respectively. In this mode each Route Reflector Client will only peer with Route
-Reflector Servers. Each Route Reflector Server will peer other Route Reflector
-Server and with Route Reflector Clients enabling reflection.
+This model supports the common scheme of using Route Reflector Server node to concentrate
+peering from Client Peers. This has the big advantage of not needing full mesh, and
+scales better. kube-router expects that each node is configured either in Route Reflector server mode
+or in Route Reflector client mode. Each Route Reflector Client will only peer with Route
+Reflector Servers. Each Route Reflector Server will peer with other Route Reflector
+Servers and Route Reflector Clients with reflection enabled. This is achieved in one of 2 ways:
+
+1. Manual Mode
+
+If kube-router is started with `--route-reflector-mode=manual` which is the default,
+kube-router checks the Kubernetes node object for the annotations `kube-router.io/rr.server=ClusterID`
+and `kube-router.io/rr.client=ClusterId` to be configured in Route Reflector server mode or
+Route Reflector client mode respectively.
 
 Users can annotate node objects with the following command:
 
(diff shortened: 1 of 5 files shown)


bincyber:
Pushed container image with these changes to Docker Hub: `bincyber/kube-router-git:route-reflector-mode`

bincyber:
@murali-reddy @ut0mt8 can you please review this PR?

murali-reddy:
@bincyber thanks for the PR. Will do the review.

murali-reddy:
While this PR addresses the problem, I am wondering if we are having too opinionated way of selecting route reflector servers?

> The problem is cluster provisioning tools such as kubeadm and kops do not support adding annotations to nodes during bootstrapping.

this is generic chicken and egg problem for any daemon-set that rely on node annotations etc. has there been any effort in Kubernetes community to address this?

bincyber:
> While this PR addresses the problem, I am wondering if we are having too opinionated way of selecting route reflector servers?

I agree that it is a bit of an opinionated approach. That is why the default should remain `--route-reflector-mode=manual` and users can opt-in to automatically configuring their kubemaster nodes as route reflector servers.

> this is generic chicken and egg problem for any daemon-set that rely on node annotations etc. has there been any effort in Kubernetes community to address this?

From my discussions with the kubeadm maintainers, adding custom annotations to nodes during cluster provisioning is not something they are currently interested in implementing. Kops has 2 closed issues that have requested this feature: https://github.com/kubernetes/kops/issues/3344 and https://github.com/kubernetes/kops/issues/6026. The Kubernetes community is currently working around this by wrapping these tools (kubeadm, kops, etc.) with configuration management tools (Puppet, Chef, Ansible, etc.).

bincyber:
@murali-reddy I could make use of this feature in my self-hosted, on-prem Kubernetes clusters. Do you want to merge this PR or should I close it?

jraby:
I like this, I actually implemented something similar locally but then threw it out.

One thing however, automatically using the masters as rr servers seems very opinionated.
Having a separate label for setting the mode to rr server would be more flexible, while still putting less administrative burden (annotating a small subset of nodes instead of all of them) on cluster operators.

We ended up using 2 daemonsets, one for servers, one for clients, but we also had to annotate every node with ASNs and clusterID, which is a bit redundant.

For reference, rr servers use the following node selector:
```
    spec:
      nodeSelector:
        kube-router.io/rr.server: ""
```

and clients  use this :
```
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: kube-router.io/rr.server
                operator: DoesNotExist
```

ut0mt8:
Interesting. Originally when I wrote this patch I was thinking at something like that, aka auto select route select mode based on annotation master presence, but think it was too opinionated.
Also I've planned to add some command line option but was too lazy to implement it and finally decide to make the simplest patch possible. 
Anyway I am happy that some people use this feature. It serve us well for some years now.

Issue #725, opened 2019-04-29
Missing NPC heartbeat because of interfering Pod updates


Cytrian:
We are experiencing regular kube-router failures. The health check sometimes fails.
I could nail it down to this situation:

Normally network_policy_controller Run() executes Sync() every 5 minutes, regulated by a Ticker.
After executing Sync() it sends a NPC heartbeat.

On pod updates, OnPodUpdate() also executes Sync().

If OnPodUpdates() executes Sync() just before the heartbeat is expected by healthcheck and Sync() takes too long, the health_controller misses a heartbeat.

TLDR: Run() cannot execute Sync()  when Sync() is already started by OnPodUpdate().

Here's a shortened/commented log of this situation.
```
# Sync() started by Run() at 13:11:37
I0429 13:11:37.919713       1 network_policy_controller.go:234] Starting sync of iptables with version: 1556543497919691345
I0429 13:11:38.696038       1 network_services_controller.go:1499] Successfully synced iptables masquerad rule
I0429 13:11:38.915911       1 network_services_controller.go:1163] IPVS servers and services are synced to desired state
I0429 13:11:38.915972       1 network_services_controller.go:799] sync ipvs services took 217.201384ms
I0429 13:11:38.918723       1 health_controller.go:74] Received heartbeat from NSC
I0429 13:11:39.826326       1 network_routes_controller.go:259] Syncing ipsets
I0429 13:11:40.950845       1 network_routes_controller.go:278] Performing periodic sync of service VIP routes
I0429 13:11:40.951166       1 network_routes_controller.go:282] Performing periodic sync of pod CIDR routes
I0429 13:11:40.960030       1 health_controller.go:74] Received heartbeat from NRC
I0429 13:11:48.838850       1 network_policy_controller.go:340] Iptables chains in the filter table are synchronized with the network policies.
I0429 13:11:48.838873       1 network_policy_controller.go:276] Syncing network policy chains took 10.916588158s
I0429 13:11:55.630715       1 network_policy_controller.go:231] sync iptables took 17.711018071s

# heartbeat sent in Run() at 13:11:55. Next heartbeat is expected at 14:16:55
I0429 13:11:55.630735       1 health_controller.go:74] Received heartbeat from NPC
I0429 13:15:05.057334       1 network_policy_controller.go:173] Received update to pod: stage-intelligence/flink-fluber-stage-controller-cron-1556543700-7pwf9
[...]

PR #741, opened 2019-05-31
Send heartbeats during NetworkPolicy and NetworkService sync.

ticpu:
In reference to issue #725, we modified kube-router to send
heartbeats during policy sync to prevent missing heartbeats
while running iptables commands.

This commit is the first part to make kube-router faster and
more robust when applying network policies.

Signed-off-by: Jérôme Poulin <jeromepoulin@gmail.com>

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -58,6 +58,7 @@ type NetworkPolicyController struct {
 	MetricsEnabled  bool
 	v1NetworkPolicy bool
 	readyForUpdates bool
+	healthChan      chan<- *healthcheck.ControllerHeartbeat
 
 	// list of all active network policies expressed as networkPolicyInfo
 	networkPoliciesInfo *[]networkPolicyInfo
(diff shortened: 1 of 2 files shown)


ticpu:
Sorry for the spam, I had some issue setting up the repo on my work PC.

This pull request is meant to be a quick fix until we make the changes to use iptables-save/restore so that we can get iptables transactions instead of running the iptables command twice for each elements.
It would also be possible to use this method for ipset has it also have a save/restore command.

@danboucher78 will be reviewing the changes I made and I will make sure this holds good for at least one day in our dev environment before I make this pull request ready for review.

ticpu:
I see pull request #680 is largely related to what I'm trying to do, but the fact that is restarts multiple time every day might warrant for this pull request to go through until then.

murali-reddy:
@roffe could you please take a look?

ticpu:
Since this has been in production for a while, it might be worthy to take a look until the migration to NFTables or even after in case someone would prefer the iptables handler.

Cytrian:
I'm happy about this fix! I backported it to our current 0.2.5 kube-router and it helped a lot to stabilize our cluster!

murali-reddy:
thanks for the PR @ticpu. thanks @Cytrian for confirming

>  it might be worthy to take a look until the migration to NFTables or even after in case someone would prefer the iptables handler.

 very likely we will have both (nftables, iptables)  implementation for now. 

changes LGTM

Issue #592, opened 2018-12-03
iptables >= 1.8 compatibility issue


murali-reddy:
https://github.com/kubernetes/kubernetes/issues/71305

uablrek:
I have tested basic function of `kube-router` (except network policies) with iptables v1.8.2 without problems.

I run `kube-proxy` outside any pod/container, in "main" namespace so to speak, so the host's version of iptables is used. If `kube-router` executes in a pod then the version of `iptables` (and libs) must match the version of the host. I don't think that is an easy problem to solve in a generic way.

murali-reddy:
Right there is no easy way out. Even the latest of alpine images that kube-router uses still has iptables version ` v1.6.1` packages, so for now replacing the host binary with older version is workaround.

murali-reddy:
`alpine:3.10` now has both `legacy` and `nft` version of iptables binaries bundled

aauren:
This should no longer be an issue because we now incorporate https://github.com/cloudnativelabs/kube-router/blob/master/build/image-assets/iptables-wrapper-installer.sh into our container images

PR #745, opened 2019-06-20
NFTables handler for the network policy controller (replaces #680)

ticpu:
This pull requests features an updated and rebased version of #680 by @johanot. It fixes the Dockerfile to use a more recent version of alpine that supports the nftables rules that are generated by the policy controller.

This version also make it possible to use the labels defined in podSelector with services. When you expose services, they don't instantly get exposed the the pods inside the cluster, you have to either add a rule to allow your services CIDR or expose them one by one as you would do with pods. I'd be willing to document this somewhere and I'd like feedback if I should continue that way and where to document it.

This work is currently used in production in a 9 nodes cluster and it greatly improve the network policies rollout time and make my pull request #741 obsolete too.


MarkDeckert:
This is awesome work. I will definitely be finding a way to use this when it gets released.

I'm a little bit fearful about changing default networkpolicy behavior though. Is it possible to have kube-router automatically add rules to allow the service subnet when nftables support is turned on?

It would make migration more seamless, and it's slightly unintuitive to have to do this since netpol typically ignores services. Eg. When u set up a netpol, not only is it between pods but it also references the containers port rather than the service's port.

Also bigger orgs often have many teams separately involved in a single cluster, operating independently within their own namespaces. A switch to Nftables would require coordination and retraining with every team.

Again, this is really awesome work -- but anything you can do to make this as close as possible to a clean drop and replace would be super appreciated.

murali-reddy:
@johanot Let me know if you are okay with overriding #680 

I would prefer to merge #680 (so we have due credit to the work done), then patch improvements on top. But if @johanot if fine, then we can go ahead. I am happy do testing and merge.

johanot:
@murali-reddy @ticpu Sorry for the late reply. Thanks for all the great considerations and the work you put on top of this, @ticpu. I just had a brief look, and it looks good. Haven't had the chance to test yet, though. I agree that it is necessary to have some kind of solution for services, given that the default behavior of "just-always-open" has now changed.

@murali-reddy I'm fine with this PR superseding mine. :-)

ticpu:
Since this PR is already adding an option to set the default policy for nftables when no policy apply to a pod, which default to allow right now, I'd suggest adding "allow-plus-services" as an option and make it default. This option would be explained in the user guide as "The NPC will automatically append a rule to allow the services CIDR in egress in any namespaces that uses network policies." I would also document the other options behaviour. Maybe adding that this argument only applies to nftables if it doesn't work on iptables as I didn't check the code yet.

Would that be good enough?

johanot:
@ticpu Sounds good to me. Only thing I'm left wondering is whether we should do anything to make the iptables-handler behave similarly? What are your thought on that?

ticpu:
My personal though on this is that the iptables handler does not scale in its current form and will become a burden in the future. It may also make people steer away from kube-router when they notice the apply time going up linearly when they add pods. We did consider switching to Calico but I really like how kube-router uses IPVS, BGP and other technologies I already know. As long as no one is going to rewrite the iptables handler to use iptables-save / iptables-restore, I think we should only port bugfixes back.

I don't mind making both handlers the same for this PR though, however, I suspect some day they will start to diverge.

I'll look at what is needed to make the iptables handler acts similarly to the nftables one and comment back.

arianvp:
How about documenting the differences instead? I would really like to see this work merged! Or do you think getting the two implementations into feature parity is easy enough?

murali-reddy:
@ticpu My sincere apologies for not giving enough attention to this PR. I started reviewing and testing the patch. I would to get this PR in to next release. I am just fixing some issues with CI etc for the project once that's done i will revert back with my testing result and review. thanks

ticpu:
I'm started working again on this so to get an easy way to test if nftables and iptables act the same according to network policies and exposed services. It's just that I had to build something to make that testing easier on multiple nodes.
Maybe I should get `make vagrant` functional and test with that instead. It doesn't seem to work for me it the current state.

murali-reddy:
@ticpu I went through the complete patch and overall approach to support both `iptables` and `nftables` with different handler looks good to me. We can mark the nftable support as experimental or beta support and let it stabilise (if at all there are any issues). But since network policy controller is refactored I want to be little more cautious on the iptables implementation with your change. So i am going to spend some time reviewing the iptables implementation for any possible regression.

I want to run through these high level items.

- Why do we need to monitor services and sync network policies on service add/delete/update events? Service updates has no bearing on the network policies. 
- I am not sure why we will need `--network-policy-default-action`.  Prescribed Kubernetes  way is to apply network policy that does default allow/deny
- AFAIK mixing iptables and nftbales is not recommended. So how do we deal with rest of kube-router functionality which still going to use legacy iptables.

ticpu:
Thanks for the review!

> Why do we need to monitor services and sync network policies on service add/delete/update events? Service updates has no bearing on the network policies.

Commits 7510c7f and c80ef96 can be removed from the pull request, I'll keep them in my tree for our specific use-case.  It uses the podSelector to select services to limit what pod can use specific services. This is to firewall from the inside between namespaces.  This is what we're using instead of just allowing the whole services IP range. This specific use-case seems to require modifications of the Kubernetes API to add a serviceSelector which is quite out of scope of this pull request.
Instead of those 2 commits, I'll replace this with one that allows the service-cluster-ip-range directly.

> I am not sure why we will need `--network-policy-default-action`.  Prescribed Kubernetes  way is to apply network policy that does default allow/deny

I agree with this. I'll modify it to fit with https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-policies

> AFAIK mixing iptables and nftbales is not recommended. So how do we deal with rest of kube-router functionality which still going to use legacy iptables.

From my experience in the last 6 months and for the use-case of kube-router only, it does not seem to trigger any unexpected reaction.  It has been used in 4 of our clusters without issue.  I don't think it should halt the project from transitioning everything to nftables step by step as long as it is well tested before.

murali-reddy:
@ticpu thanks for response.

Yes please do update the PR removing the mentioned commits and removing support for `--network-policy-default-action`

> From my experience in the last 6 months and for the use-case of kube-router only, it does not seem to trigger any unexpected reaction. It has been used in 4 of our clusters without issue. I don't think it should halt the project from transitioning everything to nftables step by step as long as it is well tested before.

Sure. I am not way resisting the change. In fact I all for using nftables. Reason I asked is I am going to fix https://github.com/cloudnativelabs/kube-router/issues/592 so we have opportunity to use all nftables. We could still use iptables frontend but use nftables backend so we can do smooth progression.

murali-reddy:
> Commits 7510c7f and c80ef96 can be removed from the pull request,

@ticpu please update the PR when you get a chance?

mrueg:
@ticpu @murali-reddy would be great to see this change merged. is anyone of you able to pick it up again?

ticpu:
I have two issues stopping me right now. First, I don't have an easy way to test this pull request now that all our environments are in use.  Second, removing those two commits makes it impossible to put the PR to a real workload because I can't put it in our dev environment as we use the feature I need to remove a lot.

arianvp:
>     * Why do we need to monitor services and sync network policies on service add/delete/update events? Service updates has no bearing on the network policies.

The example in the official Kubernetes tutorial uses `NetworkPolicy` to restrict access to a service https://kubernetes.io/docs/tasks/administer-cluster/declare-network-policy/

So it does sound to me like it's desired behavior that `NetworkPolicy` takes services into account no?
Or does that usecase still work as expected without adding rules for Cluster IPs explicitly?

Looking at the commit in question; it kind of looks odd to me we're using the `podSelector` to select services: https://github.com/cloudnativelabs/kube-router/commit/c80ef96c24d3a0f8e5e91c7cc40ba5a589f009b8#diff-61248fd525f15b9b270112c7242e93d27a1824b1c0f618fda2d6fd680c66bf6fR593

Doesn't that only accidentally work if it happens to be the case if  the `metadata.labels` of your service matches the `spec.podSelector` of the service? Which often is the case by convention but is definitely not a given

ticpu:
@arianvp you're right on this.  This is the main issue why this isn't merged yet.  I couldn't find a way to make this work for both my needs and accommodate Kubernetes specifications at the same time.

I need to be able to control using the firewall, which service has access to another, it seemed reasonable at first hand to use the podSelector to do such a things, then I found out there was just basically no way of doing so in Kubernetes right now, if it is a service, it is deemed public.

This means that either I have to plan a restructuring of our Kubernetes network to allow direct routing to pods (costly and time consuming with no gain for us right now), have Kubernetes implement this (in years) or remove the feature (not acceptable).

The next issue that remains is to have a backward compatible option to have the firewall not filter services by default. This is in link with what I was trying to accomplish. So in the end, removing the ability to filter services and making it accept everything to services IPs again would also fix this issue.

arianvp:
Eventhough the `ClusterIP` is public; the pods that it targets will still get the correct source IP of the pod that hit the service  so as long as your `NetworkPolicy` has the same `podSelector` as your `Service` you can still use `NetworkPolicy`s to fend off access to specific services. Wouldn't that be sufficient for your usecase?

It's just that the   `SOURCE <-> CLUSTER-IP` hop will always be allowed; and just the `CLUSTER-IP <-> DESTINATION` hop will still be blocked, as the IP of `SOURCE` is visible for `DESTINATION` eventhough there is a `CLUSTER-IP` inbetween.

Is your probem with the fact that the `SOURCE <-> CLUSTER-IP` traffic is still allowed? I wonder why as that is always "virtual" traffic that is always local to the node as the cluster-ip is not something that ever goes over a physical link.

arianvp:
Concretely; my understanding is that:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
        - name: nginx
          image: nginx
---
apiVersion: v1
kind: Service
metadata:
  name: website
spec:
  selector:
    app: nginx
  ports:
    - port: 80
---
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: access-website
spec:
  podSelector:
    matchLabels:
      app: nginx
  ingress:
  - from:
    - podSelector:
        matchLabels:
          access: "true"
```


Would deny all traffic to the `website`  ClusterIP;  even without special handling of `services`  in the network policy code in `kube-router` as   `website` service and the `access-website`  network-policy have the same pod-selector.  and the pods that back the `website` service will see the source IP of the pods that try to access the cluster-ip and thus the firewall rules of `kube-router` already trigger.

Only if a pod has a label `access: true` will a request to the `website` clusterIP succeed.

aauren:
Closing this. At this point I think its unlikely that we'll pursue nftables for a couple of reasons:
1) The CLI executables that interface us with Linux's networking functionality have mostly been converted to use nftables on the backend. This means that the runtime performance will be identical between nftables and iptables implementations so there isn't any benefit to be had there.
2) nftables does have some performance benefit in leaning into its functionality for inserting the rules. However, now that we batch iptables rules via save/restore (see #1010) we're already seeing all of iptables get updated in less than a second even in large clusters. So I don't think that we'll get much performance benefit.
3) Projects like Cilium have demonstrated that better runtime performance and integration can be had via eBPF. So if kube-router desires either, we'll likely spend the time investing there instead.

I definitely appreciate the effort that multiple users put into this request, but I think that the complexity of a half-step in the firewall technology advance isn't worth the performance gained now that the NPC is performant via batching.

Issue #675, opened 2019-02-20
update kube-router daemon set manifests to enable BGP graceful restart by default


murali-reddy:
kube-router does not have official single daemonset manifest. Given the flexibility and choices users can make with kube-router it was intended users craft their own manifest that suits them. There are reference manifests for various combinations of functionality https://github.com/cloudnativelabs/kube-router/tree/master/daemonset

However they dont have BGP graceful restart enabled by default. What that means is routes installed to FIB from the RIB can be withdrawn during rolling updates or kube-router pod crash. Though users can enable it themselves, apparently users having to notice this (data-path getting affected when kube-router pods dies/restarts) hard way as it is not enabled by default.

At lease for the cases (non baremental/non BGP) like kops provisioned cluster running on AWS it would be nice to enable bgp graceful restart by default.

ufou:
@murali-reddy is this what you had in mind? https://github.com/cloudnativelabs/kube-router/pull/747

murali-reddy:
@ufou Yes :)

aauren:
Closed via #747

PR #747, opened 2019-06-24
Enabling --bgp-graceful-restart by default when the router component is deployed via daemonset

ufou:
I believe this `fixes` this issue:

https://github.com/cloudnativelabs/kube-router/issues/675


aauren:
A belated thank you @ufou!

Issue #230, opened 2017-11-28
Support for BGP communities via node annotation


jdconti:
In advanced BGP configurations it may become necessary to advertise prefixes with specific [communities](https://www.cisco.com/c/en/us/support/docs/ip/border-gateway-protocol-bgp/26634-bgp-toc.html#communityattribute) attached.  At present we have upstream routers that attach communities upon announcement but it would be nice to advertise with the appropriate community upfront. 

Considering gobgp supports communities it should be trivial to add this support to kube-router if there's enough interest.

Cc: @sejeff

murali-reddy:
@jdconti so basically read the community value from the annotation and add it to the prefix that is getting advertised?

```
gobgp global rib add -a ipv4 10.0.0.0/24 community 100:100
```

jdconti:
Correct, with the ability to assign multiple communities (e.g. `gobgp global rib add -a ipv4 10.0.0.0/24 community 100:100,200:200`)

nehaseth:
Any plans of this being picked up ? We were hitting this problem on trying to advertise externalIPs to external BGP peers (which expects communities attached)

nehaseth:
Hi @murali-reddy, I have made a small change to enable this feature. The change enables us to add list of communities to a node (using annotations) and if configured, sets them in export policy. 

1. Please let me know if this approach is ok to go ahead ([link](https://github.com/cloudnativelabs/kube-router/compare/master...nehaseth:issue-230?expand=1)). Also, we don't have other ext-communities/ large communities addition requirement but can be added if required.

2. One issue I am seeing while testing is that - it is working when change is done over 0.3.1 tagged release. But when the change is added over latest master, the external vips' routes are getting removed from rib, adj-out.
Difference in policy (0.3.1 + change vs lastest-master + change) that I can see is the additional import policy in latest below.  Pls let me know if this is expected..

```
Name kube_router_import:
    StatementName kube_router_import_stmt0:
      Conditions:
        PrefixSet: any clusteripprefixset
      Actions:
         reject
```

aauren:
@nehaseth no, that isn't the intent of that patch. It's an unintended consequence of the code. I'm looking into it this morning after it was reported by @ticpu on #721

murali-reddy:
@nehaseth on quick glance patch looks good to me. Please go ahead with PR.

aauren:
Closed with #1082 merge

PR #752, opened 2019-07-03
Issue 230 - Support for BGP communities via node annotation

nehaseth:
For #230 
This change enables adding a list of communities to a node via annotations, which then gets configured as part of export policy for external bgp peers + svc vips(clusteripprefixset) match only.


aauren:
Closing in favor of #1082

PR #753, opened 2019-07-03
Allow setting the BGP graceful restart deferral time. See RFC4724 4.1

Cytrian:
GoBGP default value for deferral time is 360 seconds. That means that the
routes are not sent to the BGP peer until this timer is elapsed.

We painfully noticed this when we restarted kube-router in our cluster. The pod routes were first stale on the routers. After they expired it took some more minutes until they were updated again. The pods and services on the node were not reachable during this time. We also were hit hard by https://github.com/cloudnativelabs/kube-router/issues/725, which caused regular kube-router restarts.

### DIFF
--- a/pkg/controllers/routing/bgp_peers.go
+++ b/pkg/controllers/routing/bgp_peers.go
@@ -104,10 +104,12 @@ func (nrc *NetworkRoutingController) syncInternalPeers() {
 		if nrc.bgpGracefulRestart {
 			n.GracefulRestart = config.GracefulRestart{
 				Config: config.GracefulRestartConfig{
-					Enabled: true,
+					Enabled:      true,
+					DeferralTime: uint16(nrc.bgpGracefulRestartDeferralTime.Seconds()),
 				},
 				State: config.GracefulRestartState{
 					LocalRestarting: true,
+					DeferralTime:    uint16(nrc.bgpGracefulRestartDeferralTime.Seconds()),
 				},
 			}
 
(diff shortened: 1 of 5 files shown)


bhcleek:
I can confirm that this is a problem that has affected us, too. I've tested this PR in our cluster, and it works as expected.

IMO, this PR is essential given in order to be able to use the RollingUpdate strategy for the daemonset.

bengadbois:
@andrewsykim do you mind taking a look at this one?

pkg/options/options.go
```diff
@@ -130,6 +131,8 @@ func (s *KubeRouterConfig) AddFlags(fs *pflag.FlagSet) {
 		"Each node in the cluster will setup BGP peering with rest of the nodes.")
 	fs.BoolVar(&s.BGPGracefulRestart, "bgp-graceful-restart", false,
 		"Enables the BGP Graceful Restart capability so that routes are preserved on unexpected restarts")
+	fs.Uint16Var(&s.BGPGracefulRestartDeferralTime, "bgp-graceful-restart-deferral-time", 10,
```
andrewsykim: This should probably be a `DurationVar` and we should convert it to uint16 later?

pkg/controllers/routing/bgp_peers.go
```diff
@@ -105,9 +105,11 @@ func (nrc *NetworkRoutingController) syncInternalPeers() {
 			n.GracefulRestart = config.GracefulRestart{
 				Config: config.GracefulRestartConfig{
 					Enabled: true,
+					DeferralTime: nrc.bgpGracefulRestartDeferralTime,
```
andrewsykim: This is in seconds I'm guessing? Doesn't seem like gobgp documents this either.

pkg/options/options.go
andrewsykim: Or at least mention "in seconds" in the CLI help string

bhcleek:
@cytrian this change is important to me, and I'm willing to complete the changes that @andrewsykim has requested if you're busy or no longer interested. Can you please let us know whether you'd prefer that someone else pick up the work in your stead?

Cytrian:
I'll do it, it's not too difficult.

andrewsykim:
PR looks good, please squash commits into meaningful messages.

pkg/options/options.go
```diff
@@ -12,64 +12,66 @@ import (
...
-		IpvsGracefulPeriod: 30 * time.Second,
-		RoutesSyncPeriod:   5 * time.Minute,
-		EnableOverlay:      true,
-		OverlayType:        "subnet",
+		CacheSyncTimeout:               1 * time.Minute,
+		IpvsSyncPeriod:                 5 * time.Minute,
+		IPTablesSyncPeriod:             5 * time.Minute,
+		IpvsGracefulPeriod:             30 * time.Second,
+		RoutesSyncPeriod:               5 * time.Minute,
+		BGPGracefulRestartDeferralTime: 360 * time.Second,
```
andrewsykim: How was this default decided?

bhcleek: FWIW, it's the gobgp default, so is fully backward compatible with prior kube-router releases.

andrewsykim: ah, good to know thanks!

andrewsykim:
Thanks! cc @murali-reddy @jimmy-zh

PR #756, opened 2019-07-09
Fix: Send BGP Withdrawals for Service VIPs Upon Service Deletion

tompntn:
This fixes an issue where when a service is removed from Kubernetes, a BGP withdrawal UPDATE is never sent. Export policies are modified to prevent the routes from appearing in `adj-out`, but they are still present in the RIB.

The root of the issue is that `toAdvertise` and `toWithdraw` are incorrectly computed in `getVIPsForService` for deleted services. This PR corrects that logic.

### DIFF
--- a/pkg/controllers/routing/ecmp_vip.go
+++ b/pkg/controllers/routing/ecmp_vip.go
@@ -322,6 +322,10 @@ func (nrc *NetworkRoutingController) getVIPsForService(svc *v1core.Service, only
 		}
 	}
 
+	if svcDeleted {
+		advertise = false
+	}
+
 	ipList := nrc.getAllVIPsForService(svc)
 
 	if !advertise {
--- a/pkg/controllers/routing/ecmp_vip.go
+++ b/pkg/controllers/routing/ecmp_vip.go
@@ -82,13 +82,13 @@ func getServiceObject(obj interface{}) (svc *v1core.Service) {
 	return
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/routing/ecmp_vip.go
```diff
@@ -141,7 +142,7 @@ func getMissingPrevGen(old, new []string) (withdrawIPs []string) {
 
 // OnServiceDelete handles the service delete updates from the kubernetes API server
 func (nrc *NetworkRoutingController) OnServiceDelete(obj interface{}) {
-	nrc.tryHandleServiceUpdate(obj, "Received event to delete service: %s/%s from watch API")
+	nrc.tryHandleServiceUpdate(obj, "Received event to delete service: %s/%s from watch API", true)
```
tompntn: I couldn't find a reliable way of determining from the `v1core.Service` object if the service was deleted, so sadly resorted to passing a flag down.

murali-reddy:
@tompntn thanks for the PR. Logic looks fine, but I was wondering, if we should make the intent more clear by explicitly handling the delete part than overloading. I mean if we can introduce `handleServiceDelete` and `tryHandleServiceDelete` to make it more explicit its for delete, then code will be more clear and simple. (It should have been like that to start with so atleast we can change it to be clean)

asteven:
This PR works just fine.

murali-reddy:
LGTM. Thanks for testing @asteven

Issue #758, opened 2019-07-15
update daemonset from "extensions/v1beta1" to "apps/v1"


murali-reddy:
1.16 wil retire `extensions/v1beta1` so update all daemon set to use `apps/v1`

MarkDeckert:
Created a PR.  When testing on a v1.14.3 cluster I noticed it also required a selector to be added.

PR #759, opened 2019-07-17
update to apps/v1 and add selector

MarkDeckert:
Fixes https://github.com/cloudnativelabs/kube-router/issues/758

Also had to add a selector for this to work for me (k8s version 1.14.3)

### DIFF
--- a/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
+++ b/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
@@ -44,7 +44,7 @@ data:
     current-context: kube-router-context
 
 ---
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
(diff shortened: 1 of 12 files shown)


murali-reddy:
@MarkDeckert thanks for the PR. `apps/v1` change is obvious. Not sure why you need label selector. What error you are running into if dont add it?

MarkDeckert:
Without selector:

```
$ kubectl apply -f kubeadm-kuberouter-all-features.yaml
configmap/kube-router-cfg unchanged
error: error validating "kubeadm-kuberouter-all-features.yaml": error validating data: ValidationError(DaemonSet.spec): missing required field "selector" in io.k8s.api.apps.v1.DaemonSetSpec; if you choose to ignore these errors, turn validation off with --validate=false
```

PR #766, opened 2019-07-31
add heartbeat

chrisUsick:
We are noticing `Network Routing Controller heartbeat missed` log messages coming from the health controller. We are wondering if this is because route syncing is taking too long on the NetworkRoutesController. It looks like this issue was addressed in  [this commit](https://github.com/cloudnativelabs/kube-router/commit/42a046b5d17ef421f5216747fdebf46074eb581d) so we are trying to apply a similar fix to the NetworkRoutesController.

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -255,6 +255,7 @@ func (nrc *NetworkRoutingController) Run(healthChan chan<- *healthcheck.Controll
 		default:
 		}
 
+		healthcheck.SendHeartBeat(healthChan, "NRC")
 		// Update ipset entries
 		if nrc.enablePodEgress || nrc.enableOverlays {
 			glog.V(1).Info("Syncing ipsets")


murali-reddy:
thanks @chrisUsick for the PR. Do you have logs for the case when `Network Routing Controller heartbeat missed`. There is amount of time taken in the logs to perform sync.

chrisUsick:
Thanks for your reply, I didn’t find timing logs for the network routing
controller, just for the services and policy controller.

aauren:
Closing this issue as this is more of a workaround (granted an established one from other controllers) and in the 1.2 release (where this would get slotted) we already have a lot of PRs already that are to work on the overall performance of the network routes controller itself which should help address this issue at a wholistic level.

If we get more user reports of this issue after the 1.2 release I'm open to re-opening it and looking to see if more frequent heartbeating is still appropriate.

chrisUsick:
Thanks for the consideration,

Cheers!

PR #721, opened 2019-04-25
Add Import Policy for Service VIPs

aauren:
@murali-reddy I finally got around to making good on an implementation for #633 based on your recommendation.

The main goal of this request is to add a BGP import policy that rejects cluster VIPs so that they don't get added to the local routing table and instead are routed via pure ECMP.

In the process of doing this, I found that there was quite a bit of logic crossover between the export policies and the import policies (all of the prefixset and neighborset stuff). So rather than duplicate the logic or make the function larger than it was already I broke `addImportPolicies` and `addExportPolicies` into their own functions. This makes it easier to read and identify them as separate functionality. They are now called from a new `AddPolicies` function that first sets up the common prefix and neighbor sets and then calls the import/export functions.

Everything that used to call `addExportPolicies` has been converted to run `AddPolicies`.

PR #767, opened 2019-08-01
Revert "Add Import Policy for Service VIPs (#721)"

aauren:
This reverts the export policy additions that were added in #721 and cause issues with export policies not being added to the local RIB and thus not advertised beyond the node.


aauren:
@murali-reddy You'll need to merge this PR or one like it to make the master branch sane again after the Import policy backfired.

aauren:
Closing as #771 is a better fix to this issue.

Issue #630, opened 2019-01-10
kube-router: Direct Server Return (tunneling mode) can't forward packet (DF flag, mtu > 1480)


napat1412:
**What happened**:
kube-router: Direct Server Return (tunneling mode) can't forward packet (DF flag, mtu > 1480).
I measure TCP bandwidth with iperf3.
`iperf3 -C <$public-ip>    //fail`
`iperf3 -C <$public-ip> -M 1441  //fail`
`iperf3 -C <$public-ip> -M 1440  //success`

----
**What you expected to happen**:
kube-router: Direct Server Return can reply normal packet (default mtu 1500).
`iperf3 -C <$public-ip>    //should be success`

---
**Question**:
Q1) Could kube-router clear DF flag before send packet to tunneling?
Q2) how to increase MTU size for tunneling interface and related interface (e.g. bridge interface or physical interface)?
---
**How to reproduce it (as minimally and precisely as possible)**:
I deploy kube-router following [kube-router: DSR](https://github.com/cloudnativelabs/kube-router/blob/master/docs/user-guide.md#direct-server-return)
I deploy iperf3 service on kubernetes

```
kind: Deployment
metadata:
  name: iperf
  namespace: default
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: iperf
    spec:
      containers:
      - name: iperf
        image: "networkstatic/iperf3"
        args: ["-s"]
        ports:
          - name: iperf
            containerPort: 5201
            protocol: TCP
---
apiVersion: v1
kind: Service
metadata:
  #annotations:
  #  kube-router.io/service.dsr: tunnel
  name: iperf
  namespace: default
  labels:
    app: iperf
spec:
  selector:
    app: iperf
  externalIPs: [$public-ip]
  ports:
  - name: iperf
    port: 5201
    targetPort: 5201
    protocol: TCP
---
apiVersion: v1
kind: Service
metadata:
  annotations:
    kube-router.io/service.dsr: tunnel
  name: iperf-dsr
  namespace: default
  labels:
    app: iperf
spec:
  selector:
    app: iperf
  externalIPs: [$public-ip]
  ports:
  - name: iperf
    port: 5201
    targetPort: 5201
    protocol: TCP
```

icefed:
my fork fixed this problem by set mss with iptables.

murali-reddy:
@icefed would you mind raising a PR?

icefed:
@murali-reddy PR opened.
[...]

PR #769, opened 2019-08-14
Fix DSR(tunneling) mode mtu limit

icefed:
fix https://github.com/cloudnativelabs/kube-router/issues/630

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -2063,6 +2063,18 @@ func setupMangleTableRule(ip string, protocol string, port string, fwmark string
 	if err != nil {
 		return errors.New("Failed to run iptables command to set up FWMARK due to " + err.Error())
 	}
+
+	// setup iptables rule TCPMSS for DSR mode to fix mtu problem
+	mtuArgs := []string{"-d", ip, "-m", "tcp", "-p", "tcp", "--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS", "--set-mss", "1440"}
+	err = iptablesCmdHandler.AppendUnique("mangle", "PREROUTING", mtuArgs...)
+	if err != nil {
+		return errors.New("Failed to run iptables command to set up TCPMSS due to " + err.Error())
+	}
+	mtuArgs[0] = "-s"
+	err = iptablesCmdHandler.AppendUnique("mangle", "POSTROUTING", mtuArgs...)
+	if err != nil {
+		return errors.New("Failed to run iptables command to set up TCPMSS due to " + err.Error())
+	}
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


rearden-steel:
@murali-reddy any news on merging it??

murali-reddy:
@rearden-steel Will take a look at the fix next week.

rearden-steel:
@murali-reddy please check it, this issue affects our production environment.
Thank you.

murali-reddy:
@rearden-steel Sure. I dont have first hand knowledge of solutions that can address the problem. So I had to educate a bit. Did you run the fix in prodcution successfully?

@icefed sorry for getting back on this PR embarrassingly late. From what I understand approach in this fix is similar to MSS clamping done by middleboxes. How did you come up with 1440 value for MSS?

icefed:
@murali-reddy In the TCP/IP protocal, IP Header and TCP Header both use 20bytes, if client and server have same MTU like 1500, the MSS which is exchanged by TCP HandShake only have 1460bytes; But DSR(tunneling)ip-ip use 20bytes, the packet is biger than server MTU. 
Make client side MTU lower 20bytes than server (client 1480 and server 1500)is a solution, but we can not do that. So we can set MSS to 1440  that make 20bytes space for ip-ip tunnel.

murali-reddy:
@icefed thanks for explanation. 

I was wondering in case of cross subnet/zone traffic we will end up traffic double encapsulated So when client access external IP on node A, it get IPIP encapsulated to say pod B running on node B. If node A and node B are in different subnet then kube-router uses IP-IP tunneling. Did you by any chance try the solution on multi subnet scenario?

icefed:
@murali-reddy Sorry, I have no resource to test now.

MrFreezeex:
Will this works if the IP packet has options ? Not sure if this is relevant though...

icefed:
@murali-reddy I think 1400 is a acceptable MSS value for all network env, also need someone to test.

qingkunl:
I've tested this change. It works for the case where kube-bridge and pod eth0 have 1500 as MTU.

However, v1.1.0 has added --auto-mtu flag (https://github.com/cloudnativelabs/kube-router/blob/master/docs/user-guide.md#mtu), which might set kube-bridge and pod MTU to a different value.

Is that possible to calculate the MSS in this change based on kube-bridge and pod MTU value, instead of hardcoding it to 1440?

Thanks

qingkunl:
> @murali-reddy I think 1400 is a acceptable MSS value for all network env, also need someone to test.

@icefed Yeah, since the cross subnet tunnel has 1460 MTU, 1400 MSS works. We've tested in our production.

yydzhou:
@aauren @murali-reddy is there any progress or timeline on this? This is a desired fix for me too. We have cross sbunet tunnel setup and need to set the mss to 1400.

aauren:
@icefed Thanks for your work on this!

I cherry-picked your commit into #1063 and added additional functionality to automatically find the TCPMSS based upon interface MTU.

Closing this as it is replaced by #1063

PR #771, opened 2019-08-15
fix clusteripprefixset import policy

tamihiro:
This patch is a fix for each kube-router to properly originate `clusteripprefixset` (next-hop-self) injected by controller while still rejecting to import the same NLRI advertised from peers, so it can export the originated paths to eBGP peer(s). 

- without this patch:
```
Name kube_router_import:
    StatementName kube_router_import_stmt0:
      Conditions:
        PrefixSet: any clusteripprefixset
      Actions:
         reject
```

- with this patch:
```
Name kube_router_import:
    StatementName kube_router_import_stmt0:
      Conditions:
        PrefixSet: any clusteripprefixset
        NeighborSet: any allpeerset
      Actions:
         reject
```


murali-reddy:
changes lgtm. thanks @aauren for testing.

PR #777, opened 2019-09-05
start peering connection to neighbors from node's advertise-ip

tamihiro:
This patch explicitly sets `local-address` with node's advertise-ip for each neighbor configuration.

### DIFF
--- a/pkg/controllers/routing/bgp_peers.go
+++ b/pkg/controllers/routing/bgp_peers.go
@@ -91,13 +91,18 @@ func (nrc *NetworkRoutingController) syncInternalPeers() {
 
 		currentNodes = append(currentNodes, nodeIP.String())
 		nrc.activeNodes[nodeIP.String()] = true
+		// explicitly set neighbors.transport.config.local-address with nodeIP which is configured
+		// as their neighbor address at the remote peers.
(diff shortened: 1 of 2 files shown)


aauren:
@tamihiro Have you tested these changes in your cluster and found that they fix #778 ?

@murali-reddy The idea behind this work seems like it might be a reasonable fit for 1.1, do you have any thoughts on this?

tamihiro:
> @tamihiro Have you tested these changes in your cluster and found that they fix #778 ?

Yes this fixes #778. 

kube-routers in my cluster need this and also #779 to work properly.
https://github.com/tamihiro/kube-router/commits/mydev02

aauren:
@tamihiro Sorry, that this is taking so long. Recently, the project has had other priorities.

RE: https://github.com/cloudnativelabs/kube-router/issues/778#issuecomment-817393713 I'm seeking Murali's thoughts. You can wait for that to go through, or if you're willing it would help if you would proactively rebase so this is ready whenever I hear back.

tamihiro:
@aauren  Sorry for taking it so long, I did rebase the current master.

aauren:
@tamihiro Thanks for doing all of this work. Sorry it took so long for us to finally come back around to this!

Issue #778, opened 2019-09-05
gobgp's neighbor config needs local-address 


tamihiro:
My lab setup has nodes with multiple L3 interfaces, and each node advertises from a loopback interface address to others.

- node3:
```
$ ip a l | grep -E '^[1-9]|inet\b'
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    inet 127.0.0.1/8 scope host lo
    inet 192.168.0.13/32 scope global lo
2: ens3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    inet 192.168.10.251/29 brd 192.168.10.255 scope global ens3
3: ens4: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    inet 192.168.30.251/29 brd 192.168.30.255 scope global ens4
4: kube-bridge: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    inet 10.97.0.1/24 scope global kube-bridge
5: veth109580db@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master kube-bridge state UP group default 
6: kube-dummy-if: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN group default 
    inet 10.96.0.10/32 brd 10.96.0.10 scope link kube-dummy-if
    inet 10.96.0.1/32 brd 10.96.0.1 scope link kube-dummy-if
7: tunl0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN group default qlen 1000
8: tun-192168014@lo: <POINTOPOINT,NOARP,UP,LOWER_UP> mtu 65496 qdisc noqueue state UNKNOWN group default qlen 1000
```

- node4:
```
$ ip a l | grep -E '^[1-9]|inet\b'
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    inet 127.0.0.1/8 scope host lo
    inet 192.168.0.14/32 scope global lo
[...]

PR #779, opened 2019-09-09
skip binding device to ipip tunnel if node's interface is loopback

tamihiro:
When tunnel interface's source address is on loopback, the tunnel should not be bound  to `lo`. Otherwise, packets cannot leave from egress interface to the tunnel peer.


aauren:
@murali-reddy I'd appreciate your thoughts on this one also

aauren:
@tamihiro Sorry, that this is taking so long. Recently, the project has had other priorities.

RE: https://github.com/cloudnativelabs/kube-router/issues/778#issuecomment-817393713 I'm seeking Murali's thoughts. You can wait for that to go through, or if you're willing it would help if you would proactively rebase so this is ready whenever I hear back.

tamihiro:
@aauren I rebased this one as well  as #777, working good on my dev cluster.

aauren:
@tamihiro Thanks for doing all of this work. Sorry it took so long for us to finally come back around to this!

PR #786, opened 2019-09-24
Allow to configure RR cluster id using IPv4 strings

jdrahos:
The type of bgpClusterID field was changed to string in order to hold IPv4 and integer string representations from the annotations.
When parsing of the integer cluster if fails, IPv4 parsing will be attempted before exiting with error.

### DIFF
--- a/pkg/controllers/routing/bgp_peers.go
+++ b/pkg/controllers/routing/bgp_peers.go
@@ -146,11 +146,11 @@ func (nrc *NetworkRoutingController) syncInternalPeers() {
 				n.RouteReflector = config.RouteReflector{
 					Config: config.RouteReflectorConfig{
 						RouteReflectorClient:    true,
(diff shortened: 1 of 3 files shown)


andrewsykim:
Overall this looks fine, will review more thoroughly soon

aauren:
This seems reasonable to me, I noticed that there appears to be something in here that causes unit tests to fail that I still want to look into, but overall this seems fine.

@murali-reddy any objections to me merging this as part of the 1.1 BGP work?

aauren:
Ahh... I found out it's only failing because it's based on an old master. Once I rebased it locally, the tests started passing again.

murali-reddy:
> @murali-reddy any objections to me merging this as part of the 1.1 BGP work?

@aauren No. Please go ahead with review and merge.

aauren:
LGTM

PR #787, opened 2019-09-26
[Closed and see #920] Add large-scale network optimization and advertise-cluster-subnet function modules and documents.

houseonline:
Each kubernetes cluster in our production environment has 4000 nodes, and the whole network is interconnected by BGP, which has been running stably for more than one year. There are many problems with kube-router in the large kubernetes cluster, and we have done a lot of optimization, so I want to contribute some information to the community.I have contributed an enhanced function in the large kubernetes cluster network to Kube router, as well as several practical documents about the large kubernetes cluster network.

1. Added the "advertise-cluster-subnet" flag parameter for optimizing the number of routes.
I added the "advertise-cluster-subnet" flag parameter to kube-router. When you set the parameters of'- advertise - cluster - IP = true'and'- advertise - cluster - subnet = subnet' at the same time, the kubernetes node will only announce the aggregate cluster route you specified to the on-line router device.

2. Documents for optimization of large kubernetes cluster network are compiled.
In order for your architecture to support a larger network, you need to do the following two things: 
    (3) You need to set both "--advertise-cluster-IP=true" and "--advertise-cluster-subnet=subnet" parameters.Let k8s node only notify k8s service aggregate routes to the upstream routers, reducing the service routing entries of the upstream routers.(See large-networks03 documentation).


cloudnativer:
There are some problems with this PR code, we moved to https://github.com/cloudnativelabs/kube-router/pull/920.We expect to be merged.

cloudnativer:
This PR was abolished. We have submitted a new PR, please check https://github.com/cloudnativelabs/kube-router/pull/920.

PR #796, opened 2019-10-15
[FIX] Don't ignore silently service proxy errors.

nerzhul:
Currently we can have error on service proxy if we cannot set sysctl (in my case), but those errors are return and not shown. Just show them, like other controllers.

I found this error by running service proxy as daemonset like we did for the network part. It's very annoying to have a silent fail on that part. Instead continue because we can set sysctl on the host part directly


murali-reddy:
Thanks @nerzhul for the PR.

While `glog.Errorf` at-least logs the error to give indication, but ideally it should be `glog.Fatalf` as we should not progress further at least where its fatal. Perhaps that can be addressed in a separate PR.

Issue #797, opened 2019-10-16
Add Condition to remove ACCEPT all from KUBE-ROUTER-SERVICES


DandyDeveloper:
This is a duplicate of #682. Following a change made in #602.

It would be good to get a flag to remove the ipvs Firewall chain "accept all". 

I will be adding this, I just wanted a clear feature/issue to reference.

murali-reddy:
@DandyDeveloper this can be closed now?

DandyDeveloper:
@murali-reddy Yes! How often do releases go out?

Do you have a CI in place that checks master (we can build from master until the next RC is ready).

PR #798, opened 2019-10-16
#797 Conditionally disable "Allow All" input on IPVS KUBE-ROUTER-SERVICES

DandyDeveloper:
See #797 for more details. 

fix#797
fix#682

This is to support people wanting to prevent all incoming connections from being accepted. 

In our case, we selectively whitelist items, so we would rather have this specific rule disabled. 

Can someone confirm if there's a better naming convention here for these types of flags?

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -61,6 +61,7 @@ Usage of kube-router:
       --ipvs-graceful-period duration                 The graceful period before removing destinations from IPVS services (e.g. '5s', '1m', '2h22m'). Must be greater than 0. (default 30s)
       --ipvs-graceful-termination                     Enables the experimental IPVS graceful terminaton capability
       --ipvs-sync-period duration                     The delay between ipvs config synchronizations (e.g. '5s', '1m', '2h22m'). Must be greater than 0. (default 5m0s)
+      --ipvs-allow-all-input                          Enables rule to accept all incoming traffic on the node. (default true).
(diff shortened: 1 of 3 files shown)


DandyDeveloper:
I have tried running tests against this, but it looks like the tests in master are currently broken / not working. 

Is anyone working to fix this?

DandyDeveloper:
This needs more testing. This breaks some networking.

docs/user-guide.md
```diff
@@ -61,6 +61,7 @@ Usage of kube-router:
       --ipvs-graceful-period duration                 The graceful period before removing destinations from IPVS services (e.g. '5s', '1m', '2h22m'). Must be greater than 0. (default 30s)
       --ipvs-graceful-termination                     Enables the experimental IPVS graceful terminaton capability
       --ipvs-sync-period duration                     The delay between ipvs config synchronizations (e.g. '5s', '1m', '2h22m'). Must be greater than 0. (default 5m0s)
+      --ipvs-allow-all-input                          Enables rule to accept all incoming traffic on the node. (default true).
```
andrewsykim: prefer --ipvs-deny-all which defaults to false

DandyDeveloper: @andrewsykim We may be better off just not applying the chain based on my testing. I am spending a little more time on this just to make sure it does exactly what's expected and has no adverse negative impact.

andrewsykim: thanks, appreciate the thoroughness

Issue #794, opened 2019-10-08
Request for opinion on adding more CLI flags


jraby:
We are about to deploy a custom version of kube-router on our clusters where we needed a bit more control over the default behavior.

We added a few options, namely:
 - `--peer-router-passwords-file`: to read peer password from a file (which can be sourced from a k8s secret)
 - `--bgp-graceful-restart-time`: to set the BGP graceful-restart-time to something different than gobgp's default value of 90s (in our case, we're using kube-router for podCIDRs on a baremetal cluster with nodes that will not change often, so having a long graceful restart time makes sense so routes do not go away even if kube-router is down)

And we are thinking of adding a few more:
  - `bridge-interface`: to set the name of the bridge to create / manipulate. We need this for a smooth online migration from flannel (cni0) to  kube-router. Right now we simply hardcoded it, but I'm thinking of either adding a CLI flag for it, or reading it from the CNI json config.

 - ~~`--node-asn` + `--route-reflector-{client,server}`: these are currently only settable from node annotations, but we would rather have them set via the CLI since we deploy route-reflector servers in a different deployment than the route-reflector clients (and we mostly do not use node annotations, it would keep our node deployment simple)~~ edit: this wouldn't work, annotations are required to select which nodes to peer with. 

Would these be accepted if we submitted (separate) PRs for them?
(asking before hammering  them into good shape :)  )

wu0407:
--bgp-graceful-restart-time in this commit https://github.com/cloudnativelabs/kube-router/commit/803bd9025699ef15ddd87c8189d804107198dda5

jraby:
graceful deferral time is different from graceful restart time. (see https://tools.ietf.org/html/rfc4724#page-3)

andrewsykim:
I'm not opposed to `bridge-interface` but I'm going to noodle on it a little bit first. 

Do you need it only for migration purposes?

jraby:
I needed it for migration purposes, and i still need it after.

Ill try to send something soon.

aauren:
This has been closed through #953 and #986

PR #799, opened 2019-10-18
Allow setting BGP Graceful restart time from CLI

jraby:
Default value remains the same as GoBGP (90s)

We currently use kube-router for podCIDRs only, in a baremetal environment  where nodes do not change that much. If kube-router goes down, we want to route to survive more than 90s.
Tweaking this allows us to achieve this goal.

Support for LLGR would prossibly be better and more granular, but I haven't got around to implementing the proper gobgp config in kube-router yet.

ref: https://tools.ietf.org/html/rfc4724

### DIFF
--- a/pkg/cmd/kube-router.go
+++ b/pkg/cmd/kube-router.go
@@ -143,6 +143,13 @@ func (kr *KubeRouter) Run() error {
 	}
 
 	if kr.Config.BGPGracefulRestart {
+		if kr.Config.BGPGracefulRestartTime > time.Second*4095 {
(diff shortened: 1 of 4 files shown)


wu0407:
https://github.com/cloudnativelabs/kube-router/issues/794

cloudnativer:
restart time and holdtime work together for better results.
You can see https://github.com/cloudnativelabs/kube-router/issues/922 and https://github.com/cloudnativelabs/kube-router/pull/921

aauren:
Closing this PR in favor of #953 which builds on your work with some minor fix-ups so that it can be merged.

PR #807, opened 2019-10-25
Fix connection resets during firewall sync

djboris9:
For very busy tcp connections there is a small possibility to receive
a TCP RST during the iptables sync. It happens only when `--run-firewall=true` is set.

In function `NetworkPolicyController.syncPodFirewallChains` a default `REJECT`
rule is chronologically added before the accept-`RELATED,ESTABLISHED` rule for ingress and
egress connections. In between of the creation of these two rules a connection reset can happen for already established connections.

The solution is to add the allow-`RELATED,ESTABLISHED` rule before the `REJECT` and
jumps to this iptables chain. I see no reason the `REJECT` rule is inserted before the `ACCEPT` rule,
but maybe I miss something.

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -748,6 +748,20 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(version string) (map[s
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


murali-reddy:
good catch @djboris9 thanks for your PR

Changes LGTM

PR #808, opened 2019-10-29
external ip for egress

jdrahos:
This pull request implements following functionality:
- external ip is specified for a node using annotation
- kube-router sets up snat egress rules in iptables instead of masquerading using node's ip
- kube-router advertises route to this node for the external ip via bgp to peers so packets sent by remote host are routed back to the node (and pod)
- if the remote address is from 10.0.0.0/8 range just masquerade (internal communication knows about node's internal ip)

Signed-off-by: Jakub Drahos <jack.drahos@gmail.com>


jdrahos:
Hey @aauren , we are calling out partners up to 800k qps. All our outgoing traffic from internal IPs is router through load balancers where it is NATed which is causing substantial load on lb cpus. In order to avoid this we implemented use of external IP for this traffic when calling remote hosts and route the traffic back to the correct kube node. Calling our partners directly allowed us to scale without the need to spend funds for more load balancing hardware.

aauren:
@jdrahos Thanks for submitting a PR for this work, however, I think that the use-case for this is too niche for kube-router to support at this time.

If you're still interested in this effort, please open an issue first to discuss the use-case further.

PR #809, opened 2019-10-31
#797 Conditionally disable "Allow All" input/chain on IPVS KUBE-ROUTER-SERVICES

DandyDeveloper:
See #797 for more details.

fix#797
fix#682

This is to support people wanting to prevent all incoming connections to the node from being accepted.

In our case, we selectively whitelist items, so we would rather have this specific rule disabled.

Perhaps in future we could make this more clever and use the CHAIN in a more creative way?


murali-reddy:
thanks @DandyDeveloper for the PR

Can i suggest to reverse flag's intent i.e.) `--ipvs-deny-all` to something like `--ipvs-permit-all` that will  be more intuitive

perhaps you could just return instead of large chunk of code in if block.

please go fmt the files modified.

DandyDeveloper:
@murali-reddy On it.

DandyDeveloper:
@murali-reddy Sorted, you mention gofmt stuff. Is this a specific spot in my code, or do you want me fixing the stuff that's been there from previous changes? 

My stuff looks Ok.

pkg/options/options.go
```diff
@@ -108,6 +109,8 @@ func (s *KubeRouterConfig) AddFlags(fs *pflag.FlagSet) {
 		"The graceful period before removing destinations from IPVS services (e.g. '5s', '1m', '2h22m'). Must be greater than 0.")
 	fs.BoolVar(&s.IpvsGracefulTermination, "ipvs-graceful-termination", false,
 		"Enables the experimental IPVS graceful terminaton capability")
+	fs.BoolVar(&s.IpvsPermitAll, "ipvs-permit-all", true,
+		"Enables rule to accept all incoming traffic on the node. Default: false.")
```
murali-reddy: No need for `Default: false` in the flags comment.  Please remove it.

Can you please rephrase to `Enables rule to accept all incoming traffic to service VIP's on the node`

docs/user-guide.md
```diff
@@ -61,6 +61,7 @@ Usage of kube-router:
       --ipvs-graceful-period duration                 The graceful period before removing destinations from IPVS services (e.g. '5s', '1m', '2h22m'). Must be greater than 0. (default 30s)
       --ipvs-graceful-termination                     Enables the experimental IPVS graceful terminaton capability
       --ipvs-sync-period duration                     The delay between ipvs config synchronizations (e.g. '5s', '1m', '2h22m'). Must be greater than 0. (default 5m0s)
+      --ipvs-permit-all                               Enables ipvs INPUT/CHAIN for traffic coming into nodes (default true)
```
murali-reddy: Can we keep it more user-friendly description without getting into implementation details like `INPUT/CHAIN` etc.

Something like `permit all inbound traffic to service VIP's`

pkg/options/options.go
DandyDeveloper: Done.

docs/user-guide.md
DandyDeveloper: Yeah I'm not the best at describing these scenarios as I'm not super privvy to CNI terminology. 

Done! Thank you.

PR #811, opened 2019-11-13
set cniVersion in 10-kuberouter.conf

bhcleek:
set `cniVersion` in `cni/10-kuberouter.conf`. `cnvVersion` is required in kubernetes 1.16 and is already set in all the manifests in the daemonset directory.

### DIFF
--- a/cni/10-kuberouter.conf
+++ b/cni/10-kuberouter.conf
@@ -1,4 +1,5 @@
 {
+    "cniVersion": "0.3.0",
     "name":"mynet",
     "type":"bridge",
     "bridge":"kube-bridge",


murali-reddy:
Good Catch. thanks for PR @bhcleek

mattlqx:
Would it be possible to get a new rc release with this? I'm having to hack around this issue to upgrade to k8s 1.16.

murali-reddy:
@mattlqx Will do a release in a day or two. This is certainly blocker for upgrade warrants a release

murali-reddy:
@mattlqx here you go https://github.com/cloudnativelabs/kube-router/releases/tag/v0.4.0-rc2

PR #819, opened 2019-12-09
restrict externalTrafficPolicy=Local interpretation only to NodePort and LoadBalancer services

### LINKED ISSUE #818: restrict externalTrafficPolicy=Local interpretation only to NodePort and LoadBalancer services
In current implementation of service proxy in kube-router externalTrafficPolicy=Local is implemented in a way that only local endpoints are added even for cluster IP's. So if there are no service endpoints running on the node locally then service is unavailable for all the pods running on the node through cluster IP. So this behaviour seriously affects inter service communication if they just use service DNS name (which resolve to cluster IP). Its surprising this issue existed for long.

`externalTrafficPolicy=Local` should make sense only for external traffic incoming to the node i.e) NodePort and LoadBalancer services w.r.t service proxy.

However when advertising the service VIP's (cluster IP, external IP, load balancer IP) VIP will be advertised only if there are local endpoints if the service is set to `externalTrafficPolicy=Local` or annotated with `kube-router.io/service.local`

cc:@aauren

murali-reddy:
- change only impacts how IPVS service and service endpoints are configured for `clusterIP` service
- current behaviour for NodePort and LoadBalancer services remains same
- current behaviour for advertising the service VIP's remains same

Fixes #818


pkg/controllers/proxy/network_services_controller.go
```diff
@@ -910,6 +905,11 @@ func (nsc *NetworkServicesController) syncIpvsServices(serviceInfoMap serviceInf
 		var clusterServiceId = generateIpPortId(svc.clusterIP.String(), svc.protocol, strconv.Itoa(svc.port))
 		activeServiceEndpointMap[clusterServiceId] = make([]string, 0)
 
+		if svc.local && !hasActiveEndpoints(svc, endpoints) {
+			glog.V(1).Infof("Skipping NodePort/LoadBalancer service %s/%s as it does not have active endpoints\n", svc.namespace, svc.name)
+			continue
```
aauren: I think that the problem is that if you `continue` here, then you end up with an IPVS service without any endpoints in it as endpoints are added below in the other section you modified.

So you need to make it to the section below, but you don't want to run anything imbetween. Probably the easiest way would be to abstract the nodeport and externalIP stuff into their own methods, then gate them by a variation of this if statement.

murali-reddy:
@aauren thanks for the review. I have addressed the issue you raised.

I have also taken this opportunity to restructure the sync logic to be more modular PTAL

murali-reddy:
Accidently merge and closed this PR. 

I have opened new PR https://github.com/cloudnativelabs/kube-router/pull/836

Issue #821, opened 2019-12-16
kube-router pods are not scheduled to NotReady nodes since 1.16


kvaps:
Hi since 1.16 nodes might have additional taints, 1.17 also implements TaintNodesByCondition feature, example taints:
```yaml
taints:
- effect: NoSchedule
  key: node.kubernetes.io/unreachable
  timeAdded: "2019-12-15T18:19:29Z"
- effect: NoExecute
  key: node.kubernetes.io/unreachable
  timeAdded: "2019-12-15T18:19:35Z"
```
this way `kube-router` daemonset will never be scheduled on those nodes becaue it have only next tolerations:
```yaml
tolerations:
- key: CriticalAddonsOnly
  operator: Exists
- effect: NoSchedule
  key: node-role.kubernetes.io/master
  operator: Exists
- effect: NoSchedule
  key: node.kubernetes.io/not-ready
  operator: Exists
```

other networking plugins already switched to use:

```yaml
tolerations:
- effect: NoSchedule
  operator: Exists
- key: CriticalAddonsOnly
  operator: Exists
- effect: NoExecute
  operator: Exists
```

bhcleek:
There are times that I've needed kube-router to _not_ run so that the node won't advertise any services via BGP; tolerating all `NoSchedule` taints would be a problem for that scenario.

PR #822, opened 2019-12-16
fix tolerations for >1.16

kvaps:
This PR sets tolerations:
```yaml
tolerations:
- effect: NoSchedule
  operator: Exists
- key: CriticalAddonsOnly
  operator: Exists
- effect: NoExecute
  operator: Exists
```

fixes https://github.com/cloudnativelabs/kube-router/issues/821

### DIFF
--- a/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
+++ b/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
@@ -136,13 +136,11 @@ spec:
           mountPath: /var/lib/kube-router          
       hostNetwork: true
       tolerations:
+      - effect: NoSchedule
+        operator: Exists
       - key: CriticalAddonsOnly
         operator: Exists
-      - effect: NoSchedule
-        key: node-role.kubernetes.io/master
-        operator: Exists
-      - effect: NoSchedule
-        key: node.kubernetes.io/not-ready
+      - effect: NoExecute
         operator: Exists
       volumes:
       - name: lib-modules
--- a/daemonset/generic-kuberouter-all-features.yaml
+++ b/daemonset/generic-kuberouter-all-features.yaml
@@ -132,13 +132,11 @@ spec:
           mountPath: /var/lib/kube-router          
       hostNetwork: true
       tolerations:
+      - effect: NoSchedule
+        operator: Exists
       - key: CriticalAddonsOnly
         operator: Exists
-      - effect: NoSchedule
-        key: node-role.kubernetes.io/master
-        operator: Exists
-      - effect: NoSchedule
-        key: node.kubernetes.io/not-ready
+      - effect: NoExecute
         operator: Exists
       volumes:
       - name: lib-modules
--- a/daemonset/generic-kuberouter-only-advertise-routes.yaml
+++ b/daemonset/generic-kuberouter-only-advertise-routes.yaml
@@ -56,13 +56,11 @@ spec:
           privileged: true
       hostNetwork: true
       tolerations:
+      - effect: NoSchedule
+        operator: Exists
       - key: CriticalAddonsOnly
         operator: Exists
-      - effect: NoSchedule
-        key: node-role.kubernetes.io/master
-        operator: Exists
-      - effect: NoSchedule
-        key: node.kubernetes.io/not-ready
+      - effect: NoExecute
         operator: Exists
 
 ---
--- a/daemonset/generic-kuberouter.yaml
+++ b/daemonset/generic-kuberouter.yaml
@@ -102,13 +102,11 @@ spec:
           name: kube-router-cfg       
       hostNetwork: true
       tolerations:
+      - effect: NoSchedule
+        operator: Exists
       - key: CriticalAddonsOnly
         operator: Exists
-      - effect: NoSchedule
-        key: node-role.kubernetes.io/master
-        operator: Exists
-      - effect: NoSchedule
-        key: node.kubernetes.io/not-ready
+      - effect: NoExecute
         operator: Exists
       volumes:
       - name: lib-modules
--- a/daemonset/kube-router-all-service-daemonset-advertise-routes.yaml
+++ b/daemonset/kube-router-all-service-daemonset-advertise-routes.yaml
@@ -101,13 +101,11 @@ spec:
           mountPath: /etc/kube-router
       hostNetwork: true
       tolerations:
+      - effect: NoSchedule
+        operator: Exists
       - key: CriticalAddonsOnly
         operator: Exists
-      - effect: NoSchedule
-        key: node-role.kubernetes.io/master
-        operator: Exists
-      - effect: NoSchedule
-        key: node.kubernetes.io/not-ready
+      - effect: NoExecute
         operator: Exists
       volumes:
       - name: lib-modules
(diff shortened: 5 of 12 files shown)


murali-reddy:
> I have had cases where I need to stop the BGP advertisements on a particular node completely so that the node doesn't receive any traffic. 

@bhcleek You raise valid use-cases but i feel we should not be piggy-backing on semantics of `NoSchedule` for that. We can always implement custom kube-router specific annotation which can control if node should do BGP advertisements to external peers.

> Tolerating all NoSchedule taint would make it difficult to reboot nodes safely without interrupting traffic unexpectedly and would also make it more difficult to keep kube-router off of a node for other reasons, too (e.g. node maintanence).

If you `drain` a node does it still receive traffic?

bhcleek:
I agree on decoupling the use cases. Until an annotation to can control whether to advertise to BGP external BGP peers or not is implement, though, this is the only lever I have to pull.

As-is, the node will still receive traffic even when it's drained, because the pod for kube-router is still running on it and kube-router is advertising that it can handle the service IPs.

murali-reddy:
Ok, so i deployed 1.16  with current tolerations  

```
    tolerations:
      - effect: NoSchedule
        key: node-role.kubernetes.io/master
        operator: Exists
      - effect: NoSchedule
        key: node.kubernetes.io/not-ready
        operator: Exists
       ...
```
existing toleration are sufficient to setup the cluster. I guess issue of `node.kubernetes.io/unreachable` happens when node is unreachable from the node controller

I am not sure we should even add toleration for this taint. I would imagine once node is ready dameon set controller should trigger pod creation on the node?

@kvaps are you running into any issue if we dont add this toleration

If at all if we do need this toleration I guess as @bhcleek mentioned we can be conservative and add toleration just for "node.kubernetes.io/unreachable"

kvaps:
@murali-reddy I have a case with using diskless-nodes, example after initial setup they can be rebooted and lose their cni config, last time they had the next toleration set:

```yaml
  - effect: NoSchedule
    key: node.kubernetes.io/not-ready
    timeAdded: "2020-01-09T05:40:05Z"
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    timeAdded: "2020-01-09T06:40:49Z"
```

kvaps:
BTW here is listed all possible conditions, we can decide which one we want to ignore:
https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/#taint-based-evictions

murali-reddy:
@kvaps @bhcleek how do we go about this?

I can not find a prescribed way/best practice on how to use tolerations. If we tolerate all `NoSchedule` does it even make sense for cases like `node.kubernetes.io/out-of-disk`?

bhcleek:
I am mostly in agreement with @kvaps that we should explicitly tolerate `nodes.kubernetes.io` conditions. If it were me, though, I'd only tolerate:
* node.kubernetes.io/not-ready
* node.kubernetes/unschedulable
because those other conditions may indicate problems that could keep kube-router from running. By only tolerating those two taints, the concern raised in #821 would be addressed.

I'm against tolerating all `NoSchedule` taints.

aauren:
Wouldn't `not-ready` or `unschedulable` also potentially indicate that it might not be safe to send traffic to the node? Once kube-router starts up, if it has been configured for routing and service proxying, it will begin to advertise non service.local VIPs and begin to proxy service traffic through the node.

Can we be sure that neither of those taints will affect it's ability to do that?

kvaps:
Hi, I just make some research of the other network plugins:

Flannel [uses](https://github.com/coreos/flannel/blob/485649719be5a21b0fe7e44f699f7acbb66137fa/Documentation/kube-flannel.yml#L166-L168):
```yaml
      tolerations:
      - operator: Exists
        effect: NoSchedule
```

Calico [uses](https://github.com/heptio/aws-quickstart/blob/3c6f6e4240cfc05447876fc08e27f24e67b1a126/scripts/calico.yaml#L172-L179):
```yaml
      tolerations:
        # Make sure calico/node gets scheduled on all nodes.
        - effect: NoSchedule
          operator: Exists
        # Mark the pod as a critical add-on for rescheduling.
        - key: CriticalAddonsOnly
          operator: Exists
        - effect: NoExecute
          operator: Exists
```

Cilium [uses](https://github.com/cilium/cilium/blob/cfeaf6e559801ea49897dc9c31221421f4da482c/install/kubernetes/quick-install.yaml#L524-L525):
```yaml
      tolerations:
      - operator: Exists
```

Weave [uses](https://github.com/weaveworks/weave/blob/ff9bcdf5613bd4931dda44783aeae9615dea9c7b/prog/weave-kube/weave-daemonset-k8s-1.7.yaml#L172-L176):
```yaml
          tolerations:
            - effect: NoSchedule
              operator: Exists
            - effect: NoExecute
              operator: Exists
```

Contiv [uses](https://github.com/contiv/vpp/blob/08d792703b96a170e688b1639454212832794ebd/k8s/contiv-vpp.yaml#L348-L355):

```yaml
      tolerations:
        # We need this to schedule on the master no matter what else is going on, so tolerate everything.
        - key: ''
          operator: Exists
          effect: ''
        # This likely isn't needed due to the above wildcard, but keep it in for now.
        - key: CriticalAddonsOnly
          operator: Exists
```

bhcleek:
@aauren daemonsets run on nodes that are unschedulable and not ready already; kube-router should be there for those daemonsets. In the clusters I manage, we use a taint that isn't used by kubernetes generally to prevent kube-router from running on a node when desired.

aauren:
@kvaps Thanks for doing research on that. I think that it seems compelling that kube-router should handle this use case given all of the other network providers that do already do so.

The only thing that I would add, is that we may still run into issues down the road that some of these other providers do not given that I think kube-router is still one of the few network providers that advertises services via BGP. And this functionality will happen regardless of whether any service pods are scheduled on it. So there may still be corner cases that some of those don't experience that we will when we send traffic to a node in this state via the BGP announcements it will make.

But I think that we should probably take those as they come, and move forward with the tolerations as is.

bhcleek:
@aauren given that kube-router's bgp configuration is sourced from node annotations, that a node can't be annotated until it exists, and that these tolerations would allow kube-router to be scheduled onto a node before it is annotated, won't that require that kube-router be manually deleted after annotating the node?

aauren:
@bhcleek No, I believe that without the annotation (which is only required when full mesh mode is deactivated) that kube-router will go into a loop of it's own (https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/routing/network_routes_controller.go#L228) as well as a backoff loop (when the network_routing_controller fails to heartbeat if you have k8s health checking enabled).

Once the annotation is applied kube-router will start correctly and stay running.

bhcleek:
Thanks for confirming, @aauren. That sets my mind at ease somewhat, but without the annotation to stop bgp advertisements, it will be difficult to stop traffic from flowing through a node if this PR is merged as-is.

I think there's an argument to be made that other cni providers' decisions have a different set of criteria than kube-router because of the bgp advertisement component.

I'm fine either way if you and @murali-reddy want to merge this in light of my concerns; my goal is only to make sure we're making decisions with all aspects considered.

aauren:
Sorry, that it has taken so long for me to wrap back around to this! Somehow it fell off my radar.

I think that this has been discussed enough, there are pros and cons to both sides, but I think that in general merging this PR will put us more in line with what other network providers are doing which I think would do a lot to adhere to the principle of least surprise. Also, as discussed in this thread there are other ways to limit when kube-router begins advertising BGP.

@bhcleek & @kvaps Thanks so much for your work on this and your long patience! Merging this.

PR #823, opened 2019-12-16
fix broken CI

murali-reddy:


murali-reddy:
thanks @DandyDeveloper for the review.

PR #826, opened 2020-01-03
Make it possible to override node IP address

ldx:
When a node has multiple IP addresses, it is useful to be able to specify which one kube-router should use. This adds a command line flag `--ip-address-override <ip>` to do that, and falls back to `utils.GetNodeIP()` if it is not specified.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -57,6 +57,7 @@ Usage of kube-router:
       --health-port uint16                            Health check port, 0 = Disabled (default 20244)
   -h, --help                                          Print usage information.
       --hostname-override string                      Overrides the NodeName of the node. Set this if kube-router is unable to determine your NodeName automatically.
+      --ip-address-override ip                        Overrides the IP address of the node. Set this if kube-router is unable to determine your node IP address automatically.
       --iptables-sync-period duration                 The delay between iptables rule synchronizations (e.g. '5s', '1m'). Must be greater than 0. (default 5m0s)
       --ipvs-graceful-period duration                 The graceful period before removing destinations from IPVS services (e.g. '5s', '1m', '2h22m'). Must be greater than 0. (default 30s)
       --ipvs-graceful-termination                     Enables the experimental IPVS graceful terminaton capability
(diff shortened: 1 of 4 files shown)


murali-reddy:
@ldx thanks for the PR.

Could you please elaborate use-case you are trying to achieve?

DandyDeveloper:
> @ldx thanks for the PR.
> 
> Could you please elaborate use-case you are trying to achieve?

I believe he's saying he has a pool of IP addresses bound to the physical node on an interface, but wants the option to define a specific one that kube-router uses, rather then finding whatever is first which is fetched by `GetNodeIP()`

But I think this is the wrong way to go about it, the node should bind to the IPadress via kubeadm/kubelet manifests/configs. Not kube-router, right? 

@ldx `GetNodeIP()` and `GetNodeObject()` don't directly interface with any host information, it's all obtain via Kubernetes and gets the node stats that way. I believe you can only bind the kubernetes services to a single interface/ip. So I don't know if this PR is warranted.

ldx:
I double-checked, and indeed, even with multiple IP addresses and/or network interfaces, the kubelet only registers one internal IP address.

My use case: I work on a virtual-kubelet provider, and would like to use kube-router as a service proxy for pods that run via VK, on another cloud instance, separate from the kubelet. Kube-router, however, does not start up if the local IP address does not match the IP address on the Node object.

ldx:
@murali-reddy ping

Right now I'm maintaining a fork with this patch. Wondering if my use case, explained above, can be addressed the way done in this PR?

aauren:
@ldx Thanks for creating a PR for us, I appreciate your willingness to contribute to kube-router.

However, your use-case seems like a very niche use-case. Without some more concrete examples where this could be beneficial, I don't think that it overcomes the complexity of additional command-line options and maintenance of the code (even small as it may be).

We strive to stick as closely as possible with upstream paradigms, of which utilizing the addresses on the node definition is one such paradigm.

If you disagree with this, or you feel like there are more valid user scenarios that would benefit from this functionality, feel free to create an issue where we can discuss more.

PR #830, opened 2020-01-14
Ignore deletion of unknown IPVS rules

### LINKED ISSUE #829: Avoid removal of IPVS rules not related to kubernetes services
I am willing to use keepalived as a loadbalancer for kubernetes. It adds an IPVS rules and kube-router in proxy mode is deleting this rule.

I suggest that kube-router should have an optional argument to pass the service cidr and avoid removing any ipvs rules not in the cidr.

MrFreezeex:
Add a `--service-cidr` argument and ignore deletion of IPVS rules not in the service cidr.

Fixes #829

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -53,6 +53,7 @@ Usage of kube-router:
       --enable-overlay                                When enable-overlay is set to true, IP-in-IP tunneling is used for pod-to-pod networking across nodes in different subnets. When set to false no tunneling is used and routing infrastructure is expected to route traffic for pod-to-pod networking across nodes in different subnets (default true)
       --enable-pod-egress                             SNAT traffic from Pods to destinations outside the cluster. (default true)
       --enable-pprof                                  Enables pprof for debugging performance and memory leak issues.
+      --excluded-cidrs strings                        Excluded CIDRs are used to exclude IPVS rules from deletion.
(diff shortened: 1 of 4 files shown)


murali-reddy:
thanks @MrFreezeex for the PR

I think we will still have problem with service `external IP`. If some one deploys a service with external IP defined then an IPVS service and endpoints are created in the IPVS. At later point if external IP is removed from the service definition we will still endup having stale ipvs entries

MrFreezeex:
Indeed, this is quite problematic...

After studying the implementation of ipvs in kube-proxy, It appears that they are using an array of excluded CIDR (see https://github.com/kubernetes/kubernetes/blob/f4db8212be53c69a27d893d6a4111422fbce8008/pkg/proxy/ipvs/proxier.go#L2039).

I will rewrite my PR to implement the same kind of behavior, this should do the job.

murali-reddy:
array of exclude CIDR sounds good to me.

MrFreezeex:
I changed my code to have an array of excluded cidrs.
Let me know if further changes are needed.

murali-reddy:
LGTM

murali-reddy:
@MrFreezeex sorry after i approved the PR. There are some conflicts to merge the PR. Can you please rebase. Sorry about that. I would to push this PR in coming release.

MrFreezeex:
I have rebased my PR.

murali-reddy:
thanks @MrFreezeex

PR #836, opened 2020-01-22
restrict externalTrafficPolicy=Local interpretation only to NodePort and LoadBalancer services

### LINKED ISSUE #818: restrict externalTrafficPolicy=Local interpretation only to NodePort and LoadBalancer services

murali-reddy:
- change only impacts how IPVS service and service endpoints are configured for clusterIP service
- current behaviour for NodePort and LoadBalancer services remains same
- current behaviour for advertising the service VIP's remains same

Fixes #818


pkg/controllers/proxy/service_endpoints_sync.go
```diff
@@ -0,0 +1,517 @@
...
+		// add IPVS remote server to the IPVS service
+		for _, endpoint := range endpoints {
+			dst := ipvs.Destination{
+				Address:       net.ParseIP(endpoint.ip),
+				AddressFamily: syscall.AF_INET,
+				Port:          uint16(endpoint.port),
+				Weight:        1,
+			}
+
+			err := nsc.ln.ipvsAddServer(ipvsClusterVipSvc, &dst)
```
aauren: @murali-reddy I tested this and overall it looks good as it concerns ClusterIP services. The only problem I found is that if the service is a local service and it has active endpoints on the node, then this code still adds all of the endpoints to the IPVS service which means that service local services could still end up with an extra hop instead of staying on the host.

If you add the following code block right above this line then it mitigates this and keeps service local traffic on the node:
```
			// Conditions on which to add an endpoint on this node:
			// 1) Service is not a local service
			// 2) Service is a local service, but has no active endpoints on this node
			// 3) Service is a local service, has active endpoints on this node, and this endpoint is one of them
			if svc.local {
				if hasActiveEndpoints(svc, endpoints) && !endpoint.isLocal {
					continue
				}
			}
```

murali-reddy: @aauren thanks for the review. I have updated the PR

PR #837, opened 2020-01-22
fix .goreleaser.yml

murali-reddy:
go releaser failing with error `field archive not found in type config.Project` chaning it to `archives`


murali-reddy:
merging its go releaser config change only

Issue #828, opened 2020-01-06
withdraw external ip in use


skyever:
If two services share the same external ip, and one has been deleted, the external ip will be widthrawn even if another service is still in use。

murali-reddy:
@skyever thanks for reporting the issue.

Sounds like a bug to me. Several services should be able to use different tuples of external IP: port.

PR #842, opened 2020-02-14
use endpoint (IP, port) tuple to track active endpoints of a service 

### LINKED ISSUE #841: LVS doesn't follow nodeport service update accordingly
Hi kube-router,

We tried v0.4.0 kube-router recently, and found issue with nodeport service.

At the beginning, we create a nodeport service and point to target port 80

```
apiVersion: v1
kind: Service
metadata:
  name: mp
  namespace: test
spec:
  clusterIP: 172.30.105.6
  externalTrafficPolicy: Cluster
  ports:
  - name: nontls
    nodePort: 30356
    port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: mp
  sessionAffinity: None
  type: NodePort
status:
  loadBalancer: {}
```

after service being created, we change target port to others and take a look on the node, the RS port doesn't follow updates.

```
### we try to change target port several times (to 81, 443, 8080) and the orphan ports are not being removed.
TCP  192.168.3.1:30356 rr
  -> 172.16.3.4:80                Masq    1      0          0
  -> 172.16.3.4:81                Masq    1      0          0
  -> 172.16.3.4:443               Masq    1      0          0
  -> 172.16.3.4:8080              Masq    1      0          0
  -> 172.16.4.17:80               Masq    1      0          0
  -> 172.16.4.17:81               Masq    1      0          0
  -> 172.16.4.17:443              Masq    1      0          0
```

any advised?

cluster version: 1.15.7
kube-router: v0.4.0

murali-reddy:
Currently only endpoint IP used so any change in port of the endpoint leaves stale ipvs server config. With this change (ip, port) tuple is used to track active endpoitns of service and to identify stale IPVS config of server of a service

also service info that is built does not carry `targetport` info, so any change to just `targetPort` in service spec was not trigerring sync

Fixes #841

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -240,6 +240,7 @@ type serviceInfo struct {
 	namespace                string
 	clusterIP                net.IP
 	port                     int
+	targetPort               string
 	protocol                 string
 	nodePort                 int
 	sessionAffinity          bool
@@ -1033,6 +1034,7 @@ func (nsc *NetworkServicesController) buildServicesInfo() serviceInfoMap {
 			svcInfo := serviceInfo{
 				clusterIP:   net.ParseIP(svc.Spec.ClusterIP),
 				port:        int(port.Port),
+				targetPort:  port.TargetPort.String(),
 				protocol:    strings.ToLower(string(port.Protocol)),
 				nodePort:    int(port.NodePort),
 				name:        svc.ObjectMeta.Name,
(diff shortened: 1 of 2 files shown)


murali-reddy:
@aauren Please review when you get a chance

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -240,6 +240,7 @@ type serviceInfo struct {
 	namespace                string
 	clusterIP                net.IP
 	port                     int
+	targetPort               string
```
mrueg: Why a string type? Shouldn't it have the same type as port?

murali-reddy: `TargetPort` in service spec is `intstr` type.

murali-reddy:
@aauren thanks for the review

> Are there any unit tests that need to be augmented to test the new port tracking?

No exisiting unit test covers this scenario. I will add unit tests in a seperate PR for this issue and for #828

PR #850, opened 2020-03-03
withdraw external IP from advertisement only if the deleted service is the last service using external IP

### LINKED ISSUE #828: withdraw external ip in use

murali-reddy:
Fixes #828

### DIFF
--- a/pkg/controllers/routing/ecmp_vip.go
+++ b/pkg/controllers/routing/ecmp_vip.go
@@ -121,15 +121,15 @@ func (nrc *NetworkRoutingController) handleServiceDelete(svc *v1core.Service) {
 		glog.Errorf("Failed to get active VIP's on service delete event due to: %s", err.Error())
 		return
 	}
-	activeVIPsMap := make(map[string]string)
+	activeVIPsMap := make(map[string]bool)
 	for _, activeVIP := range activeVIPs {
-		activeVIPsMap[activeVIP] = ""
+		activeVIPsMap[activeVIP] = true
 	}
 	serviceVIPs := nrc.getAllVIPsForService(svc)
 	withdrawVIPs := make([]string, 0)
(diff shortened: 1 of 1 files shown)


pkg/controllers/routing/ecmp_vip.go
```diff
@@ -116,7 +116,24 @@ func (nrc *NetworkRoutingController) handleServiceDelete(svc *v1core.Service) {
...
 
-	nrc.withdrawVIPs(nrc.getAllVIPsForService(svc))
+	activeVIPs, _, err := nrc.getActiveVIPs()
+	if err != nil {
+		glog.Errorf("Failed to get active VIP's on service delete event due to: %s", err.Error())
+		return
+	}
+	activeVIPsMap := make(map[string]string)
+	for _, activeVIP := range activeVIPs {
+		activeVIPsMap[activeVIP] = ""
```
aauren: Small nitpic here. Since you are only using the map to check to see if a service exists or not later on without always looping over the `activeVIPs` array, you don't actually use the string value here.

As such you could define the map as `make(map[string]bool)` and then the `if` statement becomes a bit more apparent later on with something like, since the when you reference a non-existent value from a map it defaults to a zero value of the type, in this case false:
```
if ! activeVIPsMap[serviceVIP] {
```

aauren:
Not sure why travis-ci failed it's build. Build and test from your branch all work / pass for me.

murali-reddy:
There is one [flaky test](https://github.com/cloudnativelabs/kube-router/blob/v1.0.0-rc1/pkg/controllers/routing/network_routes_controller_test.go#L2433) failing occasionnaly, which needs investigation. Rerunning the CI went fine.

Merging the PR. thanks

PR #856, opened 2020-03-16
handle DeletedFinalStateUnknown objects in DeleteFunc handlers

murali-reddy:
in DeleteFunc handlers across the controllers handle the case where received object can be of type DeletedFinalStateUnknown

fixes one of the symptoms (panic on receiving DeletedFinalStateUnknown objects) reported in #712

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -1631,7 +1631,7 @@ func (npc *NetworkPolicyController) newPodEventHandler() cache.ResourceEventHand
 			}
 		},
 		DeleteFunc: func(obj interface{}) {
-			npc.OnPodUpdate(obj)
+			npc.handlePodDelete(obj)
 		},
 	}
 }
@@ -1647,7 +1647,7 @@ func (npc *NetworkPolicyController) newNamespaceEventHandler() cache.ResourceEve
 
 		},
 		DeleteFunc: func(obj interface{}) {
-			npc.OnNamespaceUpdate(obj)
+			npc.handleNamespaceDelete(obj)
 
 		},
 	}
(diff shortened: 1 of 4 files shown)


pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -1663,12 +1663,88 @@ func (npc *NetworkPolicyController) newNetworkPolicyEventHandler() cache.Resourc
...
+			return
+		}
+	}
+	// namespace (and annotations on it) has no significance in GA ver of network policy
+	if npc.v1NetworkPolicy {
+		return
+	}
+	glog.V(2).Infof("Received namespace: %s delete event", namespace.Name)
+
+	err := npc.Sync()
```
aauren: Do we not need to check if the npc is ready for updates before we call `Sync` like we do in the other functions?
e.g.
```
if !npc.readyForUpdates
```

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -2027,32 +2016,108 @@ func (nsc *NetworkServicesController) Cleanup() {
...
+	endpoints, ok := obj.(*api.Endpoints)
+	if !ok {
+		glog.Errorf("unexpected object type: %v", obj)
+		return
+	}
+	nsc.OnEndpointsUpdate(endpoints)
+}
+
+func (nsc *NetworkServicesController) handleEndpointsUpdate(oldObj, newObj interface{}) {
+	_, ok := oldObj.(*api.Endpoints)
```
aauren: Is there a specific reason to process these at all if we don't use them? While I would expect that we would ever get this error, I would imagine that the greater benefit for the stability of the controller would be to process the update anyway assuming that we could cast the newEndpoints and send them on to the OnUpdate function.

murali-reddy:
kube-router perform below three processing activities during which it access the API server

1. - during the starting to perform initial configuration
2. - while processing the add/delete/update events corresponding to k8s api objects
3. - periodic full-sync's

Kube-router should perform gracefully in both short term network particition (that affects above case #2) and long term partition (case #3)

I have done testing to ensure kube-router perform gracefully in above combinations.

@atibdialpad Could you please help testing this patch (in general #712) 

You can use this brach of the PR (https://github.com/cloudnativelabs/kube-router/tree/issue-712) or pick the commits  d2405229f72affe027e4b536f61ccbc2aa59d124 and 33724aac05a642c1f10d88b4df661153996ce877

pkg/controllers/netpol/network_policy_controller.go
murali-reddy: @aauren thanks for the review. Sorry abouth the dealy. I updated the PR to include suggested change.

pkg/controllers/proxy/network_services_controller.go
murali-reddy: Ideally we should use both the new and old objects to check if there is an update which deserved processing rule.

Somthing like https://github.com/cloudnativelabs/kube-router/blob/v1.0.0-rc1/pkg/controllers/netpol/network_policy_controller.go#L1628

I will raise a seperate PR to address this issue.

atibdialpad:
@murali-reddy 
sure,I will test your changes and get back with results

murali-reddy:
> @murali-reddy
> sure,I will test your changes and get back with results

Please share the test results when you get to test it. Will raise a seperate PR if there are any issues still found.

PR #863, opened 2020-03-31
add option to not masquerade pod cidrs

bhcleek:
Add an option, masquerade-pod-cidrs, to allow pod cidrs to not be
masquerading is unnecessary (e.g. when pods CIDRs are routeable on a
network via BGP).

The new option defaults to true for backward compatibility.


bhcleek:
closing. There's already an option for this. I'm going to make sure it's working as expected and will submit a new PR if needed

PR #864, opened 2020-04-01
read the necessary API objects from local cache instead of listing from the API server

### LINKED ISSUE #862: kube-router skips heartbeat when api server is not reachable
kube-router skips heartbeat when api server is not reachable resulting in shutting down the kube-router pod.

While testing #856, #739  I ran into scenario where if API server is down, or node is network partitioned from master node, its observed during full sync goroutine running full sync ( of networking routing cotroller or proxy or network policy controller) gets hung for almost 15 min, Which will eventually result in missing sending heartbeat.

It turns out call to list the API object gets stuck due to https://github.com/kubernetes/client-go/issues/374. client-go issue is still open, so need a way to gracefully perform full sync (which lists all required API objects from the API serer rather than using the cache) when API server is down.

murali-reddy:
Fixes #862

this PR removes last and very few instances where API objects were directly read from the API server

### DIFF
--- a/pkg/controllers/routing/aws.go
+++ b/pkg/controllers/routing/aws.go
@@ -11,7 +11,8 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/golang/glog"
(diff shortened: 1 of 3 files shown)


murali-reddy:
thanks @aauren for review.

PR #866, opened 2020-04-03
Add the possibility to specify CIDRs that will be skipped by the MASQUERADE iptables target

FabienZouaoui:
I'm using a tunnel to encapsulate traffic between two kubernetes clusters.

Unfortunately kube-router sets up an iptables rules that MASQUERADE traffic when trying to access remote targets.
 
Using --other-known-cidrs argument to configure cidrs will populate kube-router-other-ips ipset which will be skipped by the iptables rule.
The old iptables rule is cleaned as well.


FabienZouaoui:
Closing this one to create a new PR following the guidelines [het](https://github.com/cloudnativelabs/kube-router/blob/master/CONTRIBUTING.md)

PR #867, opened 2020-04-03
Add Numberly to USERS.md

Lujeni:
Thanks again for your hard work


murali-reddy:
thanks @Lujeni. Great to know Numberly usage.

PR #869, opened 2020-04-06
fix Test_syncInternalPeers test failure due to #864

murali-reddy:
Fix unit test failure due to switch of listing node API objects from API server to cached informer. Modify test to use informer


murali-reddy:
Fixed Test_syncInternalPeers https://travis-ci.org/github/cloudnativelabs/kube-router/builds/671620113

Merging. test only fix.

Issue #748, opened 2019-06-25
add routes in the rib created by kube-router 


atibdialpad:
Is it possible to add entries in the gobgp rib created by kube-router ? I have a use case where I want to create a subinterface, say eth0:1 and puts its IP into the rib.

atibdialpad:
any ideas/comment ?

murali-reddy:
> I have a use case where I want to create a subinterface, say eth0:1 and puts its IP into the rib.

@atibdialpad can you elaborate a bit please? What do you mean by putting IP in to the RIB

atibdialpad:
Hey Murali, 
I have a use case where I need to DNAT/SNAT all traffic coming_in/going_out from a POD. This is because, the application that this POD talks to whitelists only a particular IP. So, I have this sort of a deployment:
external_app ---> DNAT ingress traffic from whitelisted_ip to pod_ip ----> POD
POD ---> SNAT egress pod traffic to whitelisted_ip ---> external IP
This makes the external application assume that it is talking to the whitelisted IPs and not the POD ip.
To advertise this whitelisted IP, I need to install this IP into the RIB.
With KubeRouter running a bgp instance, I am not able to add this.
so, right now I am running a gobgp process and installing the route via :
gobgp global rib add -a ipv4 <whitelisted_ip> nexthop <node_ip>

It would have been much easier had there been a way to install this router in the RIB created by the kube-router itself.

atibdialpad:
To add : The reason why I am not using k8s services with external IP is because egress traffic gets SNAT 'ed to NODE_IP which being a non whitelisted IP is dropped by the external application.

atibdialpad:
any ideas/comment ?

atibdialpad:
Hey @murali-reddy , 
I did some code browsing and found that kuberouter uses old version of gobgp 1.33 (https://github.com/cloudnativelabs/kube-router/blob/54eedcd0490dd28191d53971d9442f57d45da25d/Gopkg.toml) 
but latest version of grpc 
(https://github.com/cloudnativelabs/kube-router/search?q=bgpapi&unscoped_q=bgpapi). 

Because of gobgp protobuf api change post 1.33, if I use gobgp cli to manipulate rib of kube-router, it fails saying :
"rpc error: code = Internal desc = grpc: error unmarshalling request: proto: bad wiretype for field gobgpapi.Path.Age: got wiretype 2, want 0" 

What are your thoughts about it ?

[...]

PR #872, opened 2020-04-07
Bgp policy fix

atibdialpad:
Please refer to https://github.com/cloudnativelabs/kube-router/issues/748
In short, 
1. kuberouter creates bgp policies to control export and import of bgp routes to peers/other nodes etc.
2. This is done via a periodic sync where kr gets the list of advertisable IPs and calls a "ReplacePolicyAssignment" which flushes any policy that was not created by kube-router. 
3. Fix : Post an update, the code already updates the prefix and neighbor defined set objects pointed by the existing bgp policy (and thus updating the bgp policy). All it needs to do is to ensure the defaut policy is set to REJECT. It need not call "ReplacePolicyAssignment" .

### DIFF
--- a/pkg/controllers/routing/bgp_policies.go
+++ b/pkg/controllers/routing/bgp_policies.go
@@ -256,10 +256,10 @@ func (nrc *NetworkRoutingController) addExportPolicies() error {
 		}
 	} else {
 		// configure default BGP export policy to reject
-    err = nrc.bgpServer.SetDefaultPolicy("", table.POLICY_DIRECTION_EXPORT, table.ROUTE_TYPE_REJECT)
-    if err != nil {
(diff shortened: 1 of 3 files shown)


murali-reddy:
@atibdialpad thanks for your PR.

Modifying direclty vendored osrg/gobgp library is not correct way to go. Can the change be kube-router code and achieve same result?

aauren:
@atibdialpad Is this something that you have an interest in following up with in a way that doesn't change vendored code? Also, FWIW we're in the process of updating GoBGP to 2.X, in this version GoBGP no longer has the `Replace*` functions, so it's possible that your use-case will begin working after that upgrade.

aauren:
Closing from of lack of response.

PR #874, opened 2020-04-09
switch --set to less ambiguous --match-set

aauren:
While --set is still ambiguous it can clash with other module options, so it is better to be more specific and use the --match-set option. This also more closely aligns with all other areas of the code that already use --match-set.

From iptables-extensions man page:
The option --match-set can be replaced by --set if that does not clash with an option of other extensions.


murali-reddy:
thanks @aauren

PR #875, opened 2020-04-13
run network polices on pod egress traffic going through the OUTPUT chain of filter table

### LINKED ISSUE #609: "Deny all egress" policy does not restrict access to private address
**[K8s]**
version: v1.13
created with kubeadm with one master and one node

**[kube-router]**
version: v0.2.3
router,firewall and service-proxy are all enabled

**[How to produce]**
```
kubectl create ns netpol-test

cat <<EOF | kubectl apply -f -
apiVersion: extensions/v1beta1
kind: NetworkPolicy
metadata:
  name: default-deny
  namespace: netpol-test
spec:
  podSelector: {}
  policyTypes:
  - Egress
EOF

kubectl run -n netpol-test busybox --rm -ti --image=busybox /bin/sh
```

**[Current Result]**
In the busybox container, the CoreDNS and kubernetes apiserver endpoint were still accessible:
10.96.0.10 is CoreDNS address. 10.96.0.1 is the kubernetes apiserver.
```/ # ping 10.96.0.10
PING 10.96.0.10 (10.96.0.10): 56 data bytes
64 bytes from 10.96.0.10: seq=0 ttl=64 time=0.304 ms
64 bytes from 10.96.0.10: seq=1 ttl=64 time=0.141 ms

/ # wget -O - https://10.96.0.1
Connecting to 10.96.0.1 (10.96.0.1:443)
wget: note: TLS certificate validation not implemented
wget: server returned error: HTTP/1.1 403 Forbidden
```
**[Expected Result]**
All egress access should be restricted, including the service cidr, pod cidr and all other private addresses(10.0/8, 172.16/12,192.168/16)

murali-reddy:
There is a iptables rule in `filter` table `FORWARD` chain to ensure pod egress traffic is run through the network policies however this rule only takes care of pod accessing destination pod directly by its IP.

When a pod access a service which eventually gets DNAT'ed we need to add a rule in `OUTPUT` chain as well to enforce network policies. 

Fixes #609

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -879,7 +879,8 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(version string) (map[s
 		}
 
 		// ensure there is rule in filter table and FORWARD chain to jump to pod specific firewall chain
-		// this rule applies to the traffic getting routed (coming for other node pods)
+		// this rule applies to the traffic getting forwarded/routed (traffic from the pod destinted
+		// to pod on a different node)
 		comment = "rule to jump traffic from POD name:" + pod.name + " namespace: " + pod.namespace +
 			" to chain " + podFwChainName
 		args = []string{"-m", "comment", "--comment", comment, "-s", pod.ip, "-j", podFwChainName}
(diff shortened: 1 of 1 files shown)


aauren:
@murali-reddy I can confirm that this fixed the issues for us! Thanks!

I compiled this and deployed it out to one of our testing clusters today and ran tests for some 20 minutes or so. Before I put the new binary into play I could see DNS queries succeeding even without a network policy that allowed them. After I put the patched binary in place all DNS queries failed.

Then, just to make sure that it was causal I manually deleted the OUTPUT rule for that pod's IP address and was able to see DNS queries succeed until kube-router updated iptables again and put the rule back in it.

murali-reddy:
thanks @aauren for the review.

PR #876, opened 2020-04-14
Use bridge ip as source for routing rule associated with services

FabienZouaoui:
When IPSEC in transport mode is enabled between nodes, I cannot reach services IPs anymore.
When using kube-router IP instead of node local IP, everything works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .vscode
+.idea
 /kube-router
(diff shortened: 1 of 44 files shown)


aauren:
Closing this as it couples the network routes controller to the network services controller which is a design pattern that we cannot accept for now.

PR #882, opened 2020-04-22
honor the ClientIP session affinity timeout when set. 

bumyongchoi:
10800 secs (5 hours) is hardcoded for ipvs when ClientIP is set for SessionAffinity for a service. The value of "service.spec.sessionAffinityConfig.clientIP.timeoutSeconds" should be used to set the timeout instead.


murali-reddy:
thanks @bumyongchoi for your contribution. :+1: for writing unit test as well.

PR #883, opened 2020-04-22
use Spec.PolicyTypes for the type of network policy

murali-reddy:


murali-reddy:
thanks @aauren for the review

PR #884, opened 2020-04-23
mount host /run/xtables.lock to kube-router container 

### LINKED ISSUE #506: use xtables.lock
When deploying networkpolicies to two namespaces in the same cluster on Azure Kubernetes Service simultaneously we have observed the following error. It appears to be due to resource contention on the iptables command. The issue is that both deployments will complete "successfully" but the traffic flows the policy allows will fail and the error is only apparent when you view the logs of the kube-router pod/s.

```
E0801 10:27:45.538722       1 network_policy_controller.go:182] Error syncing network policy for the update to network policy: <NAMESPACE_NAME>/<POLICY_NAME> Error: Aborting sync. Failed to sync network policy chains: Failed to run iptables command: running [/sbin/iptables -t filter -N KUBE-NWPLCY-YYN4V3F74PSONPOB --wait]: exit status 4: iptables: Resource temporarily unavailable.
```

murali-reddy:
Fixes #506

### DIFF
--- a/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
+++ b/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
@@ -106,6 +106,9 @@ spec:
         - name: kubeconfig
           mountPath: /var/lib/kube-router
           readOnly: true
+        - name: xtables-lock
+          mountPath: /run/xtables.lock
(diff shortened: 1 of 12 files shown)


murali-reddy:
thanks @DandyDeveloper

PR #889, opened 2020-04-26
nflog the packet that will be dropped by network policy enforcement

### LINKED ISSUE #505: Firewall Observability
We've recently been experimenting with Kuberouter using the Azure Kubernetes Service and are finding if difficult to surface atypical iptables logging information including blocked requests; this makes it incredibly difficult to determine system traffic flows which need to be permitted or just debugging in general.

Perhaps I've missed the information in the readme/elsewhere, but it isn't clear to me what the recommended guidance is on how to debug network policies to understand what traffic is being blocked or just generally getting observability of the firewall behaviour so we can reason about its behaviour (to know that its working as expected between deployments)

murali-reddy:
logged packets can be further by read by ulogd

Fix only works with kernel > 4.11 please see https://github.com/cloudnativelabs/kube-router/issues/505#issuecomment-619633699

Fixes #505

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -817,6 +817,14 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(version string) (map[s
 			}
 		}
 
+		// add rule to log the packets that will be droopped due to network policy enforcement
+		comment = "rule to log dropped traffic POD name:" + pod.name + " namespace: " + pod.namespace
+		args = []string{"-m", "comment", "--comment", comment, "-j", "NFLOG", "--nflog-group", "100"}
+		err = iptablesCmdHandler.AppendUnique("filter", podFwChainName, args...)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to run iptables command: %s", err.Error())
+		}
+
 		// add default DROP rule at the end of chain
 		comment = "default rule to REJECT traffic destined for POD name:" + pod.name + " namespace: " + pod.namespace
 		args = []string{"-m", "comment", "--comment", comment, "-j", "REJECT"}
@@ -931,6 +939,14 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(version string) (map[s
 			if err != nil {
 				return nil, fmt.Errorf("Failed to run iptables command: %s", err.Error())
 			}
+		}
+
+		// add rule to log the packets that will be droopped due to network policy enforcement
+		comment = "rule to log dropped traffic POD name:" + pod.name + " namespace: " + pod.namespace
+		args = []string{"-m", "comment", "--comment", comment, "-j", "NFLOG", "--nflog-group", "100"}
+		err = iptablesCmdHandler.AppendUnique("filter", podFwChainName, args...)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to run iptables command: %s", err.Error())
 		}
 
 		// add default DROP rule at the end of chain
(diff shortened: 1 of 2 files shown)


pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -817,6 +817,14 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(version string) (map[s
 			}
 		}
 
+		// add rule to log the packets that will be droopped due to network policy enforcement
+		comment = "rule to log dropped traffic POD name:" + pod.name + " namespace: " + pod.namespace
```
mrueg, line 821:
```suggestion
		comment = "rule to log dropped traffic to POD name:" + pod.name + " namespace: " + pod.namespace
```

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -933,6 +941,14 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(version string) (map[s
...
```
mrueg, line 945:
```suggestion
		comment = "rule to log dropped traffic from POD name:" + pod.name + " namespace: " + pod.namespace
```

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -817,6 +817,14 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(version string) (map[s
 			}
 		}
 
+		// add rule to log the packets that will be droopped due to network policy enforcement
```
mrueg, line 820:
```suggestion
		// add rule to log the packets that will be dropped due to network policy enforcement
```

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -933,6 +941,14 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(version string) (map[s
...
```
mrueg, line 944:
```suggestion
		// add rule to log the packets that will be dropped due to network policy enforcement
```

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -817,6 +817,14 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(version string) (map[s
...
+		comment = "rule to log dropped traffic POD name:" + pod.name + " namespace: " + pod.namespace
+		args = []string{"-m", "comment", "--comment", comment, "-j", "NFLOG", "--nflog-group", "100"}
```
aauren, line 822:
```suggestion
		args = []string{"-m", "comment", "--comment", comment, "-j", "NFLOG", "--nflog-group", "100", "-m", "limit", "--limit", "10/minute", "--limit-burst", "10"}
```

I recommend that we add some limiting here. I know that a few of our users put their kube-router instances on or near the edge and it would be good to ensure we don't get overwhelmed with failed packets.

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -933,6 +941,14 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(version string) (map[s
...
```
aauren, line 946:
```suggestion
		args = []string{"-m", "comment", "--comment", comment, "-j", "NFLOG", "--nflog-group", "100", "-m", "limit", "--limit", "10/minute", "--limit-burst", "10"}
```

murali-reddy:
> Do you have any thoughts on how users can best run ulogd? Maybe a ulogd container (I'm assuming it can be run from within a container) with example config snippet?

@aauren Thank for the review. I will add an e.g in usage guide on how to enable loggiging.

For more convenience we can log in kube-router logs as well by doing like [this](https://github.com/weaveworks/weave/blob/v2.6.2/npc/metrics/metrics.go#L27)

pkg/controllers/netpol/network_policy_controller.go
murali-reddy: @mrueg thanks for the review. I started with same but actually a only a single IPtable rule gets created so unfortunaltye can not add to/from. Resolved other comment

PR #891, opened 2020-04-30
fixes egress network policy enforcements

murali-reddy:
- outbound traffic from pod should be intercepted in filter table INPUT chain (pod's traffic that is destined to node's local ip). with out this fix even with network policy to drop all egress traffic, pod can reach host IP's. Pod's can access any service hosted in host network as well

~~- rules to permit forwarding pod traffic should be appended at the end to the FORWARD chain. In some distro's default policy for FORWARD chain is drop. So explicit rule must be added in FORWARD chain for the traffic from the pods. These rules should be lower order than any network policy chains~~

removing the changes to network routing controller to append the rule to enable forwarding from the pods. Will address in seperate PR

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -880,37 +880,23 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(version string) (map[s
 			}
 		}
 
-		// ensure there is rule in filter table and FORWARD chain to jump to pod specific firewall chain
-		// this rule applies to the traffic getting forwarded/routed (traffic from the pod destinted
-		// to pod on a different node)
-		comment = "rule to jump traffic from POD name:" + pod.name + " namespace: " + pod.namespace +
-			" to chain " + podFwChainName
-		args = []string{"-m", "comment", "--comment", comment, "-s", pod.ip, "-j", podFwChainName}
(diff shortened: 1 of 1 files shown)


pkg/controllers/routing/network_routes_controller.go
```diff
@@ -613,7 +613,7 @@ func (nrc *NetworkRoutingController) enableForwarding() error {
 		return fmt.Errorf("Failed to run iptables command: %s", err.Error())
 	}
 	if !exists {
-		err := iptablesCmdHandler.Insert("filter", "FORWARD", 1, args...)
+		err := iptablesCmdHandler.Append("filter", "FORWARD", args...)
```
aauren, line 616: This doesn't really impact our use case, but felt that it's worth the reminder that this was specifically set a while back for handling with RedHat variant firewall engines in #596

aauren: Also, before the method call that was used to append was `AppendUnique`, not sure how it wouldn't be unique, but unless there is a performance cost, AppendUnique seems reasonable.

murali-reddy: thanks @aauren for the review. Good catch on #596 i totally forgot. 

I have taken out the part of the PR which was modiying the rules that enable forwarding. So it looks like based distro there may be still race condition. I am going to take it up seperate PR. For now will open an issue.

PR #892, opened 2020-05-01
service proxy: ensure hearbeats are sent during sync

### LINKED ISSUE #879: kube-router shuts down due to network service controller false negative heartbeat misses
In the current network service contoller implementation of heart beats they are sent only during the periodic syncs **only**.

https://github.com/cloudnativelabs/kube-router/blob/v1.0.0-rc1/pkg/controllers/proxy/network_services_controller.go#L393

Ideally network service controller should be sending the heartbeats when ever it performs (or during) syncs i.e.) durning periodic full sync and sync performed on service/endpont add/delete/update events.

Combined with the fact this [select-case](https://github.com/cloudnativelabs/kube-router/blob/v1.0.0-rc1/pkg/controllers/proxy/network_services_controller.go#L363) logic means if there are parallel events (add/update/delete events for the service/endpoint) there is no gaurentee full sync is performed in timeley manner resulting in missing heartbeats.

Fix is to send heartbeats (similar to network policy controller) when performing sync irrespective if its full sync or due to API event.

Snip of the logs when this issue found

```
I0330 22:46:05.009831       1 service_endpoints_sync.go:77] IPVS servers and services are synced to desired state
I0330 22:46:05.009846       1 service_endpoints_sync.go:28] sync ipvs services took 43.856920286s
I0330 22:46:05.009877       1 network_services_controller.go:377] Performing requested sync of ipvs services
I0330 22:46:05.082905       1 health_controller.go:183] Health controller tick
I0330 22:46:10.082902       1 health_controller.go:183] Health controller tick
I0330 22:46:10.975521       1 health_controller.go:74] Received heartbeat from NRC
I0330 22:46:15.082910       1 health_controller.go:183] Health controller tick
E0330 22:46:15.082963       1 health_controller.go:125] NetworkService Controller heartbeat missed
I0330 22:46:20.082913       1 health_controller.go:183] Health controller tick
E0330 22:46:20.082969       1 health_controller.go:125] NetworkService Controller heartbeat missed
I0330 22:46:24.011165       1 service_endpoints_sync.go:424] Cleaning up if any, old service IPs on dummy interface
I0330 22:46:24.077035       1 service_endpoints_sync.go:466] Cleaning up if any, old ipvs service and servers which are no longer needed
I0330 22:46:25.082902       1 health_controller.go:183] Health controller tick
E0330 22:46:25.082949       1 health_controller.go:125] NetworkService Controller heartbeat missed
I0330 22:46:28.241897       1 network_services_controller.go:801] Received update to service: ###REMOVED### from watch API
I0330 22:46:30.082917       1 health_controller.go:183] Health controller tick
E0330 22:46:30.082950       1 health_controller.go:125] NetworkService Controller heartbeat missed
I0330 22:46:35.082911       1 health_controller.go:183] Health controller tick
E0330 22:46:35.082961       1 health_controller.go:125] NetworkService Controller heartbeat missed
I0330 22:46:35.397739       1 health_controller.go:161] Shutting down health controller
I0330 22:46:35.397802       1 health_controller.go:178] Shutting down HealthController RunCheck
E0330 22:46:35.397821       1 health_controller.go:150] Health controller error: http: Server closed
I0330 22:46:49.019408       1 service_endpoints_sync.go:405] Setting up policy routing required for Direct Server Return functionality.
I0330 22:46:49.027262       1 service_endpoints_sync.go:410] Custom routing table kube-router-dsr required for Direct Server Return is setup as expected.
I0330 22:46:49.027292       1 service_endpoints_sync.go:412] Setting up custom route table required to add routes for external IP's.
I0330 22:46:49.044220       1 service_endpoints_sync.go:418] Custom routing table external_ip required for Direct Server Return is setup as expected.
I0330 22:46:49.044246       1 service_endpoints_sync.go:77] IPVS servers and services are synced to desired state
I0330 22:46:49.044264       1 service_endpoints_sync.go:28] sync ipvs services took 44.03437442s
I0330 22:46:49.044310       1 network_services_controller.go:359] Shutting down network services controller

```

murali-reddy:
ensure heartbeats are sent when performing sync in both the cases of  periodic sync and on add/delete/update evernts corresponding to services and endpoints

this change just moves sending the hearbeat to method `syncIpvsServices` that is common method that is invoked in all cases.

Fixes #879

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -375,6 +375,7 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
 			}
 
 		case perform := <-nsc.syncChan:
+			healthcheck.SendHeartBeat(healthChan, "NSC")
 			switch perform {
 			case synctypeAll:
 				glog.V(1).Info("Performing requested full sync of services")
@@ -390,6 +391,9 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
 				if err != nil {
 					glog.Errorf("Error during ipvs sync in network service controller. Error: " + err.Error())
 				}
+			}
+			if err == nil {
+				healthcheck.SendHeartBeat(healthChan, "NSC")
 			}
 
 		case <-t.C:


pkg/controllers/proxy/network_services_controller.go
```diff
@@ -394,7 +397,6 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
 
 		case <-t.C:
 			glog.V(1).Info("Performing periodic sync of ipvs services")
-			healthcheck.SendHeartBeat(healthChan, "NSC")
```
aauren, line 397: This approach makes sense from a minimalistic logic point of view, but I think that it would be a mistake to remove the network services controller heartbeat from the network_services_controller file. I think that putting it in the service_endpoints_sync file makes it harder to find and also it seems like it might get lost there.

My recommendation is that we leave it here, and also place another `SendHeartBeat` call just above the `switch perform` statement on line 381. I think that the intent will be more obvious and it won't be victim of a refactor later.

murali-reddy: While I agree that code to send explicit heartbeats makes intent more obvious, I am worried that its error prone as one can easily miss to add if there is any future events that can trigger sync. Just like the case this bug fixes it got missed after refactor.

murali-reddy: Ok, updated the PR as per suggesttion. Effectivley they achive same result, but its just matter of readability, less error prone code.

Plan to raise another PR to better structure health checks and sync's across the controllers.

PR #894, opened 2020-05-04
.goreleaser.yml: Multiarch build

mrueg:
Support other arches as well, so they will appear as tarballs on releases


murali-reddy:
LGTM

PR #897, opened 2020-05-11
Add toleration for node.cloudprovider.kubernetes.io/uninitialized

cfrantsen:
When kubelet is started with --cloud-provider=external the node is
tainted until the cloud-controller-manager has initialized the node.
If the provider responsible for untainting the node doesn't tolerate
node.kubernetes.io/not-ready it will never be deployed as the node
will not be ready before kube-router is running.

### DIFF
--- a/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
+++ b/daemonset/generic-kuberouter-all-features-advertise-routes.yaml
@@ -148,6 +148,9 @@ spec:
       - effect: NoSchedule
         key: node.kubernetes.io/not-ready
         operator: Exists
+      - effect: NoSchedule
+        key: node.cloudprovider.kubernetes.io/uninitialized
+        value: "true"
       volumes:
       - name: lib-modules
         hostPath:
--- a/daemonset/generic-kuberouter-all-features.yaml
+++ b/daemonset/generic-kuberouter-all-features.yaml
@@ -144,6 +144,9 @@ spec:
       - effect: NoSchedule
         key: node.kubernetes.io/not-ready
         operator: Exists
+      - effect: NoSchedule
+        key: node.cloudprovider.kubernetes.io/uninitialized
+        value: "true"
       volumes:
       - name: lib-modules
         hostPath:
--- a/daemonset/generic-kuberouter-only-advertise-routes.yaml
+++ b/daemonset/generic-kuberouter-only-advertise-routes.yaml
@@ -69,6 +69,9 @@ spec:
       - effect: NoSchedule
         key: node.kubernetes.io/not-ready
         operator: Exists
+      - effect: NoSchedule
+        key: node.cloudprovider.kubernetes.io/uninitialized
+        value: "true"
       volumes:
         - name: xtables-lock
           hostPath:
--- a/daemonset/generic-kuberouter.yaml
+++ b/daemonset/generic-kuberouter.yaml
@@ -114,6 +114,9 @@ spec:
       - effect: NoSchedule
         key: node.kubernetes.io/not-ready
         operator: Exists
+      - effect: NoSchedule
+        key: node.cloudprovider.kubernetes.io/uninitialized
+        value: "true"
       volumes:
       - name: lib-modules
         hostPath:
--- a/daemonset/kube-router-all-service-daemonset-advertise-routes.yaml
+++ b/daemonset/kube-router-all-service-daemonset-advertise-routes.yaml
@@ -113,6 +113,9 @@ spec:
       - effect: NoSchedule
         key: node.kubernetes.io/not-ready
         operator: Exists
+      - effect: NoSchedule
+        key: node.cloudprovider.kubernetes.io/uninitialized
+        value: "true"
       volumes:
       - name: lib-modules
         hostPath:
--- a/daemonset/kube-router-all-service-daemonset.yaml
+++ b/daemonset/kube-router-all-service-daemonset.yaml
@@ -109,6 +109,9 @@ spec:
       - effect: NoSchedule
         key: node.kubernetes.io/not-ready
         operator: Exists
+      - effect: NoSchedule
+        key: node.cloudprovider.kubernetes.io/uninitialized
+        value: "true"
       volumes:
       - name: lib-modules
         hostPath:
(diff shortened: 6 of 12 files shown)


murali-reddy:
@cfrantsen Not sure this is required. On what cloud did you run in to this issue.

From what i see CCM's does not have  tolerations for `node.kubernetes.io/not-ready`. For e.g. 

https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/releases/v0.1.24.yml#L23

https://github.com/kubernetes/cloud-provider-aws/blob/master/manifests/aws-cloud-controller-manager-daemonset.yaml#L22

this should have been problem all the while?

@bhcleek / @andrewsykim any thoughts?

andrewsykim:
I don't think we want to tolerate it here because kube-router depends on the node addresses which the external cloud provider is responsible for setting.  The provider should be setup to work before a CNI plugin is deployed, typically they work around this by running on host network.

murali-reddy:
thanks @andrewsykim

cfrantsen:
I encountered this when setting up an on-prem cluster with kubeadm, kube-router and the cloud-provider-vsphere (out-of-tree). It might be a case of doing to much stuff at the same time during bootstrap, I'l try and reproduce the issue with a bare minimum of kubeadm config.

cfrantsen:
I re-ran this whole scenario with a stripped down version of my `kubeadm-config.yaml`:
```
apiVersion: kubeadm.k8s.io/v1beta2
kind: InitConfiguration
nodeRegistration:
  kubeletExtraArgs:
    cloud-provider: external
---
apiVersion: kubeadm.k8s.io/v1beta2
kind: JoinConfiguration
nodeRegistration:
  kubeletExtraArgs:
    cloud-provider: external
---
apiVersion: kubeadm.k8s.io/v1beta2
kind: ClusterConfiguration
kubernetesVersion: v1.18.2
clusterName: dev1.local
controlPlaneEndpoint: api.dev1.local:6443
imageRepository: registry.dev1.local/k8s
networking:
  serviceSubnet: 10.96.0.0/12
  podSubnet: 172.18.0.0/16
---
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
cgroupDriver: systemd
```

After `kubeadm init` the overall status is the following:

`# kubectl get nodes -o wide`
```
NAME      STATUS     ROLES    AGE   VERSION   INTERNAL-IP    EXTERNAL-IP   OS-IMAGE                       KERNEL-VERSION   CONTAINER-RUNTIME
master1   NotReady   master   12s   v1.18.2   10.10.10.173   <none>        Debian GNU/Linux 10 (buster)   4.19.0-8-amd64   containerd://1.3.4
```

`# kubectl describe nodes | grep -A6 Conditions`
```
Conditions:
  Type             Status  LastHeartbeatTime                 LastTransitionTime                Reason                       Message
  ----             ------  -----------------                 ------------------                ------                       -------
  MemoryPressure   False   Wed, 13 May 2020 08:10:14 +0200   Wed, 13 May 2020 08:04:51 +0200   KubeletHasSufficientMemory   kubelet has sufficient memory available
  DiskPressure     False   Wed, 13 May 2020 08:10:14 +0200   Wed, 13 May 2020 08:04:51 +0200   KubeletHasNoDiskPressure     kubelet has no disk pressure
  PIDPressure      False   Wed, 13 May 2020 08:10:14 +0200   Wed, 13 May 2020 08:04:51 +0200   KubeletHasSufficientPID      kubelet has sufficient PID available
  Ready            False   Wed, 13 May 2020 08:10:14 +0200   Wed, 13 May 2020 08:04:51 +0200   KubeletNotReady              runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized
```
  
`# kubectl describe nodes | grep -A3 Taints`
```
Taints:             node-role.kubernetes.io/master:NoSchedule
                    node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule
                    node.kubernetes.io/not-ready:NoSchedule
Unschedulable:      false
```

`# kubectl get pods -n kube-system -o wide`
```
NAME                              READY   STATUS    RESTARTS   AGE   IP             NODE      NOMINATED NODE   READINESS GATES
coredns-cb696dd9-78vhj            0/1     Pending   0          10s   <none>         <none>    <none>           <none>
coredns-cb696dd9-wj5wx            0/1     Pending   0          10s   <none>         <none>    <none>           <none>
etcd-master1                      1/1     Running   0          19s   10.10.10.173   master1   <none>           <none>
kube-apiserver-master1            1/1     Running   0          19s   10.10.10.173   master1   <none>           <none>
kube-controller-manager-master1   1/1     Running   0          19s   10.10.10.173   master1   <none>           <none>
kube-proxy-5jwxd                  1/1     Running   0          11s   10.10.10.173   master1   <none>           <none>
kube-scheduler-master1            1/1     Running   0          19s   10.10.10.173   master1   <none>           <none>
```

As you can see the node is `NotReady` due to CNI not being configured, nothing unexpected there. The node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule taint has been added due to `cloud-provider: external` being used.

- If I try to deploy `cloud-provider-vsphere`, nothing happens as that Daemonset doesn't tolerate node.kubernetes.io/not-ready:NoSchedule.
- If I try to deploy `kube-router`, nothing happens as that Daemonset doesn't tolerate node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule.

I'm thinking the solution to this problem can be either:
- This PR
- Add tolerations for node.kubernetes.io/not-ready:NoSchedule to `cloud-provider-vsphere`
- Some kind of bootstrap workaround where the cluster is initialized without a cloud-provider and then add it after the CNI is installed/configured but that means I have to resort to manually configure kubelet settings that should be handled by kubeadm.

andrewsykim:
@cfrantsen can you share the manifest you're using for the vSphere daemonset?

cfrantsen:
https://github.com/kubernetes/cloud-provider-vsphere/blob/v1.1.0/manifests/controller-manager/vsphere-cloud-controller-manager-ds.yaml

andrewsykim:
Thanks, I think you are correct and this is an issue on Kube >= 1.18 because the `TaintBasedEvictions` was enabled by default there. Before v1.18 the `NotReady` state was only a node condition and running the CCM on host network was enough to bypass it. Can you open an issue on the vSphere cloud provider and assign it to me? It should be addressed there instead.

andrewsykim:
For now you probably want update the vSphere daemonset to tolerate the not-ready taint

cfrantsen:
Closing this as work will be done in the cloud-provider-vsphere project to fix this.

PR #898, opened 2020-05-11
Use same image for container and initContainer

cfrantsen:
Seems kinda pointless to depend on two images when one is enough :-)


murali-reddy:
thanks @cfrantsen LGTM

PR #914, opened 2020-05-27
whitelist traffic to cluster IP and node ports in INPUT chain to bypass netwrok policy enforcement

### LINKED ISSUE #905: Egress network policies should not be applied to pod to service VIP traffic

murali-reddy:
Fixes #905

Following rules will be prepended in `KUBE-ROUTER-INPUT` chain. So traffic to cluster IP's and node port services are targetted to `RETURN`.

```
Chain KUBE-ROUTER-INPUT (1 references)
    0     0 RETURN     tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            /* allow LOCAL traffic to node ports */ ADDRTYPE match dst-type LOCAL multiport dports 30000:32767
    0     0 RETURN     udp  --  *      *       0.0.0.0/0            0.0.0.0/0            /* allow LOCAL traffic to node ports */ ADDRTYPE match dst-type LOCAL multiport dports 30000:32767
   15  1464 RETURN     all  --  *      *       0.0.0.0/0            10.96.0.0/12         /* allow traffic to cluster IP */
    0     0 KUBE-POD-FW-VTNZYDWW3NHVYEUD  all  --  *      *       10.1.1.4             0.0.0.0/0            /* rule to jump traffic from POD name:frontend-56fc5b6b47-dvksf namespace: default to chain KUBE-POD-FW-VTNZYDWW3NHVYEUD */

```

Note: there is no dependency/assumptions on service proxy. Service proxy could add its filtering rules to filter traffic to services. Below is e.g. in case where kube-router is acting as service proxy and has setup filter rules to restrict traffic to service VIP's.

```
# iptables -t filter -L INPUT -n -v
Chain INPUT (policy ACCEPT 1483 packets, 419K bytes)
 265K   75M KUBE-ROUTER-INPUT  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kube-router netpol */
 265K   75M KUBE-ROUTER-SERVICES  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* handle traffic to IPVS service IPs in custom chain */ match-set kube-router-service-ips dst
```

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -54,14 +54,16 @@ const (
 
 // NetworkPolicyController strcut to hold information required by NetworkPolicyController
 type NetworkPolicyController struct {
-	nodeIP              net.IP
-	nodeHostName        string
-	mu                  sync.Mutex
-	syncPeriod          time.Duration
-	MetricsEnabled      bool
-	v1NetworkPolicy     bool
-	healthChan          chan<- *healthcheck.ControllerHeartbeat
-	fullSyncRequestChan chan struct{}
+	nodeIP                net.IP
+	nodeHostName          string
+	serviceClusterIPRange string
+	serviceNodePortRange  string
+	mu                    sync.Mutex
+	syncPeriod            time.Duration
+	MetricsEnabled        bool
+	v1NetworkPolicy       bool
+	healthChan            chan<- *healthcheck.ControllerHeartbeat
+	fullSyncRequestChan   chan struct{}
 
 	ipSetHandler *utils.IPSet
 
@@ -197,6 +199,42 @@ func (npc *NetworkPolicyController) ensureTopLevelChains() {
 		glog.Fatalf("Failed to initialize iptables executor due to %s", err.Error())
 	}
 
+	ensureRuleAtposition := func(chain string, ruleSpec []string, position int) {
+		rules, err := iptablesCmdHandler.List("filter", chain)
+		if err != nil {
+			glog.Fatalf("failed to list rules in filter table %s chain due to %s", chain, err.Error())
+		}
+
+		exists, err := iptablesCmdHandler.Exists("filter", chain, ruleSpec...)
+		if err != nil {
+			glog.Fatalf("Failed to verify rule exists in %s chain due to %s", chain, err.Error())
+		}
+		if !exists {
+			err := iptablesCmdHandler.Insert("filter", chain, 1, ruleSpec...)
+			if err != nil {
+				glog.Fatalf("Failed to run iptables command to insert in %s chain %s", chain, err.Error())
+			}
+		}
+		var ruleNo int
+		for i, rule := range rules {
+			rule = strings.Replace(rule, "\"", "", 2)
+			if strings.Contains(rule, strings.Join(ruleSpec, " ")) {
+				ruleNo = i
+				break
+			}
+		}
+		if ruleNo != position {
+			err = iptablesCmdHandler.Insert("filter", chain, position, ruleSpec...)
+			if err != nil {
+				glog.Fatalf("Failed to run iptables command to insert in %s chain %s", chain, err.Error())
+			}
+			err = iptablesCmdHandler.Delete("filter", chain, strconv.Itoa(ruleNo+1))
+			if err != nil {
+				glog.Fatalf("Failed to delete incorrect rule in %s chain due to %s", chain, err.Error())
+			}
+		}
+	}
+
 	chains := map[string]string{"INPUT": kubeInputChainName, "FORWARD": kubeForwardChainName, "OUTPUT": kubeOutputChainName}
 
 	for builtinChain, customChain := range chains {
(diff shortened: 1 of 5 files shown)


murali-reddy:
This PR has been tested through following test cases:

Four services are used: service A, service B, service C and service D. Service B and C are of type NodePort and service D is running in host network. Network policies are appliced such that pods in service A can only access service B.

- pod A corresponding to service A should be able to access pod B in service B (in both cases where pod B is on the same node and different node) directly
- pod A corresponding to service A should be able to access cluster IP of service B and get load balanced to one of the pods in service B
- pod A corresponding to service A should be able to access NodePort of service B and get load balanced to one of the pods in service B
 - pod A corresponding to service A should be able to access cluster IP of service C but load balanced traffic to one of the pods in service C should be dropped
- pod A corresponding to service A should be able to access NodePort of service C but load balanced traffic to one of the pods in service C should be dropped
- pod A should NOT be able to access the host network service D
- verified following order of chains is ensured on restart of kube-router pods and periodic syncs

```
# iptables -t filter -L INPUT -n -v
Chain INPUT (policy ACCEPT 1483 packets, 419K bytes)
 265K   75M KUBE-ROUTER-INPUT  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kube-router netpol */
 265K   75M KUBE-ROUTER-SERVICES  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* handle traffic to IPVS service IPs in custom chain */ match-set kube-router-service-ips dst
```
- verified firewall rules configured for kube-router IPVS service proxy are honoured

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -197,48 +199,65 @@ func (npc *NetworkPolicyController) ensureTopLevelChains() {
...
-			glog.Fatalf("Failed to run iptables command to create %s chain due to %s", customChain, err.Error())
+	ensureRuleAtposition := func(chain string, ruleSpec []string, position int) {
+		rules, err := iptablesCmdHandler.List("filter", chain)
+		if err != nil {
+			glog.Fatalf("failed to list rules in filter table %s chain due to %s", chain, err.Error())
 		}
-		args := []string{"-m", "comment", "--comment", "kube-router netpol", "-j", customChain}
-		exists, err := iptablesCmdHandler.Exists("filter", builtinChain, args...)
+
+		exists, err := iptablesCmdHandler.Exists("filter", chain, ruleSpec...)
```
aauren, line 208: Really minor nitpic, but we should probably do this existence checking before the `List` call above. If the iptables rule doesn't exist, then there is no reason to do the above list and it would save us an iptables call.

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -197,48 +199,65 @@ func (npc *NetworkPolicyController) ensureTopLevelChains() {
...
-			if ruleNo != 1 {
-				err = iptablesCmdHandler.Insert("filter", builtinChain, 1, args...)
-				if err != nil {
-					glog.Fatalf("Failed to run iptables command to insert in %s chain %s", builtinChain, err.Error())
-				}
-				err = iptablesCmdHandler.Delete("filter", builtinChain, strconv.Itoa(ruleNo+1))
-				if err != nil {
-					glog.Fatalf("Failed to delete wrong rule to jump to chain %s in %s chain due to %s", customChain, builtinChain, err.Error())
-				}
+			err = iptablesCmdHandler.Delete("filter", chain, strconv.Itoa(ruleNo+1))
```
aauren, line 232: I know that you've talked about this other places in the code, but at some point in the future we should really not rely on deleting iptables rules by number (I know we do it all over the place). But there exists a possibility that since we've listed the rules, that some other process has come in and manipulated iptables causing us to potentially delete the wrong rule.

pkg/options/options.go
```diff
@@ -74,6 +76,8 @@ func NewKubeRouterConfig() *KubeRouterConfig {
 		BGPGracefulRestartDeferralTime: 360 * time.Second,
 		EnableOverlay:                  true,
 		OverlayType:                    "subnet",
+		ClusterIPCIDR:                  "10.96.0.0/12",
```
aauren, line 79: Is this a safe guess for all of the user's that use our daemonset's that we publish?

aauren:
I know my review comes late. Overall, LGTM.

For posterity, as we discussed offline earlier, this will affect the ability to route to ExternalIPs or LoadBalancer type services from within the cluster. Granted, I don't think this was the intention of Kubernetes for inter-cluster traffic (i.e. it should be using a ClusterIP), but it does have the potential to affect users (I know that our current users use ExternalIPs for inter-cluster traffic routing because that's what they get back from Consul).

Other than that, I had one nit and one question, but neither are pressing.

pkg/controllers/netpol/network_policy_controller.go
murali-reddy: Agree. Will fix it.

pkg/controllers/netpol/network_policy_controller.go
murali-reddy: Yes. There is a possibility. Perhaps holding the iptables lock followed by iptables-save and iptables-restore would be a safe solution. But if there are non-cooperating processes which does not use locks we just have no way to control it with iptables. Such races has been problem:

https://github.com/weaveworks/weave/issues/3230
https://github.com/kubernetes/kubernetes/issues/20391

pkg/options/options.go
murali-reddy: Its default value kubernetes uses.

murali-reddy:
@aauren thanks for the review

> For posterity, as we discussed offline earlier, this will affect the ability to route to ExternalIPs or LoadBalancer type services from within the cluster. Granted, I don't think this was the intention of Kubernetes for inter-cluster traffic (i.e. it should be using a ClusterIP), but it does have the potential to affect users (I know that our current users use ExternalIPs for inter-cluster traffic routing because that's what they get back from Consul).

Please open an issue for the use-case. We should discuss and see how to evolve it.

aauren:
> @aauren thanks for the review
> 
> > For posterity, as we discussed offline earlier, this will affect the ability to route to ExternalIPs or LoadBalancer type services from within the cluster. Granted, I don't think this was the intention of Kubernetes for inter-cluster traffic (i.e. it should be using a ClusterIP), but it does have the potential to affect users (I know that our current users use ExternalIPs for inter-cluster traffic routing because that's what they get back from Consul).
> 
> Please open an issue for the use-case. We should discuss and see how to evolve it.

Added here: https://github.com/cloudnativelabs/kube-router/issues/934

PR #915, opened 2020-05-27
run through all applicable network policies for a flow

### LINKED ISSUE #916: network policies not fully enforced in bridged traffic
For a flow where source and destination pods are on the same node kube-router need to run through network policies for both the pods. For e.g. for the pod A-> pod B traffic egress network policies corresponding to pod A to be applied first and followed by ingress network policies for the pod B.

If the pod A's egress network policy permits the traffic but pod B's ingress network policy deny the traffic then packet must be dropped. In current implementation once pod A network policies are matched traffic is applied ACCEPT in iptables so traffic is never run through the other chains enforcing other network policies.

kube-router should run through all network policies and should deny/allow traffic.

murali-reddy:
- change ACCEPT to RETURN with mark when a netpol is matched so that we run through
all the applicable network policies for a flow

Fixes #916

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -756,11 +756,21 @@ func (npc *NetworkPolicyController) appendRuleToPolicyChain(iptablesCmdHandler *
 	if dPort != "" {
 		args = append(args, "--dport", dPort)
 	}
-	args = append(args, "-j", "ACCEPT")
-	err := iptablesCmdHandler.AppendUnique("filter", policyChainName, args...)
+
+	markComment := "rule to mark traffic matching a network policy"
+	markArgs := append(args, "-j", "MARK", "-m", "comment", "--comment", markComment, "--set-xmark", "0x10000/0x10000")
+	err := iptablesCmdHandler.AppendUnique("filter", policyChainName, markArgs...)
 	if err != nil {
 		return fmt.Errorf("Failed to run iptables command: %s", err.Error())
 	}
+
+	returnComment := "rule to RETURN traffic matching a network policy"
+	returnArgs := append(args, "-m", "comment", "--comment", returnComment, "-m", "mark", "--mark", "0x10000/0x10000", "-j", "RETURN")
+	err = iptablesCmdHandler.AppendUnique("filter", policyChainName, returnArgs...)
+	if err != nil {
+		return fmt.Errorf("Failed to run iptables command: %s", err.Error())
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


murali-reddy:
This PR fixes specific case where both source pod and destination pod are on the same node. An egress network policy applied on the source pod may permit traffic from the source pod to destination pod, but a seperate network policy applied to destination pod may reject it. So for the flow both the network policies to be applied to permit/deny traffic. Apart from verifying the fix address the above case ensure there are no regression.

Merging the PR based on testing.

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -756,11 +756,21 @@ func (npc *NetworkPolicyController) appendRuleToPolicyChain(iptablesCmdHandler *
 	if dPort != "" {
 		args = append(args, "--dport", dPort)
 	}
-	args = append(args, "-j", "ACCEPT")
-	err := iptablesCmdHandler.AppendUnique("filter", policyChainName, args...)
+
+	markComment := "rule to mark traffic matching a network policy"
+	markArgs := append(args, "-j", "MARK", "-m", "comment", "--comment", markComment, "--set-xmark", "0x10000/0x10000")
```
aauren, line 761: I take it that you chose this value, because it's above this range: https://github.com/cloudnativelabs/kube-router/blob/4d6b0b818fc6209142c7ca91b8774d4573de1349/pkg/controllers/proxy/network_services_controller.go#L1676 ? Is there anything else in kube-router than deals with marks?

aauren:
Seems reasonable, and I can't think of a better approach, but it's a bummer that it essentially doubles the amount of iptables rules that we're going to have though. Overall LGTM.

pkg/controllers/netpol/network_policy_controller.go
murali-reddy: No other use of Markings. Just choose random numbers.

murali-reddy:
> Seems reasonable, and I can't think of a better approach, but it's a bummer that it essentially doubles the amount of iptables rules that we're going to have though. Overall LGTM.

Thanks for the review. Actually its not bad. Or atleast it does not make any worse performance wise.

Since kube-router is stateful firewall, for the established sessions (i.e. "RELATED,ESTABLISHED") we still `ACCEPT`, its just for the first packet its run through all the chains.

aauren:
Sorry I should have been more clear. I was referring to the time that it takes to perform the full sync on iptables, not the performance of the data path. You're correct, this doesn't have a huge impact on the data path.

PR #917, opened 2020-05-29
network_polixy: full sync when namespace labels change

liuxu623:
we should sync firewall if namespace labels change, beacause some network policy maybe use namespace selector, and I move namespace code to namspace.go.

### DIFF
--- a/pkg/controllers/netpol/namespace.go
+++ b/pkg/controllers/netpol/namespace.go
@@ -1,0 +1,60 @@
+package netpol
+
+import (
+	"github.com/golang/glog"
+	api "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"reflect"
+)
+
+func (npc *NetworkPolicyController) newNamespaceEventHandler() cache.ResourceEventHandler {
+	return cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			npc.handleNamespaceAdd(obj.(*api.Namespace))
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			npc.handleNamespaceUpdate(oldObj.(*api.Namespace), newObj.(*api.Namespace))
+		},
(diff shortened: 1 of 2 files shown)


pkg/controllers/netpol/namespace.go
```diff
@@ -0,0 +1,69 @@
...
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			npc.handleNamespaceUpdate(oldObj.(*api.Namespace), newObj.(*api.Namespace))
+		},
+		DeleteFunc: func(obj interface{}) {
+			switch obj := obj.(type) {
+			case *api.Namespace:
+				npc.handleNamespaceDelete(obj)
+				return
+			case cache.DeletedFinalStateUnknown:
+				npc.handleNamespaceDelete(obj.Obj.(*api.Namespace))
```
murali-reddy, line 23: We still need to check if obj.Obj can be type casted to `*api.Namespace`

pkg/controllers/netpol/namespace.go
```diff
@@ -0,0 +1,69 @@
...
+func (npc *NetworkPolicyController) handleNamespaceDelete(obj *api.Namespace) {
+	if npc.v1NetworkPolicy && obj.Labels == nil {
+		return
+	}
+	glog.V(2).Infof("Received namespace: %s delete event", obj.Name)
+
+	npc.RequestFullSync()
+}
+
+func equals(a, b map[string]string) bool {
```
murali-reddy, line 59: `reflect.DeepEqual` should work here

pkg/controllers/netpol/namespace.go
liuxu623: done

pkg/controllers/netpol/namespace.go
liuxu623: ok

murali-reddy:
thanks @liuxu623 LGTM

PR #920, opened 2020-06-04
Add the advertise-service-cluster-ip-range parameter to summarize the announcement function of the service network segment, reduce the routing entries of the connected network devices, and support a larger BGP network.

cloudnativer:
Each kubernetes cluster in our production environment has 4000 nodes, and the whole network is interconnected by BGP, which has been running stably for more than two year. There are many problems with kube-router in the large kubernetes cluster, and we have done a lot of optimization, so I want to contribute some information to the community.I have contributed an enhanced function in the large kubernetes cluster network to Kube router, as well as several practical documents about the large kubernetes cluster network.

1. Added the "advertise-cluster-ip-range" flag parameter for optimizing the number of routes.
I added the "advertise-cluster-ip-range" flag parameter to kube-router. When you set the parameters of "-advertise-cluster-IP=true" and "-advertise-cluster-ip-range=you_service_ip_range" at the same time, the kubernetes node will only announce the aggregate cluster route you specified to the on-line router device.

2. Documents for optimization of large kubernetes cluster network are compiled. Please check https://github.com/cloudnativelabs/kube-router/pull/944 for details.
In order for your architecture to support a larger network, you need to do the following two things: 
    (3) You need to set both "--advertise-cluster-IP=true" and "--advertise-cluster-ip-range=subnet" parameters.Let k8s node only notify k8s service aggregate routes to the upstream routers, reducing the service routing entries of the upstream routers.(See large-networks03 documentation).


murali-reddy:
> Each kubernetes cluster in our production environment has 4000 nodes, and the whole network is interconnected by BGP, which has been running stably for more than two year.

@cloudnativer This has to be largest reported cluster of kube-router. Good to know kube-router is able to scale well.

> There are many problems with kube-router in the large kubernetes cluster, and we have done a lot of optimization, so I want to contribute some information to the community.

thanks for contributing back. 

> Added the "advertise-cluster-subnet" flag parameter for optimizing the number of routes.
I added the "advertise-cluster-subnet" flag parameter to kube-router. When you set the parameters of "-advertise-cluster-IP=true" and "-advertise-cluster-subnet=subnet" at the same time, 

I have not fully looked at the PR. I presume you are referring to `service-cluster-ip-range`. Can we please use the name same as the one used by kube-api-server, kubeadm etc for e.g. `-advertise-service-cluster-ip-range`?

> the kubernetes node will only announce the aggregate cluster route you specified to the on-line router device.

Let me clarify a concern here. This is specific to kube-router. kube-router will advertise a service VIP only if it has a backend pod for the service running on the node when annotated with `kube-router.io/service.local` or `externalTrafficPolicy=local` on the service object. So if `-advertise-cluster-subnet=subnet` is set then each node will advertise entire range. So this would break above funtionality right?

Can you please close #919 if this PR is an update?

cloudnativer:
This PR, I have transferred to https://github.com/cloudnativelabs/kube-router/pull/920 , and associated with https://github.com/cloudnativelabs/kube-router/issues/923 .
We want to route the kubernets service IP directly to the external network for external access to the services in the kubernets cluster. We set "advertisement cluster IP = true". Kube router will announce the 32-bit host route of the service to the connected network device.
If we have 10000 service IPS, Kube router will announce the 32-bit host routes of these 10000 services to the connected network devices. With the expansion of kubernetes cluster network, what if we have 100000 service IPS?
In fact, these service IPS are all in the "service cluster IP range", so we can only declare a summarized service subnet to the upper network devices. 




If we are in a large-scale kubernetes network, we may also need to turn on the BGP ECMP function of the upper network device to realize the load balance of network traffic. When BGP ECMP is turned on, the kubernetes service route received by the connected network device will be more. If there are 40 kubernetes nodes under an uplink network device, and we have 100000 services, then you will see 4000000 ECMP routes on the uplink network device.

The "advertisement cluster subnet" function I added to the kube-router allows you to customize the kubernetes service segment route announced to the upper network device. We can aggregate 100000 service routes into one route and announce it. In this way, the number of routes received on the connected network equipment will be reduced from 4000000 to 40, greatly saving the routing table and resources of the network device.


This looks like the advance of ADVERTISE_CLUSTER_IPS function, please see https://www.projectcalico.org/kubernetes-service-ip-route-advertisement/.


If you set the parameter "-advertise-cluster-subnet", you can announce the kubernetes service route in large-scale network and enable ECMP. Because "-advertise-cluster-subnet" does the routing summary, reduces the number of service IP routing entries of the connected network equipment, and reduces the performance loss of the network device.
By the way, if you think it is more reasonable to use "-advertise-service-cluster-ip-range", of course, you can change "advertise-cluster-subnet" parameter in PR to "-advertise-service-cluster-ip-range" parameter.







--------------------------------------------------------------------------------





------------------ 原始邮件 ------------------




 
 发件人: Murali Reddy  <notifications@github.com>
 

 
 发送时间: 2020-06-04 13:16:35
 

 
 收件人:cloudnativelabs/kube-router <kube-router@noreply.github.com> 

 
 抄送:house <houggonline@139.com>,Mention <mention@noreply.github.com> 

 
 主题: Re: [cloudnativelabs/kube-router] [update]Add theadvertise-cluster-subnet parameter to summarize the announcement function ofthe service network segment, reduce the routing entries of the connectednetwork devices, and support a larger BGP network. (#920)
 





 Each kubernetes cluster in our production environment has 4000 nodes, and the whole network is interconnected by BGP, which has been running stably for more than two year.

 @cloudnativer This has to be largest reported cluster of kube-router. Good to know kube-router is able to scale well.

 There are many problems with kube-router in the large kubernetes cluster, and we have done a lot of optimization, so I want to contribute some information to the community.

 thanks for contributing back.

 Added the "advertise-cluster-subnet" flag parameter for optimizing the number of routes.
 I added the "advertise-cluster-subnet" flag parameter to kube-router. When you set the parameters of "-advertise-cluster-IP=true" and "-advertise-cluster-subnet=subnet" at the same time,

 I have not fully looked at the PR. I presume you are referring to service-cluster-ip-range. Can we please use the name same as the one used by kube-api-server, kubeadm etc for e.g. -advertise-service-cluster-ip-range?

 the kubernetes node will only announce the aggregate cluster route you specified to the on-line router device.

 Let me clarify a concern here. This is specific to kube-router. kube-router will advertise a service VIP only if it has a backend pod for the service running on the node when annotated with kube-router.io/service.local or externalTrafficPolicy=local on the service object. So if -advertise-cluster-subnet=subnet is set then each node will advertise entire range. So this would break above funtionality right?

 Can you please close #919 if this PR is an update?

cloudnativer:
@murali-reddy
>Can we please use the name same as the one used by kube-api-server, kubeadm etc for e.g. -advertise-service-cluster-ip-range?


Now I've changed "advertise-cluster-subnet" to "advertise-service-cluster-ip-range".
Keep the same parameter names as kube-api-server, kubeadm etc.

cloudnativer:
>So if -advertise-cluster-subnet=subnet is set then each node will advertise entire range. So this would break above funtionality right?

The purpose of setting this parameter is to reduce the number of BGP route entries declared to the connected network devices when the service subnet needs to be exposed to the outside in a large-scale k8s cluster network. Please see https://github.com/cloudnativelabs/kube-router/issues/923.

murali-reddy:
@cloudnativer I feel this use-case is relevent. Given that L3 core routers routing table can quickly filled as cluster size increases. But we need to see how it works with services with `externalTrafficPolicy=local` set. I would imagine if the nodes continue to advertise `/32` VIP for such service along with `service-cluster-ip-range`, a `/32` route would take precendence and everything should fall in place. But we need to verify. Will work on this PR next week.

Meanwhile may i suggest to split the PR? Move all the documentation to sepearate PR. It would be far easier to focus and review?

cloudnativer:
> @cloudnativer I feel this use-case is relevent. Given that L3 core routers routing table can quickly filled as cluster size increases. But we need to see how it works with services with `externalTrafficPolicy=local` set. I would imagine if the nodes continue to advertise `/32` VIP for such service along with `service-cluster-ip-range`, a `/32` route would take precendence and everything should fall in place. But we need to verify. Will work on this PR next week.

 I will continue to pay attention.

> Meanwhile may i suggest to split the PR? Move all the documentation to sepearate PR. It would be far easier to focus and review?

OK, of course.

murali-reddy:
@cloudnativer Hi, can you please split the code changes and documentation in to seperate PR's? 1.0 release is out now. So i want to work on the PR by testing it. thanks for your patience.

cloudnativer:
> @cloudnativer Hi, can you please split the code changes and documentation in to seperate PR's? 1.0 release is out now. So i want to work on the PR by testing it. thanks for your patience.


Yes, of course.

As you requested, I have separated the code changes from the documentation.
- Only code changes are preserved in https://github.com/cloudnativelabs/kube-router/pull/920 .
- Only document changes are retained in https://github.com/cloudnativelabs/kube-router/pull/944 .

murali-reddy:
@cloudnativer thanks for seperating the code changes and documentation in to seperate PR's. 

Just so you know I am testing the patch with `--advertise-service-cluster-ip-range` enabled and services with externalTrafficPolicy=Local.I should get back to you in a day.

pkg/controllers/routing/bgp_policies.go
```diff
@@ -45,9 +45,17 @@ func (nrc *NetworkRoutingController) AddPolicies() error {
 	// creates prefix set to represent all the advertisable IP associated with the services
 	advIPPrefixList := make([]config.Prefix, 0)
 	advIps, _, _ := nrc.getAllVIPs()
-	for _, ip := range advIps {
-		advIPPrefixList = append(advIPPrefixList, config.Prefix{IpPrefix: ip + "/32"})
+
+	//If the value of advertise-service-cluster-ip-range parameter is not empty, then the value of advertise-service-cluster-ip-range parameter is put into RIB, otherwise it will be done according to the original rules.
```
murali-reddy, line 49: I would like both service CIDR and service VIP to be announced **only** for the services marked with external traffic policy set to local. So that on upstream routers with give /32 router precedence.

docs/user-guide.md
```diff
@@ -40,6 +40,7 @@ Usage of kube-router:
       --advertise-external-ip                         Add External IP of service to the RIB so that it gets advertised to the BGP peers.
       --advertise-loadbalancer-ip                     Add LoadbBalancer IP of service status as set by the LB provider to the RIB so that it gets advertised to the BGP peers.
       --advertise-pod-cidr                            Add Node's POD cidr to the RIB so that it gets advertised to the BGP peers. (default true)
+      --advertise-service-cluster-ip-range string     If this parameter is set, Kube-router will add the service cluster IP range set by this parameter to the RIB, and send the routing advertisement of the service cluster IP range to the BGP peer. The purpose of this parameter is to reduce the number of service route entries sent by the Kube-router to the uplink network device. (Please configure "advertise-cluster-ip=true" at the same time to ensure that the "advertise-service-cluster-ip-range" parameter takes effect.)
```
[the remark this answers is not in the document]
cloudnativer: Yes, of course.
I've Cut down the length of the description to a single line, and Add longer description of the flag to the "Advertising IPs" section of the "user-guide.md" document.

pkg/controllers/routing/bgp_policies.go
murali-reddy: @cloudnativer Did you give any thought to above comment? I see below three scenarios

- if `--advertise-service-cluster-ip-range` is configured then advertise **ONLY** the service cluster IP range from the nodes and **DO NOT** advertise service VIP's from the node (which is what this PR intends to achieve) irrespective of fact service has pod's running on the node and service is marked `externalTrafficPolicy=Local`.
- if `--advertise-service-cluster-ip-range` is configured then advertise the service cluster IP range from the nodes **AND** advertise `/32` service VIP's from the node if node has a endpoint pod corresponding to the service running on the node and service is marked `externalTrafficPolicy=Local`.
- if `--advertise-service-cluster-ip-range` is **NOT** configured then keep the original behaviour i.e.) advertise `/32` VIP from all the nodes if service is not marked with `externalTrafficPolicy=Local` and advertise service VIP **ONLY** from the nodes which has a endpoint pod corresponding to the service is running on the node and service is marked with `externalTrafficPolicy=Local` 

problem with #1 is it can blackhole the traffic for services set to `externalTrafficPolicy=Local`. Traffic to service VIP get's ECMP ed to a node not running any service endpoint pod and proxy running on the node will reject the traffic.

cloudnativer: > @cloudnativer Did you give any thought to above comment? I see below three scenarios
> 
> * if `--advertise-service-cluster-ip-range` is configured then advertise **ONLY** the service cluster IP range from the nodes and **DO NOT** advertise service VIP's from the node (which is what this PR intends to achieve) irrespective of fact service has pod's running on the node and service is marked `externalTrafficPolicy=Local`.
> * if `--advertise-service-cluster-ip-range` is configured then advertise the service cluster IP range from the nodes **AND** advertise `/32` service VIP's from the node if node has a endpoint pod corresponding to the service running on the node and service is marked `externalTrafficPolicy=Local`.
> * if `--advertise-service-cluster-ip-range` is **NOT** configured then keep the original behaviour i.e.) advertise `/32` VIP from all the nodes if service is not marked with `externalTrafficPolicy=Local` and advertise service VIP **ONLY** from the nodes which has a endpoint pod corresponding to the service is running on the node and service is marked with `externalTrafficPolicy=Local`
> 
> problem with #1 is it can blackhole the traffic for services set to `externalTrafficPolicy=Local`. Traffic to service VIP get's ECMP ed to a node not running any service endpoint pod and proxy running on the node will reject the traffic.


This problem does exist if ·externalTrafficPolicy=Local· is set.

(1) The applicable conditions of `advertise-service-cluster-ip-range` are as follows:
 * `externalTrafficPolicy ` can only be used when the service is set to loadbalancer or nodeport.
 * If the service is set to loadbalancer or nodeport, you should set `externalTrafficPolicy=Cluster` at the same time, so that the `advertise-service-cluster-ip-range` parameter is meaningful.
 * Except for this situation, you can directly set up `advertise-service-cluster-ip-range`.

(2)In large-scale environment, network traffic load balancing is very important. However, the configuration of `externalTrafficPolicy=Local` will lead to unbalanced network traffic load, so our production environment uses the default `externalTrafficPolicy=Cluster`.

cloudnativer:
@murali-reddy  I submitted the PR code successfully before. Now it seems that there is a conflict? 
Can you resolve this code conflict?

aauren:
Moved to 1.2 as these PRs are large items and we're focusing on library upgrades for 1.1

cloudnativer:
> Moved to 1.2 as these PRs are large items and we're focusing on library upgrades for 1.1

OK, I always pay attention to information.

cloudnativer:
I solved the above code conflict problem and put the latest code in #1049 .
Because #920's code is based on kube-router v0.3, and there is a big conflict between the new version of kube-router v1.1.1 and #920's code. Based on the new version of kube-router v1.1.1, I rewrote the advertise-service-cluster-ip-range function and put it in #1049.

cloudnativer:
> Moved to 1.2 as these PRs are large items and we're focusing on library upgrades for 1.1

Please use the code in #1050 when merging to 1.2.

aauren:
So does that mean that this PR should be closed?

aauren:
Closing in favor of #1050 after no specific feedback from user.

PR #921, opened 2020-06-04
The bgp-holdtime function parameter of setting holdtime is added to adjust the holdtime of BGP negotiation with the connected network devices.

cloudnativer:
In some scenarios, the holdtime of BGP needs to be adjusted.
(1) If you want the BGP neighbor route to disappear as soon as possible after the kube-router goes down, you should set the holdtime to be smaller.
(2) In some scenarios, after the kube-router goes down, the pod network is normal and can still communicate with the outside world. You want BGP neighbor routing to last longer, so in this case you can set the holdtime to be larger.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -42,7 +42,7 @@ Usage of kube-router:
       --advertise-pod-cidr                            Add Node's POD cidr to the RIB so that it gets advertised to the BGP peers. (default true)
       --bgp-graceful-restart                          Enables the BGP Graceful Restart capability so that routes are preserved on unexpected restarts
       --bgp-graceful-restart-deferral-time duration   BGP Graceful restart deferral time according to RFC4724 4.1, maximum 18h. (default 6m0s)
-      --bgp-holdtime float                            This parameter is mainly used to modify the holdtime declared to BGP peer. When Kube router goes down abnormally, the local saving time of BGP route will be affected. (default 90)
+      --bgp-holdtime float                            This parameter is mainly used to modify the holdtime declared to BGP peer. When Kube router goes down abnormally, the local saving time of BGP route will be affected.Holdtime must be in the range 3 to 65536. (default 90)
       --bgp-port uint16                               The port open for incoming BGP connections and to use for connecting with other BGP peers. (default 179)
       --cache-sync-timeout duration                   The timeout for cache synchronization (e.g. '5s', '1m'). Must be greater than 0. (default 1m0s)
       --cleanup-config                                Cleanup iptables rules, ipvs, ipset configuration and exit.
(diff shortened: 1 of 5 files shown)


aauren:
I would recommend that you also put in some value validation as well. According to the IETF specification if a value is non-zero it must be at least 3 seconds and, since it is a 2-octet unsigned integer it's maximum value is 65536

cloudnativer:
> I would recommend that you also put in some value validation as well. According to the IETF specification if a value is non-zero it must be at least 3 seconds and, since it is a 2-octet unsigned integer it's maximum value is 65536


Thank you for your reminding. I've added scope checking to holdtime code.


And I also wrote the range constraint prompt of holdtime after the flag.


I just updated the holdtime scope constraint code to meet the IETF specification.

pkg/options/options.go
```diff
@@ -140,6 +142,8 @@ func (s *KubeRouterConfig) AddFlags(fs *pflag.FlagSet) {
 		"Enables the BGP Graceful Restart capability so that routes are preserved on unexpected restarts")
 	fs.DurationVar(&s.BGPGracefulRestartDeferralTime, "bgp-graceful-restart-deferral-time", s.BGPGracefulRestartDeferralTime,
 		"BGP Graceful restart deferral time according to RFC4724 4.1, maximum 18h.")
+	fs.Float64Var(&s.BGPHoldtime, "bgp-holdtime", DEFAULT_HOLDTIME,
```
mrueg, line 145: can you make this DEFAULT_BGP_HOLDTIME instead of DEFAULT_HOLDTIME (so it matches the other env vars)?

cloudnativer: Yes, of course. I have now set DEFAULT_HOLDTIME changed to DEFAULT_BGP_HOLDTIME。

Here are the relevant screenshots:

docs/user-guide.md
```diff
@@ -43,6 +43,7 @@ Usage of kube-router:
       --bgp-graceful-restart                          Enables the BGP Graceful Restart capability so that routes are preserved on unexpected restarts
       --bgp-graceful-restart-deferral-time duration   BGP Graceful restart deferral time according to RFC4724 4.1, maximum 18h. (default 6m0s)
       --bgp-graceful-restart-time duration            BGP Graceful restart time according to RFC4724 3, maximum 4095s. (default 1m30s)
+      --bgp-holdtime float                            This parameter is mainly used to modify the holdtime declared to BGP peer. When Kube-router goes down abnormally, the local saving time of BGP route will be affected.Holdtime must be in the range 3 to 65536. (default 90)
```
aauren, line 46: @cloudnativer I'm ready to merge this, I just realized something really quickly. Every other unit of time in kube-router is specified via duration which is more idiomatic for Golang. Would you mind converting this to be a unit of duration like the other options?

Very sorry I missed this on my first review pass, but everything else looks good to go, even had a chance to test it in a live cluster, so once this change is made I'll merge.

cloudnativer: ok.
I have been busy with my work recently, so I didn't reply in time. 
I will solve this problem in three days.

cloudnativer: @aauren 
I have changed the type of BGPHoldtime to duration. Please check it again.

cloudnativer:
In order to keep and other time flag parameters always,I have changed the type of BGPHoldtime to duration.

aauren:
Thanks for fixing that up. LGTM.

cloudnativer:
A pleasant cooperation! 
Next time, I will still contribute code to Cloudnativelabs.

PR #929, opened 2020-06-14
Makefile: remove slashes from git branch if they exist

aauren:
@murali-reddy @mrueg 

Git allows slashes in branch names, but they are not valid in docker tags. This trims slashes from the git branch name so that `make` doesn't fail if they exist.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ container: Dockerfile.$(GOARCH).run kube-router gobgp multiarch-binverify ## Bui
 	    echo "Using qemu to build non-native container"; \
 	    $(DOCKER) run --rm --privileged $(QEMU_IMAGE) --reset -p yes; \
 	fi
-	$(DOCKER) build -t "$(REGISTRY_DEV):$(IMG_TAG)" -f Dockerfile.$(GOARCH).run .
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -6,7 +6,7 @@ BUILD_DATE?=$(shell date +%Y-%m-%dT%H:%M:%S%z)
 LOCAL_PACKAGES?=app app/controllers app/options app/watchers utils
 IMG_NAMESPACE?=cloudnativelabs
 GIT_COMMIT=$(shell git describe --tags --dirty)
-GIT_BRANCH?=$(shell git rev-parse --abbrev-ref HEAD)
+GIT_BRANCH?=$(shell git rev-parse --abbrev-ref HEAD | tr -d '/')
```
mrueg, line 9: another option would be to use Makefile substitution to filter `/` characters

```
$(subst from,to,text)
```

This might be better since someone might set GIT_BRANCH externally.

aauren: Updated per recommendation. Also made it specific to the building the docker image which is the only part that has issues with slashes in case someone needed to use that to refer to the actual git branch.

PR #942, opened 2020-06-23
proxy: only output Error log when there's an error

### LINKED ISSUE #941: kube-router v1.0.0-rc6 crashlooping
kube-router v1.0.0-rc6 crash loops due to an attempt to call a the `Error` method on a nil error at https://github.com/cloudnativelabs/kube-router/blob/v1.0.0-rc6/pkg/controllers/proxy/network_services_controller.go#L448.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xf79d13]

goroutine 247 [running]:
github.com/cloudnativelabs/kube-router/pkg/controllers/proxy.(*NetworkServicesController).doSync(0xc420daa000, 0x0, 0x0)
        /go/src/github.com/cloudnativelabs/kube-router/pkg/controllers/proxy/network_services_controller.go:448 +0x193
github.com/cloudnativelabs/kube-router/pkg/controllers/proxy.(*NetworkServicesController).Run(0xc420daa000, 0xc4203dcba0, 0xc4200e83c0, 0xc4204f92c0)
        /go/src/github.com/cloudnativelabs/kube-router/pkg/controllers/proxy/network_services_controller.go:358 +0x419
created by github.com/cloudnativelabs/kube-router/pkg/cmd.(*KubeRouter).Run
        /go/src/github.com/cloudnativelabs/kube-router/pkg/cmd/kube-router.go:164 +0xcee
```

bhcleek:
Fixes #941

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -445,8 +445,10 @@ func (nsc *NetworkServicesController) doSync() error {
 
 	if nsc.MetricsEnabled {
 		err = nsc.publishMetrics(nsc.serviceMap)
-		glog.Errorf("Error publishing metrics: %s", err.Error())
-		return err
+		if err != nil {
+			glog.Errorf("Error publishing metrics: %v", err)
+			return err
+		}
 	}
 	return nil
 }


pkg/controllers/proxy/network_services_controller.go
```diff
@@ -445,7 +445,9 @@ func (nsc *NetworkServicesController) doSync() error {
 
 	if nsc.MetricsEnabled {
 		err = nsc.publishMetrics(nsc.serviceMap)
-		glog.Errorf("Error publishing metrics: %s", err.Error())
+		if err != nil {
+			glog.Errorf("Error publishing metrics: %v", err)
```
aauren, line 449: Following the style of the above return, I would pull the return statement into the `if` expression as well. I get that it doesn't make a practical difference now, but it will keep us from getting caught later if we add more to the `doSync()` method.

bhcleek:
@aauren PTAL

Since it was such a small change I amended the commit  instead of adding a new one.

murali-reddy:
Wonder if there is any static analyzer that can catch such errors. `go vet` is not complaining.

PR #944, opened 2020-06-30
Solve the routing optimization problem in large k8s cluster and support larger BGP routing network scale.

cloudnativer:
Each kubernetes cluster in our production environment has 4000 nodes, and the whole network is interconnected by BGP, which has been running stably for more than two year. There are many problems with kube-router in the large kubernetes cluster, and we have done a lot of optimization, so I want to contribute some information to the community.I have contributed an enhanced function in the large kubernetes cluster network to Kube router, as well as several practical documents about the large kubernetes cluster network.

lanet111

1. Added the "advertise-cluster-ip-range" flag parameter for optimizing the number of routes. Please check https://github.com/cloudnativelabs/kube-router/pull/920 for details.
I added the "advertise-cluster-ip-range" flag parameter to kube-router. When you set the parameters of "-advertise-cluster-IP=true" and "-advertise-cluster-ip-range=you_service_ip_range" at the same time, the kubernetes node will only announce the aggregate cluster route you specified to the on-line router device.

2. Documents for optimization of large kubernetes cluster network are compiled.
In a large-scale kubernetes network, you will encounter problems like https://github.com/cloudnativelabs/kube-router/issues/923.In order for your architecture to support a larger network, you need to do the following two things: 
    (3) You need to set both "--advertise-cluster-IP=true" and "--advertise-cluster-ip-range=subnet" parameters.Let k8s node only notify k8s service aggregate routes to the upstream routers, reducing the service routing entries of the upstream routers.(See large-networks03 documentation). 

lanet222


cloudnativer:
@murali-reddy 

According to the requirements in https://github.com/cloudnativelabs/kube-router/pull/920, Separate the code changes from the document, leaving only the document changes.

cloudnativer:
According to the requirements of murali-Reddy, I have separated the code changes from the documentation.
Only code changes are retained in #920 .
Only document changes are retained in #944 .

salanki:
Are you running firewall in your 4,000 node clusters? We run (only) 500 nodes and are having performance issues with firewall.

Also, how much memory is your kube-router using on the nodes? Without firewall ours are close to 1GB, which is quite a lot.

cloudnativer:
> Are you running firewall in your 4,000 node clusters? We run (only) 500 nodes and are having performance issues with firewall.

We did not enable the firewall function of Kube-router. We turned off the firewall with the `--run-firewall=false` parameter.


> Also, how much memory is your kube-router using on the nodes? Without firewall ours are close to 1GB, which is quite a lot.

If the memory configuration is too small, the Kube-router will crash.
We have 8GB of memory.

murali-reddy:
@salanki @cloudnativer kube-router is completley statless by design. Ideally it needs very less CPU and memory footprint unless there is any leaks. We did fix some of the memory leaks. Please try v1.0.0 release and see if you run into any issues.

salanki:
@murali-reddy: Here is my memory usage over a week of one kube-router 1.0.0 instance on a 500 node cluster. This is router ONLY, no firewall or service proxy. 800MB average memory used. If you think this is abnormal, I'll open a ticket, I personally think it's way to much for a stateless service.

@cloudnativer what is your average memory usage per instance on your 4,000 node cluster?

Screen Shot 2020-07-23 at 10 57 18

murali-reddy:
> If you think this is abnormal, I'll open a ticket, I personally think it's way to much for a stateless service.

@salanki Please open an issue. This is definitely abnormal. Also if you have enabled `--enable-pprof` please share heap dump as well. If not, if possible please enable the flag and share the dump.

aauren:
@murali-reddy we see this at a smaller level in our clusters, when I've done profiling it looks to be caused by gobpg. There have been a few leaks patched in more recent versions of gobgp. We were waiting on your upgrade task to test.

aauren:
Moved to 1.2 as these PRs are large items and we're focusing on library upgrades for 1.1

aauren:
This goes hand-in-hand with #1050 as this is the documentation component of the code changes carried in #1050

aauren:
Closing as stale - Unfortunately the project hasn't had time to properly review this work and do something productive with it, but at this point it seems unlikely that we would come back to this. In the meantime, the docs still exist on the cloudnativer fork if people want to reference them in the future.

Issue #945, opened 2020-07-01
Network policy controller performance improvement plan


jimmy-zh:
**What is the problem:**

Under the current implementation mechanism, NPC converts each "NetworkPolicyIngressRule" or "NetworkPolicyEgressRule" into at least one iptables rule. When creating lots of "NetworkPolicy***Rule" in the cluster, a large number of iptables rules will be generated on the nodes. 

In this case, the following series of things will happen:

* a synchronization of NPC will last too long
* the heartbeat will be lost
* the health check controller will report an unhealthy `livenessProbe`
* the current kube-router pod will be terminated
* the legacy iptables rules will not get a chance to be cleaned up
* a new kube-router pod will be created, and it will generate a new version of iptables rules

The continuous repetition of the above process will cause the number of iptables rules to be out of control

**How to reproduce it:**

One of our customers encountered this situation when their cluster had 1w+ network policy rules, which ultimately resulted in hundreds of thousands more node iptables rules. A synchronization can last for several hours.

```shell
++++++++++++++++++++++++++++++++++++++++++++
+ Summary information of iptables firewall +
++++++++++++++++++++++++++++++++++++++++++++
---| Sum of current iptables chains : 71326   (chains)
---| Sum of active iptable rules : 675056    (rules)
+++++++++++++++++++++++++++++++++++++++
+ Table 'filter' of iptables firewall +
+++++++++++++++++++++++++++++++++++++++
-- Chain KUBE-NWPLCY-222DE25K6LWX4JB5 : 3    (rules)
-- Chain KUBE-NWPLCY-223DG2STPNOA2UZY : 24    (rules)
-- Chain KUBE-NWPLCY-223FVCR7GWOZ4HV5 : 1    (rules)
-- Chain KUBE-NWPLCY-223L5HUTKMBCRWZO : 1    (rules)
-- Chain KUBE-NWPLCY-223OXF4BC6PL76SR : 20    (rules)
-- Chain KUBE-NWPLCY-226IECNXKCLLYRGR : 1    (rules)
-- Chain KUBE-NWPLCY-227KWHSNCFJW42ZZ : 15    (rules)
-- Chain KUBE-NWPLCY-227QZ6AFBRB72DTT : 1    (rules)
-- Chain KUBE-NWPLCY-22CFJFWPPCQGZLSN : 0    (rules)
-- Chain KUBE-NWPLCY-22ENZTTEK3FU2BPH : 59    (rules)
[...]

PR #947, opened 2020-07-02
Simplify ipset operations in NPC

jimmy-zh:
related issue: #945

1. Completely move the logic related to ipset from buildNetworkPoliciesInfo to syncNetworkPolicyChains
2. Do not set timeout option for ipset entries because the ipset itself already has a default value


aauren:
This issue should be mitigated by batching ipset operations in #1010

PR #950, opened 2020-07-06
Isolate ipsets between different synchronizations

### LINKED ISSUE #949: NPC's new synchronization refreshes the ipsets that are still in use by the current iptables rules
The ipsets used in NPC are named based on certain constant factors, which will cause a new synchronization to refresh the ipsets that are still in use by the current iptables rules. We need to isolate the ipsets used by different synchronizations to ensure consistency.

jimmy-zh:
Fixes #949 

Achieve ipset isolation between different synchronizations by introducing the version factor.

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -5,7 +5,7 @@ import (
 	"encoding/base32"
 	"errors"
 	"fmt"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


aauren:
@jimmy-zh thanks for your contribution! Initial overview of this PR looks good. Pending more detailed review when we enter the 1.2 release cycle.

aauren:
Closing pending exploration of parent issue #949 and seeing if this is still a practical problem now that NPC executes in less than a second in most cases.

PR #951, opened 2020-07-07
[WIP] Merge all egress NWPLCY iptables rules into four specific rules

jimmy-zh:
Merge all egress NWPLCY iptables rules into four specific rules.


aauren:
Thanks for the help here @jimmy-zh!  In the end, we went a very different direction with the NPC controller. Please review where the project is post 1.2.X.

Closing for the same reason as parent issue, see: https://github.com/cloudnativelabs/kube-router/issues/945#issuecomment-802285655

If there are still pieces of this PR that are needed, please update issue #945 and I'll re-open and we can discuss so that there is no additional wasted effort.

Issue #910, opened 2020-05-26
IPVS timeout is not updated without service recreation after sessionAffinityConfig.clientIP.timeoutSeconds update


rearden-steel:
Kube-router does not update this field in IPVS when service is updated.
If you delete/create the service it is configured correctly.

aauren:
Thanks for reporting @rearden-steel 

I was able to reproduce and submitted a fix that I tested against my own cluster.

rearden-steel:
@aauren thanks!

PR #952, opened 2020-07-10
fix(nsc): update IPVS svc when timeout changes

aauren:
@murali-reddy this is a small fix for #910

We weren't checking the equality for `persistentTimeout` as part of checking to see if the persistence settings needed to be updated in the IPVS service.

I also removed a double handling of the ipvs flag changes.


murali-reddy:
LGTM

PR #953, opened 2020-07-10
Allow Setting BGP Restart Time

aauren:
Same as #799 with some minor fixups so that it will build since it had been a long time since the author had submitted and wasn't responding on the PR.

FYI @murali-reddy


aauren:
Partial fix for #794

Issue #742, opened 2019-06-02
kube-router can't bind to specific IPv6 address


dsx:
kube-router can't bind to specific IPv6 address:

```
I0602 19:37:54.897474       1 kube-router.go:207] Running /usr/local/bin/kube-router version v0.3.1, built on 2019-05-12T09:22:07+0000, go1.10.8
I0602 19:37:55.558091       1 network_routes_controller.go:986] Found annotation `kube-router.io/bgp-local-addresses` on node object so BGP will listen on local IP's: fd00:1:2:3::1
E0602 19:37:55.693395       1 network_routes_controller.go:207] Failed to enable netfilter for bridge. Network policies and service proxy may not work: exit status 1
I0602 19:37:55.693472       1 network_routes_controller.go:223] Starting network route controller
I0602 19:37:55.727357       1 network_routes_controller.go:686] Found rr.server for the node to be 42 from the node annotation
I0602 19:37:55.727944       1 network_routes_controller.go:803] Could not find BGP peer password info in the node's annotations. Assuming no passwords.
time="2019-06-02T19:37:55Z" level=info msg="Add a peer configuration for:fd00:1:2:3::5" Topic=Peer
time="2019-06-02T19:37:55Z" level=info msg="Add a peer configuration for:fd00:1:2:3::10" Topic=Peer
time="2019-06-02T19:37:55Z" level=warning msg="listen failed" Error="listen tcp: address fd00:1:2:3::1:50051: too many colons in address" Key="fd00:1:2:3::1:50051" Topic=grpc
```

aauren:
Closing as stale, but I believe that this has to do with the version of GoBGP that we are currently locked into that doesn't support IPv6 in next-hop fields. We should be upgrading as soon as the 1.0 release addresses a few last bugs.

PR #956, opened 2020-07-22
patched https://github.com/cloudnativelabs/kube-router/issues/742

supu2:
modify ip address handling for ipv6 connection

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -718,8 +718,11 @@ func (nrc *NetworkRoutingController) startBgpServer() error {
 
 	nrc.bgpServer = gobgp.NewBgpServer()
 	go nrc.bgpServer.Serve()
-
-	g := bgpapi.NewGrpcServer(nrc.bgpServer, nrc.nodeIP.String()+":50051"+","+"127.0.0.1:50051")
+	ip := nrc.nodeIP.String()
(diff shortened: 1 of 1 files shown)


pkg/controllers/routing/network_routes_controller.go
```diff
@@ -718,8 +718,11 @@ func (nrc *NetworkRoutingController) startBgpServer() error {
 
 	nrc.bgpServer = gobgp.NewBgpServer()
 	go nrc.bgpServer.Serve()
-
-	g := bgpapi.NewGrpcServer(nrc.bgpServer, nrc.nodeIP.String()+":50051"+","+"127.0.0.1:50051")
+	ip := nrc.nodeIP.String()
+	if len(strings.Split(ip, ":")) >= 0 {
```
aauren, line 722: @supu2 This logic will result in the IP being bracketed regardless of whether it's an IPv4 or IPv6 address as even if the string doesn't contain a colon `strings.Split` will result in a length of at least 1. See an example here: https://play.golang.org/p/k10huoTr9rE

Once this is updated to look for a `len` greater than 1 this seems fine to me. While I doubt that this is the only thing required for IPv6 to work in kube-router, it would fix a more glaring bug.

@murali-reddy I traced the code back, and it looks like GoBGP passes this string all the way back to net.dial.Listen in which bracketing an IPv6 address string is the correct thing to do.

aauren:
Closing as author never responded to feedback.

PR #958, opened 2020-07-28
Add --random-fully to MASQ iptables rules to mitigate conntrack issues

coufalja:
To mitigate conntrack issues in Kubernetes.
See: https://tech.xing.com/a-reason-for-unexplained-connection-timeouts-on-kubernetes-docker-abd041cf7e02 and kubernetes/kubernetes#56903 for details.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -98,12 +98,12 @@
   version = "v0.5.2"
 
 [[projects]]
-  digest = "1:df4af3a8e15cd72b068b8d898598001935edce4eac4eb5a3d7fd906036123ff4"
+  digest = "1:2debc00eaf8ffdbfb0570a791a84e64a3cfc146f5e61b54d4abd65c2de364552"
   name = "github.com/coreos/go-iptables"
   packages = ["iptables"]
   pruneopts = "UT"
-  revision = "259c8e6a4275d497442c721fa52204d7a58bde8b"
-  version = "v0.2.0"
+  revision = "f901d6c2a4f2a4df092b98c33366dfba1f93d7a0"
+  version = "v0.4.5"
 
 [[projects]]
   digest = "1:a2c1d0e43bd3baaa071d1b9ed72c27d78169b2b269f71c105ac4ba34b1be4a39"
@@ -1135,6 +1135,7 @@
     "k8s.io/api/core/v1",
     "k8s.io/api/extensions/v1beta1",
     "k8s.io/api/networking/v1",
+    "k8s.io/apimachinery/pkg/api/resource",
     "k8s.io/apimachinery/pkg/apis/meta/v1",
     "k8s.io/apimachinery/pkg/labels",
     "k8s.io/apimachinery/pkg/util/intstr",
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -1225,6 +1225,9 @@ func (nsc *NetworkServicesController) ensureMasqueradeIptablesRule() error {
 		return errors.New("Failed to initialize iptables executor" + err.Error())
 	}
 	var args = []string{"-m", "ipvs", "--ipvs", "--vdir", "ORIGINAL", "--vmethod", "MASQ", "-m", "comment", "--comment", "", "-j", "SNAT", "--to-source", nsc.nodeIP.String()}
+	if iptablesCmdHandler.HasRandomFully() {
+		args = append(args, "--random-fully")
+	}
 	if nsc.masqueradeAll {
 		err = iptablesCmdHandler.AppendUnique("nat", "POSTROUTING", args...)
 		if err != nil {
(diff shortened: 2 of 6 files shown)


coufalja:
Docker image for testing purposes: `wanderadock/kube-router:amd64-wnd-randomize`

murali-reddy:
@coufalja thanks for the PR. `--random-fully` option make sense. Changes look good but on upgrade we would end up with two rules. Old rule plus new rule with `--random-fully` option set. Are you expecting users to flush the iptables chains?

Gopkg.lock
```diff
@@ -98,12 +98,12 @@
...
 [[projects]]
-  digest = "1:df4af3a8e15cd72b068b8d898598001935edce4eac4eb5a3d7fd906036123ff4"
+  digest = "1:2debc00eaf8ffdbfb0570a791a84e64a3cfc146f5e61b54d4abd65c2de364552"
   name = "github.com/coreos/go-iptables"
   packages = ["iptables"]
   pruneopts = "UT"
-  revision = "259c8e6a4275d497442c721fa52204d7a58bde8b"
-  version = "v0.2.0"
+  revision = "f901d6c2a4f2a4df092b98c33366dfba1f93d7a0"
+  version = "v0.4.5"
```
mrueg, line 106: you're setting a constraint to v0.4.1 but this updates it to 0.4.5?

coufalja:
@murali-reddy The way I deployed was going through full rolling-update of a cluster so I didn't hit the issue. But yeah It would be great to actually allow users to update in place. What needs to be done? Adding the old one to pkg/controllers/routing/pod_egress.go:85 and pkg/controllers/proxy/network_services_controller.go:1268 ?

Gopkg.lock
coufalja: Huh? I just run dep I have no idea how it happened, will fix.

coufalja: Oh I see... actually the constraint in dep reads like this `^0.4.1` so 0.4.5 is right

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -1269,6 +1276,11 @@ func (nsc *NetworkServicesController) deleteBadMasqueradeIptablesRules() error {
 		{"-m", "ipvs", "--ipvs", "--vdir", "ORIGINAL", "--vmethod", "MASQ", "-m", "comment", "--comment", "", "!", "-s", nsc.podCidr, "!", "-d", nsc.podCidr, "-j", "MASQUERADE"},
 	}
 
+	// If random fully is supported remove the original rule as well
+	if iptablesCmdHandler.HasRandomFully() {
+		argsBad = append(argsBad, []string{"-m", "ipvs", "--ipvs", "--vdir", "ORIGINAL", "--vmethod", "MASQ", "-m", "comment", "--comment", "", "-j", "SNAT", "--to-source", nsc.nodeIP.String()})
```
aauren, line 1281: Don't you have to cover the podCidr case here too? (i.e. the 2nd change to this file above?)

coufalja: Thanks you are absolutely right!

coufalja: Done

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -1269,6 +1276,16 @@ func (nsc *NetworkServicesController) deleteBadMasqueradeIptablesRules() error {
 		{"-m", "ipvs", "--ipvs", "--vdir", "ORIGINAL", "--vmethod", "MASQ", "-m", "comment", "--comment", "", "!", "-s", nsc.podCidr, "!", "-d", nsc.podCidr, "-j", "MASQUERADE"},
 	}
 
+	// If random fully is supported remove the original rules as well
+	if iptablesCmdHandler.HasRandomFully() {
+		argsBad = append(argsBad, []string{"-m", "ipvs", "--ipvs", "--vdir", "ORIGINAL", "--vmethod", "MASQ", "-m", "comment", "--comment", "", "-j", "SNAT", "--to-source", nsc.nodeIP.String()})
+
```
aauren, line 1282: It looks like there is some dangling whitespace on this line that gofmt is screaming about. Once that is fixed though, I don't see any other problems with this PR.

coufalja: Oh missed the whitespace, done

coufalja:
@aauren Thanks, just a question when I can expect this to be released? in the 1.1 or you plan to release 1.0.1 with a round of minor fixes?

mrueg:
This would be an additional feature, so it will be in 1.1

murali-reddy:
> I tested this locally on our nodes with both iptables that are able to support --random-full and iptables that are not. I also tested the upgrade procedure and found that it cleaned the nodes appropriately when upgrading from iptables that didn't support it to iptables that did support it.

thanks @aauren

PR #960, opened 2020-07-31
cmd/kube-router: Drop superfluous flag usage

mrueg:
@murali-reddy @aauren just a small finding, when looking how commands get parsed. It looks like we mixed flag and pflag here.


mrueg:
seems like this was done on purpose.

Issue #137, opened 2017-08-25
remove --cluster-cidr from kube-router options


murali-reddy:
We fixed Kube-router #107 to make --cluster-cidr optional, and implicitly figure out the CIDR used for pods. We made it optional instead of removing totally so that KOPS which has hard coded manifest embedded with in will not fail.

I fixed KOPS not to use `--cluster-cidr` with PR https://github.com/kubernetes/kops/pull/3263

However old version of kops will still fail, so we need to keep the flag optional for some time (atleast one two release of kops)

Opening placeholder bug for pending work to remove the flag tottally.

murali-reddy:
Addressed by https://github.com/cloudnativelabs/kube-router/pull/961

PR #961, opened 2020-07-31
fix(options): remove deprecated cluster-cidr option

aauren:
@murali-reddy 

In Slack the presence of this option confused a user that tried to use it to set their cluster-cidr instead of setting it on `kube-controller-manager`. I didn't realize that we still had this option around any longer, but it looks like it's functionality was ripped out here: https://github.com/cloudnativelabs/kube-router/pull/111 but then the option was never removed, so now users can set it, but it's effectively a no-op.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -48,7 +48,6 @@ Usage of kube-router:
       --cache-sync-timeout duration                   The timeout for cache synchronization (e.g. '5s', '1m'). Must be greater than 0. (default 1m0s)
       --cleanup-config                                Cleanup iptables rules, ipvs, ipset configuration and exit.
       --cluster-asn uint                              ASN number under which cluster nodes will run iBGP.
-      --cluster-cidr string                           CIDR range of pods in the cluster. It is used to identify traffic originating from and destinated to pods.
       --disable-source-dest-check                     Disable the source-dest-check attribute for AWS EC2 instances. When this option is false, it must be set some other way. (default true)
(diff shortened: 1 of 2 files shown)


mrueg:
Can you remove those as well? 

```
./docs/user-guide.md:- If you choose to use kube-router for pod-to-pod network connectivity then Kubernetes controller manager need to be configured to allocate pod CIDRs by passing `--allocate-node-cidrs=true` flag and providing a `cluster-cidr` (i.e. by passing --cluster-cidr=10.1.0.0/16 for e.g.)
./docs/generic.md:If you choose to use kube-router for pod-to-pod network connectivity then [kube-controller-manager](https://kubernetes.io/docs/reference/generated/kube-controller-manager/) need to be configured to allocate pod CIDRs by passing `--allocate-node-cidrs=true` flag and providing a `cluster-cidr` (i.e. by passing --cluster-cidr=10.32.0.0/12 for e.g.)
./docs/generic.md:    --cluster-cidr=10.32.0.0/12
```

aauren:
@mrueg Those came up in a search for the `--cluster-cidr` option for me too but, unless I'm mistaken, they are talking about similar options that have to be set on `kube-controller-manager` in order for it to allocate pod-cidrs on the node object which kube-router picks up and uses internally.

murali-reddy:
LGTM. Fixes https://github.com/cloudnativelabs/kube-router/issues/137

mrueg:
Whops, you're right @aauren!

Issue #962, opened 2020-08-04
Service proxy causing high CPU usage with ~2.7K services


iamakulov:
Hey,

We’re running a Kubernetes cluster with ~150 nodes and ~2.7K services. Each service typically matches one pod. Service endpoints are updated quite often (e.g., due to pod restarts).

Each time a service endpoint is updated, kube-router seems to perform a full resync of services. Due to the number of services, the resync takes ~16 seconds on a `m5a.large` EC2 instance:

<img width="1098" alt="Screen Shot 2020-08-04 at 12 24 34" src="https://user-images.githubusercontent.com/2953267/89278229-81b2e700-d64e-11ea-9e73-e47fd8aeddb2.png">

While doing the resync, kube-router aggressively consumes the CPU:

<img width="1214" alt="Screen Shot 2020-08-04 at 12 33 57" src="https://user-images.githubusercontent.com/2953267/89278476-d5bdcb80-d64e-11ea-9af1-22f81f39f366.png">

which affects other pods running on that instance.

Is there any way to reduce the CPU usage _or_ make service resyncs happen faster? The ultimate goal is to make sure kube-router doesn’t affect other pods from that node.

(CPU limits are not a solution, unfortunately. We don’t want to apply CPU limits to `kube-router` because that’d make IPVS resyncs longer, and with long resyncs, we’d start experiencing noticeable traffic issues. E.g. if a service endpoint changes, and it takes kube-router 1 minute to perform an IPVS resync, the traffic to that service will be blackholed or rejected for the full minute.)

iamakulov:
Just for the sake of reference, our kube-router configuration is pretty standard:

<details>

```yaml
apiVersion: extensions/v1beta1
kind: DaemonSet
metadata:
  labels:
    k8s-app: kube-router
    tier: node
  name: kube-router
  namespace: kube-system
spec:
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      k8s-app: kube-router
      tier: node
  template:
    metadata:
      creationTimestamp: null
      labels:
        k8s-app: kube-router
        tier: node
    spec:
      containers:
      - args:
        - --run-router=true
        - --run-firewall=true
        - --run-service-proxy=true
        - --kubeconfig=/var/lib/kube-router/kubeconfig
        - --bgp-graceful-restart
        - -v=1
        - --metrics-port=12013
        env:
        - name: NODE_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
[...]

Issue #376, opened 2018-04-09
bad routing from host to service IP on same host


phemmer:
Currently if you try to access a service IP from a host in which that IP is present, it will randomly fail.

For example:
```
# oc get svc -o wide
NAME         TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)   AGE       SELECTOR
webconsole   ClusterIP   10.128.64.184   <none>        443/TCP   3m        webconsole=true

# oc get pods -l webconsole=true -o wide
NAME                          READY     STATUS    RESTARTS   AGE       IP           NODE
webconsole-7bb99b5c76-44pzs   1/1       Running   0          1h        10.128.0.3   node01
webconsole-7bb99b5c76-z9qhc   1/1       Running   0          1h        10.128.2.3   node02
```
From node01:
```
# psgrep kube-router -- h -o command
/usr/local/bin/kube-router --run-router=true --run-firewall=true --run-service-proxy=true --kubeconfig=/var/lib/kube-router/kubeconfig --bgp-graceful-restart=true --cluster-cidr=10.128.0.0/18 --cluster-asn=65101 --peer-router-ips=10.0.33.218 --peer-router-asns=65101 --peer-router-passwords= --enable-overlay=false --enable-pod-egress=false --advertise-cluster-ip=true --advertise-external-ip=true --advertise-loadbalancer-ip=true --hairpin-mode

# ip addr show | grep 10.128.64.184
    inet 10.128.64.184/32 brd 10.128.64.184 scope link kube-dummy-if

# ipvsadm -ln -t 10.128.64.184:443 
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  10.128.64.184:443 rr
  -> 10.128.0.3:8443              Masq    1      0          0         
  -> 10.128.2.3:8443              Masq    1      0          0         

# curl -k https://webconsole.openshift-web-console.svc/healthz
curl: (7) Failed connect to webconsole.openshift-web-console.svc:443; Connection timed out
```
At the same time from node02:
```
# tcpdump -nn -i ens192 'host 10.128.64.184'
19:03:19.637049 IP 10.128.64.184.44828 > 10.128.2.3.8443: Flags [S], seq 3022741939, win 43690, options [mss 65495,sackOK,TS val 12590808 ecr 0,nop,wscale 7], length 0
[...]

Issue #370, opened 2018-04-06
route missing upon pod restart


drobinson123:
When a kube-router pod is restarted it appears to lose at least one and sometimes several routes at either shutdown or startup. I have several k8s clusters, both using kube-router v0.1.0, and both have encountered this problem over the past few days. CoreOS auto-updates cause the host, and therefore kube-router, to restart -- that seems to trigger the problem.

The host with the missing routes is `ip-172-21-91-59.us-west-2.compute.internal` -- it auto-updated to CoreOS 1688.5.3. Before the update it was running CoreOS 1.6.32.2.1, the same version as other hosts.

Here's the cluster after the node auto-updated (note the last line, it's the host that updated):

```
$ kubectl get nodes -o wide
NAME                                          STATUS    ROLES     AGE       VERSION   EXTERNAL-IP   OS-IMAGE                                        KERNEL-VERSION   CONTAINER-RUNTIME
ip-172-21-46-101.us-west-2.compute.internal   Ready     node      1d        v1.8.9    <none>        Container Linux by CoreOS 1632.2.1 (Ladybug)    4.14.16-coreos   docker://17.9.1
ip-172-21-55-253.us-west-2.compute.internal   Ready     node      1d        v1.8.9    <none>        Container Linux by CoreOS 1632.2.1 (Ladybug)    4.14.16-coreos   docker://17.9.1
ip-172-21-58-9.us-west-2.compute.internal     Ready     master    24d       v1.8.9    <none>        Container Linux by CoreOS 1632.3.0 (Ladybug)    4.14.19-coreos   docker://17.9.1
ip-172-21-62-29.us-west-2.compute.internal    Ready     node      1d        v1.8.9    <none>        Container Linux by CoreOS 1632.2.1 (Ladybug)    4.14.16-coreos   docker://17.9.1
ip-172-21-71-225.us-west-2.compute.internal   Ready     node      1d        v1.8.9    <none>        Container Linux by CoreOS 1632.2.1 (Ladybug)    4.14.16-coreos   docker://17.9.1
ip-172-21-91-59.us-west-2.compute.internal    Ready     node      1d        v1.8.9    <none>        Container Linux by CoreOS 1688.5.3 (Rhyolite)   4.14.32-coreos   docker://17.12.1-ce
```

k8s shows the kube-router pod restarted (the first line):

```
$ kubectl -n kube-system get pods | grep kube-router
[...]

PR #965, opened 2020-08-06
Improve the ipAddrAdd performance

iamakulov:
This PR replaces the `Exec()` call inside `ipAddrAdd()` with `netlink.RouteReplace()`.

When calling `ipAddrAdd` in a loop (e.g., inside `setupClusterIPServices`), this gives a significant performance boost.

Ref: https://github.com/cloudnativelabs/kube-router/issues/962

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -81,7 +81,7 @@ type ipvsCalls interface {
 }
 
 type netlinkCalls interface {
-	ipAddrAdd(iface netlink.Link, ip string, addRoute bool) error
+	ipAddrAdd(iface netlink.Link, ip string) error
 	ipAddrDel(iface netlink.Link, ip string) error
 	prepareEndpointForDsr(containerID string, endpointIP string, vip string) error
 	getKubeDummyInterface() (netlink.Link, error)
@@ -122,30 +122,13 @@ func (ln *linuxNetworking) ipAddrDel(iface netlink.Link, ip string) error {
 // utility method to assign an IP to an interface. Mainly used to assign service VIP's
 // to kube-dummy-if. Also when DSR is used, used to assign VIP to dummy interface
 // inside the container.
-func (ln *linuxNetworking) ipAddrAdd(iface netlink.Link, ip string, addRoute bool) error {
+func (ln *linuxNetworking) ipAddrAdd(iface netlink.Link, ip string) error {
 	naddr := &netlink.Addr{IPNet: &net.IPNet{IP: net.ParseIP(ip), Mask: net.IPv4Mask(255, 255, 255, 255)}, Scope: syscall.RT_SCOPE_LINK}
 	err := netlink.AddrAdd(iface, naddr)
 	if err != nil && err.Error() != IfaceHasAddr {
 		glog.Errorf("Failed to assign cluster ip %s to dummy interface: %s",
 			naddr.IPNet.IP.String(), err.Error())
 		return err
-	}
-
-	// When a service VIP is assigned to a dummy interface and accessed from host, in some of the
-	// case Linux source IP selection logix selects VIP itself as source leading to problems
-	// to avoid this an explicit entry is added to use node IP as source IP when accessing
-	// VIP from the host. Please see https://github.com/cloudnativelabs/kube-router/issues/376
-
-	if !addRoute {
-		return nil
-	}
-
-	// TODO: netlink.RouteReplace which is replacement for below command is not working as expected. Call succeeds but
-	// route is not replaced. For now do it with command.
-	out, err := exec.Command("ip", "route", "replace", "local", ip, "dev", KubeDummyIf, "table", "local", "proto", "kernel", "scope", "host", "src",
-		NodeIP.String(), "table", "local").CombinedOutput()
-	if err != nil {
-		glog.Errorf("Failed to replace route to service VIP %s configured on %s. Error: %v, Output: %s", ip, KubeDummyIf, err, out)
 	}
 	return nil
 }
@@ -201,23 +184,24 @@ func newLinuxNetworking() (*linuxNetworking, error) {
 
 // NetworkServicesController struct stores information needed by the controller
 type NetworkServicesController struct {
-	nodeIP              net.IP
-	nodeHostName        string
-	syncPeriod          time.Duration
-	mu                  sync.Mutex
-	serviceMap          serviceInfoMap
-	endpointsMap        endpointsInfoMap
-	podCidr             string
-	excludedCidrs       []net.IPNet
-	masqueradeAll       bool
-	globalHairpin       bool
-	ipvsPermitAll       bool
-	client              kubernetes.Interface
-	nodeportBindOnAllIP bool
-	MetricsEnabled      bool
-	ln                  LinuxNetworking
-	readyForUpdates     bool
-	ProxyFirewallSetup  *sync.Cond
+	nodeIP                net.IP
+	nodeHostName          string
+	serviceClusterIPRange string
+	syncPeriod            time.Duration
+	mu                    sync.Mutex
+	serviceMap            serviceInfoMap
+	endpointsMap          endpointsInfoMap
+	podCidr               string
+	excludedCidrs         []net.IPNet
+	masqueradeAll         bool
+	globalHairpin         bool
+	ipvsPermitAll         bool
+	client                kubernetes.Interface
+	nodeportBindOnAllIP   bool
+	MetricsEnabled        bool
+	ln                    LinuxNetworking
+	readyForUpdates       bool
+	ProxyFirewallSetup    *sync.Cond
 
 	// Map of ipsets that we use.
 	ipsetMap map[string]*utils.Set
@@ -346,6 +330,11 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
 		glog.Error("Error setting up ipvs firewall: " + err.Error())
 	}
 	nsc.ProxyFirewallSetup.Broadcast()
+
+	err = nsc.enforceRightServiceInterface()
+	if err != nil {
+		glog.Error(err.Error())
+	}
 
 	gracefulTicker := time.NewTicker(5 * time.Second)
 	defer gracefulTicker.Stop()
@@ -1006,7 +995,7 @@ func (ln *linuxNetworking) prepareEndpointForDsr(containerID string, endpointIP 
 	}
 
 	// assign VIP to the KUBE_TUNNEL_IF interface
-	err = ln.ipAddrAdd(tunIf, vip, false)
+	err = ln.ipAddrAdd(tunIf, vip)
 	if err != nil && err.Error() != IfaceHasAddr {
 		err = netns.Set(hostNetworkNamespaceHandle)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/controllers/proxy/network_services_controller.go
```diff
@@ -140,12 +141,17 @@ func (ln *linuxNetworking) ipAddrAdd(iface netlink.Link, ip string, addRoute boo
 		return nil
 	}
 
-	// TODO: netlink.RouteReplace which is replacement for below command is not working as expected. Call succeeds but
-	// route is not replaced. For now do it with command.
```
iamakulov, line 144: I believe this comment by @murali-reddy doesn’t hold true anymore. (Happy to be proven wrong.)

What I did to verify that `netlink.RouteReplace` works as intended:

1. Exec into a `kube-router` pod and install go and git:

   ```
   apk add go git
   ```

2. Create a file called `test-netlink.go`:

	```go
   	package main

	import (
		"github.com/golang/glog"
		"github.com/vishvananda/netlink"
		"golang.org/x/sys/unix"
		"net"
	)

	func main() {
		err := netlink.RouteReplace(&netlink.Route{
			Dst:       &net.IPNet{IP: net.ParseIP("5.5.5.5"), Mask: net.IPv4Mask(255, 255, 255, 255)},
			LinkIndex: 1,
			Table:     unix.RT_TABLE_LOCAL,
			Protocol:  unix.RTPROT_KERNEL,
			Scope:     unix.RT_SCOPE_HOST,
			Type:      unix.RTN_LOCAL,
			Src:       net.ParseIP("127.0.0.1"),
		})
		if err != nil {
			glog.Errorf("Failed to replace route to service. Error: %v", err)
		}
	}
	```

3. Install dependencies and pin `netlink` to the version which `kube-router` uses:

   ```sh
   go get
   cd $(go env GOPATH)/src/github.com/vishvananda/netlink
   git checkout f504738125a57f35f87fc30fb69b8df75237ccde # Version from Gopkg.lock
   ```

4. Run `test-netlink.go` and verify that the route got added:

   ```
   cd -
   go run test-netlink.go
   ip route show table all | grep 5.5.5.5
   # Should print: local 5.5.5.5 dev lo table local proto kernel scope host src 127.0.0.1
   ```

5. Change the source IP address in `test-netlink.go`, re-run it, and verify that the route gets replaced correctly:

   ```sh
   sed -i 's/127.0.0.1/172.17.0.2/' test-netlink.go
   go run test-netlink.go
   ip route show table all | grep 5.5.5.5
   # Should print: local 5.5.5.5 dev lo table local proto kernel scope host src 172.17.0.2
   ```

Is this a correct way to verify that?

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -140,12 +141,17 @@ func (ln *linuxNetworking) ipAddrAdd(iface netlink.Link, ip string, addRoute boo
...
-	out, err := exec.Command("ip", "route", "replace", "local", ip, "dev", KubeDummyIf, "table", "local", "proto", "kernel", "scope", "host", "src",
-		NodeIP.String(), "table", "local").CombinedOutput()
+	err = netlink.RouteReplace(&netlink.Route{
+		Dst:       &net.IPNet{IP: net.ParseIP(ip), Mask: net.IPv4Mask(255, 255, 255, 255)},
+		LinkIndex: iface.Attrs().Index,
+		Table:     unix.RT_TABLE_LOCAL,
+		Protocol:  unix.RTPROT_KERNEL,
+		Scope:     unix.RT_SCOPE_HOST,
+		Type:      unix.RTN_LOCAL,
```
iamakulov, line 150: These parameters come from running the original `ip route replace ...` command through strace:

```
strace ip route replace local 10.97.216.74 dev kube-dummy-if table local proto kernel scope host src 172.17.0.2
```

which prints the system call with needed constants:

```
sendmsg(3, {msg_name={sa_family=AF_NETLINK, nl_pid=0, nl_groups=00000000}, msg_namelen=12, msg_iov=[{iov_base={{len=52, type=RTM_NEWROUTE, flags=NLM_F_REQUEST|NLM_F_ACK|NLM_F_REPLACE|NLM_F_CREATE, seq=1596701111, pid=0}, {rtm_family=AF_INET, rtm_dst_len=32, rtm_src_len=0, rtm_tos=0, rtm_table=RT_TABLE_LOCAL, rtm_protocol=RTPROT_KERNEL, rtm_scope=RT_SCOPE_HOST, rtm_type=RTN_LOCAL, rtm_flags=0}, [{{nla_len=8, nla_type=RTA_DST}, inet_addr("10.97.216.74")}, {{nla_len=8, nla_type=RTA_PREFSRC}, inet_addr("172.17.0.2")}, {{nla_len=8, nla_type=RTA_OIF}, if_nametoindex("kube-dummy-if")}]}, iov_len=52}], msg_iovlen=1, msg_controllen=0, msg_flags=0}, 0) = 52
```

murali-reddy:
@iamakulov thanks for your PR

We ran into some issues with `netlink.RouteReplace` on some distros and specific kernel versions. Please see https://github.com/cloudnativelabs/kube-router/issues/370#issuecomment-379415278

Action done by netlink routereplace was done to address specirfic issue with using service proxy in case of nodes with multiple network interface. This is the problem https://github.com/kubernetes/kubeadm/issues/102#issuecomment-291532883 and recommended solution https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#check-network-adapters. That piece of code is trying to automate above requirement.

On hindsight adding a route by exec command or with netlink libray for each VIP just does not scale. Now that kube-router has `--service-cluster-ip-range` we should all together remove setting up route for each cluster IP and set one rule to use node IP as source when destination is from the `--service-cluster-ip-range`

Not sure you have motivation to work on the fix, but here is what i think would be better approach

- remove whole `ip route` logic from the `ipAddrAdd`
- if `--service-cluster-ip-range` specified setup an ip rule to use node ip as source when destination is in given range

Unfortunatley CIDR that is used for cluster Ip's can not inferred from the kubernetes API's to we will have to rely on user input through this parameter.

iamakulov:
@murali-reddy Just took a round on your suggestion! Did I get your idea right?

murali-reddy:
thanks @iamakulov PR looks good. Will give a quick test and will update the PR.

iamakulov:
(Rebased on master.)

murali-reddy:
@iamakulov Apolgies for the delay in review. We are trying to get 1.1 release out in next couple of days. Marking for 1.2 release which is focussed on performance enhancements.

iamakulov:
Tested this patch in production, and it doesn’t work as expected.

1\) For new nodes, `nsc.enforceRightServiceInterface` gets called when `kube-dummy-if` doesn’t exist yet – and fails.

2\) The change resurrects https://github.com/cloudnativelabs/kube-router/issues/376.

`netlink.AddrAdd` not only associates an address with an interface, but also adds a route to the routing table. And the route uses the same IP address in the `src` field:

```sh
$ ip addr add 10.10.10.10 dev kube-dummy-if
$ ip addr list kube-dummy-if
8: kube-dummy-if: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN group default 
    link/ether 56:51:fe:11:e3:02 brd ff:ff:ff:ff:ff:ff
    inet 10.10.10.10/32 brd 10.10.10.10 scope link kube-dummy-if
       valid_lft forever preferred_lft forever
$ ip route show table all | grep 10.10.10.10
local 10.10.10.10 dev kube-dummy-if table local proto kernel scope host src 10.10.10.10 # ← here, `src 10.10.10.10` is the problem
broadcast 10.10.10.10 dev kube-dummy-if table local proto kernel scope link src 10.10.10.10
```

Before, we were fixing this by calling `ip route replace` for each service IP. That was costly, so this PR removed that call and started calling `ip route replace <service-ip-cidr> ...`. But that resurrected https://github.com/cloudnativelabs/kube-router/issues/376 – because the CIDR route is less specific than `netlink.AddrAdd`’s routes and doesn’t have any effect.

---

To fix both issues, I’m going to change the approach:
* I’ll bring `ip route replace` back to `nsc.ln.ipAddrAdd()`
* and I’ll make `setupClusterIPServices` call `nsc.ln.ipAddrAdd()` only once (with the full services CIDR) instead of X times (for each service’s IP)

I’ll test this change in our system and then cherry-pick it here.

iamakulov:
Also, re: UX/DX.

This PR makes the `service-cluster-ip-range` param much more important. Before, if you specified it incorrectly (or if the default `10.96.0.0/16` value was wrong), it only affected network policies (which aren’t that common, I guess). Now, an incorrect value can break the service connectivity completely. (I learned this the hard way.)

Kubernetes already knows the `service-cluster-ip-range` value. Could kube-router retrieve it automatically and use the actual value (instead of a hard-coded default) in its defaults? That’d make life quite easier for a lot of people who aren’t that experienced with networking/Kubernetes administration (like me).

aauren:
@iamakulov While it's true that Kubernetes does know the service-cluster-ip-range value, as far as I know, it doesn't expose that information anywhere that is obtainable with through the API. Although, it may just be that it's not in a common place and I haven't run across it yet?

iamakulov:
> While it's true that Kubernetes does know the service-cluster-ip-range value, as far as I know, it doesn't expose that information anywhere that is obtainable with through the API. Although, it may just be that it's not in a common place and I haven't run across it yet?

Right, it appears that it’s not implemented: https://github.com/kubernetes/kubernetes/issues/25533 https://github.com/kubernetes/kubernetes/issues/46508

Would it be a viable idea to make `--service-cluster-ip-range` a required argument then? (To ensure that people _have_ to configure it properly – and don’t accidentally skip it and break their cluster.) I’m not sure how actually common is the `10.96.0.0/16` default, but at least kops uses `100.64.0.0/13` instead.

(If yes, it might be a good idea to also document how to retrieve this value – from `kubectl cluster-info dump | grep service-cluster-ip-range`.)

aauren (approved):
Finally coming back around this one. The only thing that I can see is that we should do as you suggest and make `--service-cluster-ip-range` a required argument as well as update the documentation in this area.

Otherwise this looks good to me.

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -1006,7 +995,7 @@ func (ln *linuxNetworking) prepareEndpointForDsr(containerID string, endpointIP
 	}
 
 	// assign VIP to the KUBE_TUNNEL_IF interface
-	err = ln.ipAddrAdd(tunIf, vip, false)
+	err = ln.ipAddrAdd(tunIf, vip)
```
aauren, line 998: @iamakulov You missed another one of these changes on line 1229: https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/proxy/network_services_controller.go#L1229

aauren (changes requested):

iamakulov:
Hey!

I’m afraid I left the project I was working for when I implemented this patch, so I won’t be able to contribute fixes for this PR. Sorry!

However, as I mentioned in https://github.com/cloudnativelabs/kube-router/pull/965#issuecomment-695139094, this patch actually doesn’t work as expected, so this PR can’t be merged as-is. In the project I was working for, we ended up taking the alternative approach:

> To fix both issues, I’m going to change the approach:
> 
> * I’ll bring `ip route replace` back to `nsc.ln.ipAddrAdd()`
> * and I’ll make `setupClusterIPServices` call `nsc.ln.ipAddrAdd()` only once (with the full services CIDR) instead of X times (for each service’s IP)
> 
> I’ll test this change in our system and then cherry-pick it here.

– and this one worked well. It’s been running in production for a few months with no observable issues.

Let me know if there’s anything else I can help with!

aauren:
Closing as this PR is incomplete and the author unfortunately had to move on. In NPC we've started taking the approach of using ipset save and ipset restore, and likely we'll work that same functionality into the NSC in the future as well if we wanted to approach this problem again.

Issue #934, opened 2020-06-16
ExternalIPs and LoadBalancer IPs No Longer Work with NetworkPolicy without Manual Specification


aauren:
Since this PR was merged: https://github.com/cloudnativelabs/kube-router/pull/914 only ClusterIPs and NodePorts are specifically bypassed in the INPUT chain for inter-cluster traffic. Essentially, for ClusterIPs and NodePorts the policy enforcement is specifically bypassed in INPUT and then enforced in OUTPUT after all of the VIPs have been disambiguated by IPVS into PodIP addresses. This allows users to use them for inter-cluster traffic using the standard pod and namespace selectors.

However, since ExternalIPs and LoadBalancer IPs were not added to the functionality of #914, they will still get processed in INPUT while the VIP still exists as the destination and will be denied unless manually allowed by an `ipBlock` statement AND a namespace / pod selector (since this will be enforced when it traverses the OUTPUT chain.

While I think that the intention of the Kubernetes network SIG is that all inter-cluster traffic be routed through ClusterIPs (which is why this is returned from built-in Kubernetes DNS when you lookup a service record) we have many users that lookup records from Consul. As it concerns Consul, we wish the services within Consul to be route-able both from within the cluster and from outside the cluster, so we only add NodePorts and ExternalIPs to the service definitions in Consul from Kubernetes. As such all of our containerized applications that utilize Consul attempt to route to in-cluster services via ExternalIPs or NodePorts.

murali-reddy:
@aauren Seems like reasonable use-case. So if i understood correctly consul acts as a service discovery for both in-cluster and out-of-the cluster services so you would want services external IP to be used by both in consistent manner?

aauren:
@murali-reddy Yes, that is correct.

zerkms:
Is it connected to https://github.com/cloudnativelabs/kube-router/issues/938 ?

aauren:
[...]

PR #967, opened 2020-08-06
Add hairpin support for externalIps

Tassatux:
Hello,

This pull request add support for externalIps hairpin rules, via an annotation on service in order to add extra NAT rules only when requested.

My use case is the following, I have a Consul server on K8S and exposed to hosts outside the cluster via an external IP. The server is configured to advertise the external IP instead of the pod one (or clusterIP), to let other Consul agent gossip with him, even if they're outside the K8S cluster.
Until now, there no problem, but the Consul server periodically check other servers, including himself, as part of the Consul autopilot feature and the lack of hairpin NAT rules for externalIPs make this self-check fail with a timeout.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -212,6 +212,11 @@ To enable hairpin traffic for Service `my-service`:
 kubectl annotate service my-service "kube-router.io/service.hairpin="
 ```
 
+If you want hairpin also apply on externalIPs declared for Service `my-service`:
+```
+kubectl annotate service my-service "kube-router.io/service.hairpin.externalips="
+```
+
 ## Direct server return
 
 Please read below blog on how to user DSR in combination with `--advertise-external-ip` to build highly scalable and available ingress.
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -46,12 +46,13 @@ const (
 	IpvsSvcFSched2    = "flag-2"
 	IpvsSvcFSched3    = "flag-3"
 
-	svcDSRAnnotation        = "kube-router.io/service.dsr"
-	svcSchedulerAnnotation  = "kube-router.io/service.scheduler"
-	svcHairpinAnnotation    = "kube-router.io/service.hairpin"
-	svcLocalAnnotation      = "kube-router.io/service.local"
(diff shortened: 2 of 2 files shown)


rdegez:
Thanks @Tassatux :-)
I wonder how this feature is affected by https://github.com/cloudnativelabs/kube-router/issues/934

aauren:
@murali-reddy I'm curious to get your thoughts on this. This was an issue that @rdegez asked me about on slack. Essentially, for a given service that they are trying to containerize (Consul in this case), it communicates with itself via ExternalIP which then fails as we only support hair pinning for ClusterIP and not ExternalIP.

This was a need that we had a long time ago, but they were for apps that we controlled so we just switched to using ClusterIP anywhere that we needed hair pinning. The only other way to work around this in the Consul use case is to use a hostPort which is do-able but it's less secure and more of a hassle.

I'm still not sure what I think about adding this functionality myself, in general I think that I'm a fan of saying that ExternalIPs should not be used for inter-cluster traffic, that being said, I know that it's sometimes unavoidable so maybe it's something that you are ok supporting?

murali-reddy:
> I'm still not sure what I think about adding this functionality myself, in general I think that I'm a fan of saying that ExternalIPs should not be used for inter-cluster traffic, that being said, I know that it's sometimes unavoidable so maybe it's something that you are ok supporting?

Agree it would be clean to demarcate cluster IP for east-west traffic and extenal IP etc for north-south traffic. But we seen cases where external IP needed to be in both also Kubernetes does not prescribe anything in this regard. I dont see strong reasons to not to support it.

rdegez:
Hi @aauren, 

I see you initially planned this for merge in v1.2 but removed it to put it ... well... nowhere it seems :-)
Any reason to not merge this PR in the next release since it looks like a simple change with 0.00% chance of affecting existing users ?
Are we missing something here ?

Cheers,

aauren:
Nothing intentional for the issue. I moved it from 1.2 to 1.3 because 1.2 was getting really big with all of the performance work and bug fixes that are already in that version, so to make 1.2 a little more realistic and obtainable I moved out the most recently requested feature issue plus some other stuff that hadn't been added to a project yet from the backlog.

I grant that it shouldn't impact existing users, but every PR that we merge takes a certain amount of time to vet and test.

If 1.2 starts going faster than expected I don't have a problem moving it back in.

aauren (approved):

PR #970, opened 2020-08-13
Permit ExternalIP on input

aauren:
This PR contains a couple of things:
1) Parse ClusterIP service range CIDR as a CIDR instead of as a string before passing it to iptables
2) Parse and validate NodePort ranges before before passing it to iptables
3) A side-effect of the above is that we can now allow ranges specified with a hyphen as well as the of the iptables colon `8080-8090` vs `8080:8090`. The hyphen form is a more k8s cohesive way to specify the NodePort range as it's the same format that kube-apiserver uses
4) Added support for defining ExternalIP ranges which addresses the problem found in #934 
5) I added unit tests for all of the above

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -84,7 +84,7 @@ Usage of kube-router:
       --run-router                                    Enables Pod Networking -- Advertises and learns the routes to Pods via iBGP. (default true)
       --run-service-proxy                             Enables Service Proxy -- sets up IPVS for Kubernetes Services. (default true)
       --service-cluster-ip-range string               CIDR value from which service cluster IPs are assigned. Default: 10.96.0.0/12 (default "10.96.0.0/12")
-      --service-node-port-range string                NodePort range. Default: 30000-32767 (default "30000:32767")
+      --service-node-port-range string                NodePort range specified with either a hyphen or colon (default "30000-32767")
   -v, --v string                                      log level for V logs (default "0")
   -V, --version                                       Print version information.
 ```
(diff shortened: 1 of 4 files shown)


aauren:
Right now this is based on #969 so that PR should be merged first, then I'll rebase.

aauren:
While testing this in one of my clusters I noticed that the `KUBE-ROUTER-INPUT` did not look the way that it should. At this point I don't believe that any of it was as the result of my code,  but I did fix a few bugs in this PR.

The first one is that when the code enters the `ensureRuleAtPosition` method that method attempts to verify that the rule is at the position it is supposed to be at by listing out the chain and then finding the rule in the chain and verifying it's index against the desired position. However, it did not take into account a couple of things:
1. That GoLang indexes arrays starting at position 0 while iptables indexes at position 1, so the position always appeared to be one off
2. I think that when #958 was merged and the iptables library was added it started to add the chain create and policy statements to the top of the chain. If these `-N` or `-P` statements were added it means that the position needs to be offset to take them into account.

To that end, I added the following commit to remit these issues: https://github.com/cloudnativelabs/kube-router/pull/970/commits/cd95772d70188ad3f29acb03c523135519c919e2

Additionally, I found that iptables no longer keeps the same parameter positions when it lists the tables back out, which means that if we try for a strict string comparison here: https://github.com/cloudnativelabs/kube-router/blob/f3b36eb4743866d2a0d9e71a1ffae8d03cf3aa0d/pkg/controllers/netpol/network_policy_controller.go#L224 that it often is unable to find match which means that when it reaches the end of the loop `ruleNo` is still `0` which then causes the NPC to mistakenly try to insert the rule again and then delete the wrong rule which means that when you end up with in iptables ends up being a bunch of duplicate rules.

To fix this I added a UUID to the comment of the rule and then match against the UUID in the following commit: https://github.com/cloudnativelabs/kube-router/pull/970/commits/2fd6b937dc19d76d0fbcc3617eb8d40f7543dd59

murali-reddy:
LGTM. Did basic sanity tests in both combinations

- kube-proxy as proxy and kube-router providing npc
- kube-router as all in one solution

PR #980, opened 2020-09-01
fix(nrc): update to use nextHop attribute from path

aauren:
@murali-reddy this should resolve the NextHop issue that I pointed out in slack where neighbor IP can be either the node itself in a full mesh scenario or a different BGP speaker acting as a route-reflector. This code ensures that the route added contains the nextHop similar to the logic that we had when using GoBGP 1.X

Tested it against our cluster and all of the routes went back to normal.


murali-reddy:
LGTM

PR #983, opened 2020-09-04
Embed iptablesCmdHandler in controllers

rkojedzinszky:
Kube-router is expected to run containerized, where changes in iptables behavior are not expected, so they can be initialized and stored in controllers.

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -67,7 +67,8 @@ type NetworkPolicyController struct {
 	healthChan              chan<- *healthcheck.ControllerHeartbeat
 	fullSyncRequestChan     chan struct{}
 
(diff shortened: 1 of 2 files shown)


aauren:
@rkojedzinszky Thanks for your consideration and contribution. However, at this time we don't currently accept PRs without an issue first laying out the problem you're trying to solve or the feature you're trying to implement. Going through this process allows us to give you feedback about the overall approach and how it ties into kube-router's overall architecture and vision.

In this particular case, I see the efficiency that you're going for here, however, in version 1.2 we're going to be completely overhauling how kube-router interacts with iptables and digging into performance issues like this at an architectural level so your PR comes a bit premature. Additionally, there are benefits and drawbacks to increasing the scope of our controller structs since they are so long lived and increase the risk for memory leaks.

PR #989, opened 2020-09-15
auto configure MTU

### LINKED ISSUE #165: Kube-router should auto-configure sensible MTU for kube-bridge instead of default one
CNI bridge plug-in can be configured explictly to set MTU for the bridge and pod interfaces connected to the bridge

https://github.com/containernetworking/plugins/tree/master/plugins/main/bridge

Kube-router does not set MTU value by defaults, so it get sets to default value of 1500.

In cases where physical interface has large MTU (for e.g 9001) then `kube-bridge` can become bottleneck as path MTU effectively reduces to 1500

murali-reddy:
set `mtu` in cni spec to auto configure MTU's of the pod's veth's and kube-bridge interfaces

simple approach to pick the MTU associated with interface associated with node's IP as MTU. This should work fine for most cases.

Please see https://github.com/containernetworking/plugins/tree/master/plugins/main/bridge#network-configuration-reference for further details on how `bridge` CNI works

Fixes #165

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -2,6 +2,7 @@ package routing
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -341,6 +342,46 @@ func (nrc *NetworkRoutingController) updateCNIConfig() {
 			glog.Fatalf("Failed to insert `subnet`(pod CIDR) into CNI conf file: %s", err.Error())
 		}
 	}
+
+	err = nrc.autoConfigureMTU()
+	if err != nil {
+		glog.Fatalf("Failed to auto-configure MTU: %s", err.Error())
+	}
+}
+
+func (nrc *NetworkRoutingController) autoConfigureMTU() error {
+	mtu, err := getMTUFromNodeIP(nrc.nodeIP)
+	file, err := ioutil.ReadFile(nrc.cniConfFile)
+	if err != nil {
+		return fmt.Errorf("Failed to load CNI conf file: %s", err.Error())
+	}
+	var config interface{}
+	err = json.Unmarshal(file, &config)
+	if err != nil {
+		return fmt.Errorf("Failed to parse JSON from CNI conf file: %s", err.Error())
+	}
+	if strings.HasSuffix(nrc.cniConfFile, ".conflist") {
+		configMap := config.(map[string]interface{})
+		for key := range configMap {
+			if key != "plugins" {
+				continue
+			}
+			pluginConfigs := configMap["plugins"].([]interface{})
+			for _, pluginConfig := range pluginConfigs {
+				pluginConfigMap := pluginConfig.(map[string]interface{})
+				pluginConfigMap["mtu"] = mtu
+			}
+		}
+	} else {
+		pluginConfig := config.(map[string]interface{})
+		pluginConfig["mtu"] = mtu
+	}
+	configJSON, _ := json.Marshal(config)
+	err = ioutil.WriteFile(nrc.cniConfFile, configJSON, 0644)
+	if err != nil {
+		return fmt.Errorf("Failed to insert `mtu` into CNI conf file: %s", err.Error())
+	}
+	return nil
 }
 
 func (nrc *NetworkRoutingController) watchBgpUpdates() {
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -351,14 +351,17 @@ func (nrc *NetworkRoutingController) updateCNIConfig() {
 
 func (nrc *NetworkRoutingController) autoConfigureMTU() error {
 	mtu, err := getMTUFromNodeIP(nrc.nodeIP)
+	if err != nil {
+		return fmt.Errorf("failed to generate MTU: %s", err.Error())
+	}
 	file, err := ioutil.ReadFile(nrc.cniConfFile)
 	if err != nil {
-		return fmt.Errorf("Failed to load CNI conf file: %s", err.Error())
+		return fmt.Errorf("failed to load CNI conf file: %s", err.Error())
 	}
 	var config interface{}
 	err = json.Unmarshal(file, &config)
 	if err != nil {
-		return fmt.Errorf("Failed to parse JSON from CNI conf file: %s", err.Error())
+		return fmt.Errorf("failed to parse JSON from CNI conf file: %s", err.Error())
 	}
 	if strings.HasSuffix(nrc.cniConfFile, ".conflist") {
 		configMap := config.(map[string]interface{})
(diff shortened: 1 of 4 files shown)


pkg/controllers/routing/utils.go
```diff
@@ -120,6 +120,26 @@ func getNodeSubnet(nodeIP net.IP) (net.IPNet, string, error) {
...
+		return 0, errors.New("Failed to get list of links")
+	}
+	for _, link := range links {
+		addresses, err := netlink.AddrList(link, netlink.FAMILY_ALL)
+		if err != nil {
+			return 0, errors.New("Failed to get list of addr")
+		}
+		for _, addr := range addresses {
+			if addr.IPNet.IP.Equal(nodeIP) {
+				lintMTU := link.Attrs().MTU
```
aauren, line 135: Should probably be `linkMTU`

pkg/controllers/routing/utils.go
```diff
@@ -120,6 +120,26 @@ func getNodeSubnet(nodeIP net.IP) (net.IPNet, string, error) {
...
+				return lintMTU - 20, nil // -20 to accommodate IPIP header
```
aauren, line 136: You should do testing here to see if the user is running an overlay before automatically reducing the linkMTU by 20 bytes. We don't run our nodes in an overlay network and the 20 byte difference here a .26 Gbps difference in our performance testing when testing container to container where jumbo frames were enabled on both nodes.

In my testing, I just passed a boolean to `getMTUFromNodeIP` that I sourced from `nrc.enableOverlays`. I don't know if `nrc.enableOverlays` takes into account all possible IPIP overlay scenarios or not without going over `injectRoute` in more fine grained detail.

aauren: My mocked up `getMTUFromNodeIP` looked like:
```
func getMTUFromNodeIP(nodeIP net.IP, overlayEnabled bool) (int, error) {
	links, err := netlink.LinkList()
	if err != nil {
		return 0, errors.New("Failed to get list of links")
	}
	for _, link := range links {
		addresses, err := netlink.AddrList(link, netlink.FAMILY_ALL)
		if err != nil {
			return 0, errors.New("Failed to get list of addr")
		}
		for _, addr := range addresses {
			if addr.IPNet.IP.Equal(nodeIP) {
				lintMTU := link.Attrs().MTU
				if overlayEnabled {
					return lintMTU - 20, nil // -20 to accommodate IPIP header
				}
				return lintMTU, nil
			}
		}
	}
	return 0, errors.New("Failed to find interface with specified node ip")
}
```

And then on the calling side, I had something like:
```
func (nrc *NetworkRoutingController) autoConfigureMTU() error {
	mtu, err := getMTUFromNodeIP(nrc.nodeIP, nrc.enableOverlays)
	if err != nil {
		return fmt.Errorf("failed to generate MTU: %s", err.Error())
	}
```

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -341,6 +342,49 @@ func (nrc *NetworkRoutingController) updateCNIConfig() {
 			glog.Fatalf("Failed to insert `subnet`(pod CIDR) into CNI conf file: %s", err.Error())
 		}
 	}
+
+	err = nrc.autoConfigureMTU()
```
aauren, line 346: Configuring the MTU in the CNI only changes the MTU for the interface that is placed inside the network namespace on the container side. You still need to adjust the kube-bridge interface as well, otherwise the packet will get adjusted when exiting or entering the bridge interface on the host side.

For this, I made the following adjustment to network_routes_controller.Run():
```
	// create 'kube-bridge' interface to which pods will be connected
	kubeBridgeIface, err := netlink.LinkByName("kube-bridge")
	if err != nil && err.Error() == IfaceNotFound {
		linkAttrs := netlink.NewLinkAttrs()
		linkAttrs.Name = "kube-bridge"
		bridge := &netlink.Bridge{LinkAttrs: linkAttrs}
		if err = netlink.LinkAdd(bridge); err != nil {
			glog.Errorf("Failed to create `kube-router` bridge due to %s. Will be created by CNI bridge plugin when pod is launched.", err.Error())
		}
		kubeBridgeIf, err := netlink.LinkByName("kube-bridge")
		if err != nil {
			glog.Errorf("Failed to find created `kube-router` bridge due to %s. Will be created by CNI bridge plugin when pod is launched.", err.Error())
		}
		err = netlink.LinkSetUp(kubeBridgeIf)
		if err != nil {
			glog.Errorf("Failed to bring `kube-router` bridge up due to %s. Will be created by CNI bridge plugin at later point when pod is launched.", err.Error())
		}
		kubeBridgeIface, err = netlink.LinkByName("kube-bridge")
		if err != nil {
			glog.Error("Even after attempting to bring kube-router bridge up, we still weren't able to fetch interface: %s", err.Error())
		}
	}
	mtu, err := getMTUFromNodeIP(nrc.nodeIP, nrc.enableOverlays)
	if err != nil {
		glog.Errorf("Failed to find MTU for node IP: %s for intelligently setting the kube-bridge MTU due to %s.", nrc.nodeIP, err.Error())
	}
	if mtu > 0 {
		glog.Infof("Setting MTU of kube-bridge interface to: %d", mtu)
		err = netlink.LinkSetMTU(kubeBridgeIface, mtu)
		if err != nil {
			glog.Errorf("Failed to set MTU for kube-bridge interface due to: %s", err.Error())
		}
	} else {
		glog.Infof("Not setting MTU of kube-bridge interface")
	}
```

Without this adjustment, when running iperf, I saw no actual improvements in performance of the interface.

aauren (changes requested):

aauren:
In my testing within our cluster, when configured properly, we saw an approximate improvement of ~.5 Gbit/sec on single homed 10 Gbps links over the original 1500 MTU configuration.

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -341,6 +358,51 @@ func (nrc *NetworkRoutingController) updateCNIConfig() {
 			glog.Fatalf("Failed to insert `subnet`(pod CIDR) into CNI conf file: %s", err.Error())
 		}
 	}
+
+	if nrc.autoMTU {
+		err = nrc.autoConfigureMTU()
+		if err != nil {
+			glog.Fatalf("Failed to auto-configure MTU: %s", err.Error())
```
aauren, line 365: Are we sure that we want to make this fatal here? We're defaulting MTU to enabled by default, and this will essentially terminate the `network_routes_controller` but above, when you try to set the MTU on the `kubeBridgeIf` you handle it with just an error in the logs.

We haven't been setting MTU so far. I think it would be better to just `glog.Errorf` here and warn people that we won't be able to set the MTU like we do in the `Run()` method above. That would essentially allow user's that encounter an error with this new default option to continue running kube-router the same way they have been running it and reduce the impact of this option quite a bit.

aauren (changes requested):

pkg/controllers/routing/network_routes_controller.go
murali-reddy: Make sense to just keep it Errorf.

aauren (approved):
LGTM!

PR #990, opened 2020-09-23
go.mod: Vendor newer prometheus and k8s

mrueg:
This vendors a later version of prometheus' golang client (0.8.0 ->
0.9.4) to allow `go mod tidy`to work properly.
It also updates the k8s libraries from 0.18.6 to 0.18.8 to avoid
hitting https://github.com/kubernetes/kubernetes/issues/91073


murali-reddy:
LGTM

PR #993, opened 2020-09-29
.gitignore: Drop vendor folder

mrueg:
This was likely required for old-style vendoring and shouldn't be
necessary for the current version of kube-router

@murali-reddy


murali-reddy:
LGTM

Issue #330, opened 2018-03-07
kube-router is incompatible to cni v0.6.0 multi subnets


Thoro:
In cni 0.6.0 you can have multiple subnets for the cni config:

```
{"bridge":"kube-bridge","ipam":{"ranges":[[{"gateway":"10.253.12.1","subnet":"10.253.12.0/24"}],[{"gateway":"2a0b:37c0:1:100c::1","subnet":"2a0b:37c0:1:100c::/64"}]],"type":"host-local"},"isDefaultGateway":true,"name":"mynet","type":"bridge"}
```

Instead of editing the sub range, kube-router adds a new subnet field.

```

{"bridge":"kube-bridge","ipam":{"ranges":[[{"gateway":"10.253.12.1","subnet":"10.253.12.0/24"}],[{"gateway":"2a0b:37c0:1:100c::1","subnet":"2a0b:37c0:1:100c::/64"}]],"subnet":"10.253.12.0/24","type":"host-local"},"isDefaultGateway":true,"name":"mynet","type":"bridge"}
```

Which means that pods can't be started anymore, because that config file is invalid.

murali-reddy:
@Thoro thanks for reporting this issue. This is a concerning. `v0.6.0` is around for a quite a while, wonder why this issue did not surface so far.

Thoro:
probably because nobody was so stupid and configure ipv6 directly in the ipam module :D

Thoro:
suggestion from my side:

upgrade the cni module and remove the writing of the subnet into the file - or make it optional via flag

aauren:
Closing as duplicate of #1023

PR #996, opened 2020-10-03
Update dependencies (go-iptables, cni)

mrueg:
Since 1.1 is released, let's update dependencies for cni  and go-iptables.


murali-reddy:
Please see https://github.com/cloudnativelabs/kube-router/issues/330

From https://www.cni.dev/plugins/ipam/host-local/

`Previous versions of the host-local allocator did not support the ranges property, and instead expected a single range on the top level. This is deprecated but still supported.`

after updating libcni to v0.8.0 from my testing it is still working. But we need a code fix to properly populate ranges in `/etc/cni/net.d/10-kuberouter.conflist`

mrueg:
Yes, this is why I added the TODO here: https://github.com/cloudnativelabs/kube-router/pull/996/files#diff-3e452abac1fb0021ce1c92d3aea0d98f7e52635012288106c06756d1ecb7d53fR54

murali-reddy:
LGTM. I will raise a PR to ensure kube-router support ranges.

PR #998, opened 2020-10-09
Cleanup non-DSR externalIPs

philschleier:
We had some stale IPs in table 79 that gave me quite the headache.

Currently, any `externalIP` is added to the `activeExternalIPs` map, while only DSR `externalIP`s are added to the routing table. Only routes for deleted `externalIP`s will be removed from the table afterwards. `externalIP`s that no longer carry the DSR annotation will remain.

This PR updates the `activeExternalIPs` map to carry only those IPs that we are also actively enforcing routes for.

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -1946,12 +1946,13 @@ func (ln *linuxNetworking) setupRoutesForExternalIPForDSR(serviceInfoMap service
 	activeExternalIPs := make(map[string]bool)
 	for _, svc := range serviceInfoMap {
 		for _, externalIP := range svc.externalIPs {
-			activeExternalIPs[externalIP] = true
 			// Verify the DSR annotation exists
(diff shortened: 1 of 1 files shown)


aauren (approved):
Thanks for finding and submitting a PR for this. While I haven't tested this directly, the code and the intention seem correct to me.

Just to make sure I understand correctly, I'll try to rephrase your original description, let me know if I get it correct:

Services with a DSR annotation will have a route added to table 79 when they are first created. From there, if they are updated to no longer have the DSR annotation, as opposed to being deleted outright, they will still be added to the `activeExternalIPs` hash which will result in the route not being removed from table 79.

This change moves adding `externalIP` to the `activeExternalIPs` hash after the evaluation of whether or not this service still has the DSR annotation so that the route from table 79 can be properly removed.

philschleier:
Exactly.

aauren:
Merging this since it's been a week now with no other reviews.

PR #999, opened 2020-10-21
use custom pod informer to reduce the memory footprint in large scale clusters

murali-reddy:
Adds a custom pod informer to reduce the memory footprint of storing full detailed pod objects returned by client-go pod informer. stripped down pod object with just enough details needed for kube-router is store in the cache

In fairly large cluster with > 800 nodes and with 15K pods, it is observed that client-go's [pod informer](https://github.com/kubernetes/client-go/blob/master/informers/core/v1/pod.go) takes about > 200MB just to store the pod objects in local cache. Also there seems to be unbounded memory growth as reported in https://github.com/kubernetes/client-go/issues/871

Also see https://github.com/kubernetes/client-go/issues/832, there is metadat informer but kube-router needs additional details of the pod from pod's `Status` so can not be used.

### DIFF
--- a/pkg/cmd/kube-router.go
+++ b/pkg/cmd/kube-router.go
@@ -14,6 +14,7 @@ import (
 	"github.com/cloudnativelabs/kube-router/pkg/metrics"
 	"github.com/cloudnativelabs/kube-router/pkg/options"
 	"github.com/cloudnativelabs/kube-router/pkg/version"
+	"github.com/cloudnativelabs/kube-router/pkg/utils"
 	"k8s.io/klog/v2"
 
 	"time"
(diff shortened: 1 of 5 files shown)


aauren:
@murali-reddy I rebased this PR and updated it to remove a call to a deprecated function.

If you think that this PR is still worthwhile, maybe we should finish out your last bit of testing and then merge into master now that we've cut 1.2.X it would be good to get this into the main line.

aauren:
Closing for now, we can re-open if this functionality is still desired.

However, for most clusters that kube-router is used in, this is probably not worth the complexity that introducing our own informer introduces.

PR #1002, opened 2020-11-09
explicit ACCEPT traffic that complies with network policies

### LINKED ISSUE #984: Unable to detect if pod passed network policies
**What happened?**
Before 1.0.0-rc4, network traffic was "accepted" if network policies were matched. Now, they just return. This makes it very difficult to differentiate between whether the network policies were matched and passed or were unmatched.

Traffic to be accepted if it made it through the network policy or for a way to detect that it matched a network policy

1. Set iptables forward policy to drop
2. Try and figure out how to unmatched traffic (such as host traffic and traffic that didn't match a network policy)

None

 - Kube-Router Version (`kube-router --version`): 1.0.1
 - Kube-Router Parameters: 
```
    - --hostname-override=172.28.0.3
    - --kubeconfig=/etc/kubernetes/kube-proxy-kubeconfig
    - --run-router=false
    - --run-firewall=true
    - --run-service-proxy=true
    - --service-cluster-ip-range=172.30.128.0/17
    - --metrics-port=20249
```
 - Kubernetes Version (`kubectl version`) : 1.18.6
 - Cloud Type: On premise
 - Kubernetes Deployment Type: Custom
 - Kube-Router Deployment Type: static pod
 - Cluster Size: 9 nodes

None

**Additional context**

It would be nice if there was a flag to either go back to the old behaviour, have a post-processing table name, or a way to set a mark if the network policies were matched.

murali-reddy:
Fixes #984

make network policy enforcement by kube-router be authoritative
entity to ACCEPT traffic from/to the pods

ensure there is pod firewall chain irrespective of any network policy
applicable to the pods

default behaviour is to ACCPET the traffic from the pods if no network
policies applicable, else run through the newrok policy chains

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -304,6 +304,16 @@ func (npc *NetworkPolicyController) ensureTopLevelChains() {
 		ensureRuleAtPosition(kubeInputChainName, whitelistServiceVips, uuid, externalIPIndex+4)
 	}
 
+	// for the traffic to/from the local pod's let network policy controller be
+	// authoritative entity to ACCEPT the traffic if it complies to network policies
+	for _, chain := range chains {
+		comment := "rule to explictly ACCEPT traffic that comply to network policies"
+		args := []string{"-m", "comment", "--comment", comment, "-m", "mark", "--mark", "0x20000/0x20000", "-j", "ACCEPT"}
+		err = iptablesCmdHandler.AppendUnique("filter", chain, args...)
+		if err != nil {
+			glog.Fatalf("Failed to run iptables command: %s", err.Error())
+		}
+	}
 }
 
 // OnPodUpdate handles updates to pods from the Kubernetes api server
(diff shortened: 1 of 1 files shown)


pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -1069,13 +993,82 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(networkPoliciesInfo []
 				return nil, fmt.Errorf("Failed to run iptables command: %s", err.Error())
 			}
 		}
+	}
+
+	// loop through the pods running on the node which to which ingress network policies to be applied
+	ingressNetworkPolicyEnabledPods, err := npc.getIngressNetworkPolicyEnabledPods(networkPoliciesInfo, npc.nodeIP.String())
```
aauren, line 999: My understanding is that there is some benefit to having `Established,Related` traffic which accounts for the bulk of traffic and is an easy determination for iptables at the top of the chain. If I were you, I would move the Ingress and Egress chain handling near the top of this function so that when you do the calls to `Insert` the `--ctstate RELATED,ESTABLISHED` at position 1, it will come in above these rules and get processed in the chain first and ACCEPTED before iptables has to made determinations about these chains.

aauren (commented):
I think that there are some fairly easy optimizations that can be had here, but overall looks good.

aauren:
One other thing I just thought of, we should also change these global accept rules from the `network_routes_controller` too right? https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/routing/network_routes_controller.go#L735

murali-reddy:
Thanks for review @aauren 

I am closing this in favour of #1009. Will address your commetns in it.

PR #1009, opened 2020-12-01
streamline how traffic to/from pod is processed through KUBE-POD-FW chains 

murali-reddy:
- make NPC authoritative entity to ACCEPT pod traffic
- also with this change each pod on the node will get a pod firewall chain irrespective of the fact its has ingress or egress network policy that applies to it
- also introduces two new chains to represent default ingress and egress network policies in case when pod has no matching ingress/egress policies this network policy just accepts traffic


murali-reddy:
Closing this PR as #1010 is inclusive of the changes in the PR

Issue #997, opened 2020-10-06
refactor controllers to do partial reconcilation on watch events 


murali-reddy:
**TLDR**: kube-router leverages Kubernetes controller pattern and watches API objects and performs reconcilation (node level changes to reflect desired state). In the current design of kube-router it does full reconcilation (force-refresh the whole state), rather than just doing incremental delta configuration update corresponding to watch API object event. This approach simplifies the design but can result in scalability issues. This proposal is to refactor current controllers to restrict the processing done on watch API object event to just scope of the change needed rather full reconcilation of desired state.

**Problem**:

kube-router is primarily consists of below three contollers

- `network policy controller` which provides an implementation of network policy and watches following Kubernetes resources
    - Pod
    - Namespace
    - NetworkPolicy
- `network service controller` which provides an implementation of service proxy and watches following Kubernetes resources
    - Service
    - Endpoint
- `network routing controller` which provides pod-to-pod connectivity also advertising the service VIP's to provide network load balancer and watches following Kubernetes resources
    - Node
    - Service
    - Endpoint

each of these controllers registers a callback to Add/Delete/Update events. At present kube-router follows simple model in which each controller performs full reconcilation irrespective of the resource object for which event was received. To safeguard against cases where there is flood of events for e.g. on node restart, deleting a deployment with large number of pods etc, kube-router attempts to coalesce the events and does the minimum syncs. This approach works faily well for most deployments. So far users feedback over last few years indicated below two main problems:

- full sync time can be unacceptle. In some of the deployments for instance network policy controller full sync is found to be in the order of couple of minutes
- if there is constant churn of resources (pods, services etc) then even with coalescing kube-router can end up continuously processing the events. Combined with longer full sync's kube-router can potentially hog the system

In full syncs if it only involves system calls its pretty fast. For e.g. `network service controller` largley invokes netlink, ipvs system calls. But inherently still full sync can not scale as number of services and endpoints increases. When there is a necessasity to call external command (resulting in process context switch) like in case of `network policy controller` which relies on `iptables` command performing full sync can be expensive. As the number of network policies and corresponding pods to which policies appy increases the number of iptables calls increases linearly.

So its important to reconsider current design to strike a balance between simplicity of design and scalability

** proposed solutions **

TODO: will be adding the details

aauren:
[...]

PR #1010, opened 2020-12-03
refactor network policy controller

murali-reddy:
In the existing implementation full sync of pod and network policy firewal chains are synced for every add/delete/update corresponding to pod, network policy and namespace object.

This PR changes how pod add/delete/update event are handled to ensure only necessary pod and network policy chains are synced and avoids full-sync

Please see #997 for the details on problem of doing full sync for the watched API object events

This PR also refactors the code and adds new functionality so as to ensure implementation achives set goals documented in the included design document in the PR
```
# Design tenets and goals:

  - full compliance to Kubernetes v1 network policy specification
  - a CNI (providing pod-to-pod networking) agnostic implementation i.e.) no assumption on any particul CNI used
  - kube-router should process the traffic from/to the pods only, and should skip any other traffic
  - any configuration changes performed should not result in accept/deny of non-pod traffic  
  - provide a stateful firewall for minimal overhead in enforcing network policies
  - ensure decoupled control/management plane from datapath so that kube-router pod restarts/crash etc does not affect data-path
  - when a new pod is launched ensure network policy enforcements are in-place before pod sends the first packet
  - be authoritative (take precedence over any other policies) in accept/deny traffic concerning the traffic from/to the pods
  - provide insight in to dropped traffic due to network policy enforcement
  - tolerant to out-of-band changes and reconcile desired state
  - fully stateless design should scale to cluster of any size with any number of network policies

```


pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -129,7 +129,7 @@ type numericPort2eps map[string]*endPoints
 type protocol2eps map[string]numericPort2eps
 type namedPort2eps map[string]protocol2eps
 
-// Run runs forver till we receive notification on stopCh
+// Run runs forver till we receive notification on stopCh to shutdown
```
mrueg, line 132: forever*

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -400,15 +400,18 @@ func (npc *NetworkPolicyController) ensureTopLevelChains() {
 		ensureRuleAtPosition(kubeInputChainName, whitelistServiceVips, uuid, externalIPIndex+4)
 	}
 
-	// for the traffic to/from the local pod's let network policy controller be
-	// authoritative entity to ACCEPT the traffic if it complies to network policies
 	for _, chain := range chains {
+		// for the traffic to/from the local pod's let network policy controller be
```
mrueg, line 404: pod's -> pods

pkg/controllers/netpol/pod.go
```diff
@@ -74,18 +74,20 @@ func (npc *NetworkPolicyController) processPodAddUpdateEvents(pod *api.Pod) {
 
 	err = npc.syncAffectedNetworkPolicyChains(&podInfo, syncVersion)
 	if err != nil {
-		glog.Errorf("failed to refresh network policy chains affected by pod %s event due to %s", podNamespacedName, err.Error())
+		glog.Errorf("failed to refresh network policy chains affected by pod:%s event due to %s", podNamespacedName, err.Error())
```
mrueg, line 77: failed -> Failed (start with capital letter)

murali-reddy:
Have performed following test scenarios.

```
- when no network policy is applied to a pod, pod should be able to send and receive traffic

- when a network policy applied to a pod, pod should be able to send egress traffic:
    - only to the pods in the same namespace selected by pod selector in case when no namespace selector specified in NetworkPolicyPeer
    - to all the pods in all namespaces selected by namespace selector if only namespace selector is specified in NetworkPolicyPeer
    - to all the pods selected by pod selector in all namespaces selected by namespace selector if both namespace and pod selector are given
    - to specified ip block in NetworkPolicyPeer

- when a network policy applied to a pod, pod should not be able to egress traffic to any destination not selected by the NetworkPolicyPeer 

- when a network policy applied to a pod, pod should be able to receive ingress traffic:
    - only from the pods in the same namespace selected by pod selector in case when no namespace selector specified in NetworkPolicyPeer
    - from all the pods in all namespaces selected by namespace selector if only namespace selector is specified in NetworkPolicyPeer
    - from all the pods selected by pod selector in all namespaces selected by namespace selector if both namespace and pod selector are given
    - from specified ip block in NetworkPolicyPeer

- when a network policy applied to a pod, pod should not be able to receive ingress traffic from any source not selected by the NetworkPolicyPeer

- when a pod is spun up before network policies are enforced, during that window of time pod should be able to send/receive traffic if `--netpol-allow-pre-check` is true

- when a pod is spun up before network policies are enforced, during that window of time pod should be NOT able to send/receive traffic if `--netpol-allow-pre-check` is false

- any pod irrespective of network policies applied should be able to access all Kubernetes services through cluster IP, node port, external IP

- updating pod labels should result in enforcing network policies as per the new labels

- restart of kube-router pods should not result in any network policy enforcement

- whitelisted traffic must be allowed irrespective of source pod and destination pod are on the same node or different nodes

- non-whitelisted traffic must be  not be allowed irrespective of source pod and destination pod are on the same node or different nodes

- packets dropped by network policy enforcement must be logged

- any out-of-band changes like flushing the whole table, chains etc should eventually be reconciled with desired state
```

If any one would like to help with testing please use `cloudnativelabs/kube-router-git:amd64-pod-fw-sync`

mhumpula:
I've tested it with few of our patches on stage environment, and it seems working nicely. Noticed few things:
* the full sync of iptables went from ~3s to ~30s (measured via kube_router_controller_iptables_sync_time_bucket metric), i guess this is a result of adding more iptables command invocations.
* I've added quick metric for measuring pod changes fw speed and as expected it is below 300ms (mostly in order of 50ms).

pkg/controllers/netpol/pod.go
```diff
@@ -53,12 +105,64 @@ func (npc *NetworkPolicyController) handlePodDelete(obj interface{}) {
...
+	if err != nil {
+		glog.Errorf("failed to refresh network policy chains affected by pod %s/%s delete event due to %s", pod.Namespace, pod.Name, err.Error())
+	}
+
+	// cleanup of firewall chains needed only for local pods
+	if !isLocalPod(pod, npc.nodeIP.String()) {
+		return
+	}
+
+	npc.mu.Lock()
```
aauren, line 130: How come you don't grab the lock here before you call `npc.syncAffectedNetworkPolicyChains` above? You hold the lock before that statement in `processPodAddUpdateEvents()`

pkg/controllers/netpol/policy.go
```diff
@@ -79,72 +88,172 @@ func (npc *NetworkPolicyController) syncNetworkPolicyChains(networkPoliciesInfo
...
+		}
+		if policy.ObjectMeta.Namespace == pod.namespace && targetPodSelector.Matches(labels.Set(pod.labels)) {
+			matchesTargetPod = true
+		}
+		for _, specIngressRule := range policy.Spec.Ingress {
+			if len(specIngressRule.From) == 0 {
+				continue
+			}
+			for _, peer := range specIngressRule.From {
+				var matchesPodSelector, matchesNamespaceSelector bool
```
aauren, line 186: This doesn't take into account CIDR network policies which will also have nil PodSelector or NameSpaceSelectors.

Something like the following needs to be added to Ingress and Egress rule processing so as to not try to unnecessarily process CIDR policy changes
```
				if peer.IPBlock != nil {
					continue
				}
```

aauren:
@murali-reddy I got a chance to test these changes in one of our development environments and here is what I found.

This code change does make a difference, however, it seems to be very much dependent on how well defined your network policy is. If you have a very targeted network policy, all with specific pod selectors, then this change does indeed make a large difference. If you have network policy that is less well defined and that frequently selects entire namespaces, then this change has less of an impact. In one of our environments, I found that current full syncs were taking about 1 minute to perform, however, because of the way many of our network policies are defined, the change was anti-climatic, only going down to about 45 seconds.

Additionally, because we are no longer just relying on full syncs, and are instead pushing more partial syncs, the code in this PR is no longer coalescing them into a single run. As such, I think that this change, if deployed into the environment that I tested, would be dangerous. If it takes 45 seconds for it to clear a partial sync on pod update, and we have pod updates happening with a frequency greater than 45 seconds, the queued partial updates will begin to queue up.

The queuing there will begin to inflate the memory that kube-router requires because of the regression in the coalescence logic that was added here: https://github.com/cloudnativelabs/kube-router/pull/902 (this would especially be the case while we're still holding the full pod that we get from the default informer, i.e. without: https://github.com/cloudnativelabs/kube-router/pull/999)

Additionally, as I don't see any logic here that purges the update queue upon a successful full sync, you could end up in the situation where you have:
* Just run a full sync
* But still have a lot of partial updates queued, some potentially more than an hour old

When you finish the full sync, and move on to process an old partial update queue, you may end up reverting some of the proper changes that your full sync just finished processing based upon an older state.

Anyway, I think that before we merge this PR we need to somehow find a way to coalesce the update queue before executing on it, so that if the updates pull in a lot of iptables changes which cause the command to take longer, we don't run into memory issues or cause the pod to wait too long for an updated network policy on its node to be applied and go into a crash loop or some such.

murali-reddy:
Thanks @aauren for thorough review and feedback. I have incorporated your feedback.

I optimized the effort involved in processing add/delete/update pod event. We just need to refresh the relevent ipset's to include/remove the pod IP corresponding to the event. Refreshing the network policy chain rules are not done as its not necessary. I think what we have now with https://github.com/cloudnativelabs/kube-router/pull/1010/commits/c1eb5a729141cd4b1b113713325c33c7c70ce118 is bare-minum effort needed to process pod events.

> Additionally, because we are no longer just relying on full syncs, and are instead pushing more partial syncs, the code in this PR is no longer coalescing them into a single run. As such, I think that this change, if deployed into the environment that I tested, would be dangerous. If it takes 45 seconds for it to clear a partial sync on pod update, and we have pod updates happening with a frequency greater than 45 seconds, the queued partial updates will begin to queue up.

I though about coalescing one or more events and replace with full sync, but I think then it defeats one of the goal of refactoring to ensure network policies applied at the earliest after the pod starting. I have not addressed this conecent raised by you yet. We need to balance both the cases.

aauren:
Closing this as we are no longer proceeding with partial syncs after learnings on level-triggered and edge-triggered architecture.

PR #1015, opened 2020-12-15
NPC design document

murali-reddy:


murali-reddy:
moved design document to #1010

PR #1016, opened 2020-12-19
Update RBAC apiVersion from v1beta1 to v1

lcavajani:
rbac.authorization.k8s.io/v1 is stable since 1.8, updating
it avoids warnings when applying the manifests on recent versions
of Kubernetes (v1.19+).

Signed-off-by: lcavajani <lcavajani@suse.com>


aauren (approved):
Thanks for the submission!

mrueg (approved):

PR #1018, opened 2020-12-30
Makefile: Drop outdated glide command

mrueg:
since the project moved to go modules, glide is not used anymore.


murali-reddy:
LGTM

Issue #843, opened 2020-02-17
DSR support for containerd


andrewrynhard:
I'd like to use DSR with containerd instead of Docker.

aauren:
Maybe, I'm missing something here, but I don't believe that there is anything specifically tied to docker about our current DSR implementation. Can you explain more about what specifically doesn't work about DSR with containerd that does work with Docker?

andrewrynhard:
@aauren The docs say this:

```
Also host path /var/run/docker.sock must be made a volumemount to kube-router.
```

aauren:
Thanks for the reply. I had missed that in the documentation. Looking through the code, it appears that the kube-router code is pretty strongly tied to a docker runtime for handling DSR so it would probably be a bit of work to support other runtimes directly like containerd or cri-o.

Being realistic, I don't know that this is something that we'll have time to get to in the next couple of phases. Is this something that you would be interested in submitting and implementation for?

andrewrynhard:
@aauren As much as I would :heart to, unfortunately no. I have quite a few open source projects I am juggling, and we have since learned to live without DSR. If I get a chance, I might take a shot at it, but I can't say I can commit to this right now. Maybe in the future.

aauren:
@andrewrynhard Thanks for getting back about this. As much as it would be awesome to be able to support multiple container runtimes, we're currently focused on enhancing other areas of kube-router so I'm going to close this for now.

If someone happens across this in the future and wants to contribute the functionality just respond here, totally open to re-opening it and discussing if someone has the time to contribute this functionality.

murali-reddy:
Current implementation support docker only. There is low level plumbing required to support DSR

https://github.com/cloudnativelabs/kube-router/blob/v1.0.0/pkg/controllers/proxy/network_services_controller.go#L869-L874

which requires to use container runtime API. If there is any library that can abstract the container runtimes it would be nice to do it once for all.

sfxworks:
It should be noted that support for docker on k8s is going away https://github.com/kubernetes/kubernetes/pull/94624
All communication will have be CRI compatible so support for containerd crio and others will be wanted

aauren:
Re-opening now that @ep4eg has put in a PR

sfxworks:
Noted the merge. I'll test crio and containerd shortly.

yydzhou:
> Noted the merge. I'll test crio and containerd shortly.

@sfxworks any update on this?

sfxworks:
[...]

PR #1027, opened 2021-01-30
feat: simple CRI implementation in addition to Docker, required for DSR functionality. CRI compliant runtimes support (e.g. containerd, cri-o, etc.)

ep4eg:
Currently, DSR functionality strictly relies on Docker, so it doesn't work with other container runtimes. This PR implements a simple CRI interface in order to support popular CRI compliant runtimes such as containerd, crio-o, etc. in addition to docker, which is about to be deprecated.

related issue: https://github.com/cloudnativelabs/kube-router/issues/843


go.mod
```diff
@@ -36,11 +36,13 @@ require (
 	github.com/vishvananda/netlink v1.1.1-0.20201029203352-d40f9887b852
 	github.com/vishvananda/netns v0.0.0-20200728191858-db3c7e526aae
 	golang.org/x/net v0.0.0-20201006153459-a7d1128ccaa0
+	google.golang.org/grpc v1.27.1
 	gopkg.in/ini.v1 v1.61.0 // indirect
 	gotest.tools v2.2.0+incompatible // indirect
 	k8s.io/api v0.18.8
 	k8s.io/apimachinery v0.18.8
 	k8s.io/client-go v0.18.8
+	k8s.io/cri-api v0.20.2
```
mrueg, line 45: Can we keep the k8s dependencies in sync and use 0.18 (or update all to 0.20.x)?

ep4eg: my bad. Fixed.

go.mod
```diff
@@ -36,11 +36,13 @@ require (
 	github.com/vishvananda/netlink v1.1.1-0.20201029203352-d40f9887b852
 	github.com/vishvananda/netns v0.0.0-20200728191858-db3c7e526aae
 	golang.org/x/net v0.0.0-20201006153459-a7d1128ccaa0
+	google.golang.org/grpc v1.27.1
 	gopkg.in/ini.v1 v1.61.0 // indirect
 	gotest.tools v2.2.0+incompatible // indirect
 	k8s.io/api v0.18.8
 	k8s.io/apimachinery v0.18.8
 	k8s.io/client-go v0.18.8
+	k8s.io/cri-api v0.18.8
```
aauren, line 45: I believe that these changes require changes to go.sum as well. When I build this, I see a bunch of working directory changes to that file. Can you add those to your PR?

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -1085,6 +1092,223 @@ func (ln *linuxNetworking) prepareEndpointForDsr(containerID string, endpointIP
 	return nil
 }
 
+// The same as prepareEndpointForDsr but using CRI instead of docker
```
aauren, line 1095: We should add a comment to `prepareEndpointForDSR`  with this PR saying that it is deprecated and to evaluate removing the previous functionality when Kubernetes 1.23 becomes a prominent release.

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -1085,6 +1092,223 @@ func (ln *linuxNetworking) prepareEndpointForDsr(containerID string, endpointIP
...
+
+	var activeNetworkNamespaceHandle netns.NsHandle
+
+	hostNetworkNamespaceHandle, err := netns.Get()
+	if err != nil {
+		return errors.New("failed to get host namespace due to " + err.Error())
+	}
+	defer hostNetworkNamespaceHandle.Close()
+
+	activeNetworkNamespaceHandle, err = netns.Get()
```
aauren, line 1119: @ep4eg I know that you just copied this from the original function, but do you happen to know why it is necessary to try fetching the host namespace a second time early in the function as opposed to just fetching it once down below after you identify the containers network namespace?

Have you tested this implementation without this and seen it fail? It would be nice to remove un-needed logic if possible.

pkg/controllers/proxy/service_endpoints_sync.go
```diff
@@ -396,15 +397,30 @@ func (nsc *NetworkServicesController) setupExternalIPServices(serviceInfoMap ser
...
+
 					if containerID == "" {
 						glog.Errorf("Failed to find container id for the endpoint with ip: " + endpoint.ip + " so skipping peparing endpoint for DSR")
 						continue
 					}
 
-					err = nsc.ln.prepareEndpointForDsr(containerID, endpoint.ip, externalIPService.externalIP)
-					if err != nil {
-						glog.Errorf("Failed to prepare endpoint %s to do direct server return due to %s", endpoint.ip, err.Error())
+					if runtime == "docker" {
```
aauren, line 412: I would add a note here about deprecation as well, with the same message that you put on `prepareEndpointForDSR()`

pkg/controllers/proxy/service_endpoints_sync.go
```diff
@@ -396,15 +397,30 @@ func (nsc *NetworkServicesController) setupExternalIPServices(serviceInfoMap ser
...
+						err = nsc.ln.prepareEndpointForDsr(containerID, endpoint.ip, externalIPService.externalIP)
+						if err != nil {
+							glog.Errorf("Failed to prepare endpoint %s to do direct server return due to %s", endpoint.ip, err.Error())
+						}
+					} else {
+						// We expect CRI compliant runtimes here
+						// ugly workaround, refactoryng of pkg/Proxy is required
```
aauren, line 419: nitpic: `refactoring`

pkg/controllers/proxy/service_endpoints_sync.go
```diff
@@ -396,15 +397,30 @@ func (nsc *NetworkServicesController) setupExternalIPServices(serviceInfoMap ser
...
+						err = nsc.ln.(*linuxNetworking).prepareEndpointForDsrWithCRI(nsc.dsr.runtimeEndpoint, containerID, endpoint.ip, externalIPService.externalIP)
```
aauren, line 420: I assume that you've tested this against at least one runtime to make sure that it works?

pkg/options/options.go
```diff
@@ -184,6 +185,8 @@ func (s *KubeRouterConfig) AddFlags(fs *pflag.FlagSet) {
 		"Enables Pod Networking -- Advertises and learns the routes to Pods via iBGP.")
 	fs.BoolVar(&s.RunServiceProxy, "run-service-proxy", true,
 		"Enables Service Proxy -- sets up IPVS for Kubernetes Services.")
+	fs.StringVar(&s.RuntimeEndpoint, "runtime-endpoint", "",
+		"Path to runtime socket. Used only for DSR mode")
```
aauren, line 189: I would change the wording a bit here and add what runtimes you've specifically tested.

Maybe something like:
```
Path to CRI compatible container runtime socket (used for DSR mode). Currently known working with containerd.
```
(of course assuming that containerd is what you've tested with)

pkg/cri/remote_runtime.go
```diff
@@ -0,0 +1,95 @@
...
+
+const (
+	DefaultConnectionTimeout = 15 * time.Second
+	maxMsgSize               = 1024 * 1024 * 16 // 16 MB
+)
+
+// remoteRuntimeService is a gRPC implementation of RuntimeService.
+type remoteRuntimeService struct {
+	timeout       time.Duration
+	runtimeClient runtimeapi.RuntimeServiceClient
```
aauren, line 25: Just double-checking here, I assume that there is now `close()` method or other appropriate clean-up method that we need to expose via interface for this guy?

aauren (changes requested):
Just a few minor changes. On the whole I think that this looks very good. I didn't expect it to be this easy, and making this an alternative logic path / re-using a lot of the logic from the original function should make this an easy merge.

Thanks for your work here @ep4eg!

go.mod
ep4eg: done

pkg/controllers/proxy/network_services_controller.go
ep4eg: done

pkg/controllers/proxy/network_services_controller.go
ep4eg: Well, this is a complicated topic, as the entire method is kinda mess with a bunch of duplicated code that  I'd break into chunks tbh. Anywho, I didn't find any reason to keep this part, so I just removed that, tests passed, everything's working.

pkg/controllers/proxy/service_endpoints_sync.go
ep4eg: done

pkg/controllers/proxy/service_endpoints_sync.go
ep4eg: I've been testing this against containerd for a few days and it's working so far.

pkg/cri/remote_runtime.go
ep4eg: `RuntimeServiceClient` itself doesn't provide such a method, however, I think you're right, so I implemented `Close()` method for `RuntimeService` interface that cleanup stale gRPC connections.

aauren (changes requested):
One final change, we shouldn't need to vendor anything any longer, so remove those changes from your commits, and we should be good to go.

I'll give @murali-reddy a day or two to respond and add any comments if he wants to, and if not, I'll check tests one more time and merge.

ep4eg:
> One final change, we shouldn't need to vendor anything any longer, so remove those changes from your commits, and we should be good to go.
> 
> I'll give @murali-reddy a day or two to respond and add any comments if he wants to, and if not, I'll check tests one more time and merge.

Here we go, no more vendoring. Thank you for the review @aauren!

docs/user-guide.md
```diff
@@ -85,6 +85,7 @@ Usage of kube-router:
       --run-firewall                                  Enables Network Policy -- sets up iptables to provide ingress firewall for pods. (default true)
       --run-router                                    Enables Pod Networking -- Advertises and learns the routes to Pods via iBGP. (default true)
       --run-service-proxy                             Enables Service Proxy -- sets up IPVS for Kubernetes Services. (default true)
+      --runtime-endpoint string                       Path to CRI compatible container runtime socket (used for DSR mode). Currently known working with containerd.
```
murali-reddy, line 88: @ep4eg sorry for the late review. I was wondering cant we just keep this generic like `cri-runtime`, and then we can infer runtime socket path etc internally? Or even avoid it totally if `podObj.Status.ContainerStatuses[0].ContainerID` can give us the details?

ep4eg: Well, we can hardcode the default paths and select one based on what we get from `podObj.Status.ContainerStatuses[0].ContainerID`, however in this case we can't handle custom paths and runtimes other than containerd and cri-o. In the current design, we don't care what runtime is used, we just rely on CRI.

murali-reddy: So we want to avoid adding additional flags unless abolutley necessary, particualry something that can inferred or hardcoded.

>  In the current design, we don't care what runtime is used, we just rely on CRI.

My question is cant we infer? Also if we have hybrid CRI nodes across the clusters, then running kube-router as daemonset would be problem?

aauren: Overall I would agree with having less flags if we can help it. I know one of the biggest things that drove us to use kube-router after evaluating it was how simple it made setup as compared to Calico and some of the other options out there. We should try to keep configuring it as simple as possible if we can.

Right now Kubernetes only officially supports Docker, CRI-O, and conainerd (https://kubernetes.io/docs/setup/production-environment/container-runtimes/). Docker support is already handled. If there are reliable defaults for the socket paths for CRI-O and containerd I would say we should just hard code them and remove the flag. If, in the future, a bunch of container runtimes come on to the scene and Kubernetes wants to support them, then we can add back the socket path at that point.

ep4eg: I can't see any entropy in adding the flag but ok if you think it's better to hardcode, let's hardcode.

axeal: I could see some logic in having hardcoded defaults, but not in removing the ability to configure the CRI runtime socket path here. The socket path is an implementation detail dependent on how the cluster was configured, and whilst there are common paths used for both containerd and CRI-O, these are not guaranteed.

murali-reddy: Ok, i see your point @axeal  I dont fully understand the context for `--container-runtime-endpoint` option of kubelet. Are you saying one can use containerd/CRI-O but still endpoint can be still different than what is hardcoded due to this option?

Also as I raised earlier if some one uses different run times in the clusters, specifiying it in daemonset does not make sense.

axeal: @murali-reddy Yup, so `/run/containerd/containerd.sock` and `/var/run/crio/crio.sock` are common 'defaults' for the CRI runtime socket paths for containerd and cri-o respectively. However, these locations are configureable and can vary cluster to cluster, based on how the user/tooling set up the runtime. This PR currently sets the containerd default to `/var/run/containerd/containerd.sock`, which I assume was the path in @ep4eg's test environment and I subscribed to this PR as I'm interested in using DSR in a k3s cluster, for which the containerd socket path is `/run/k3s/containerd/containerd.sock`. Understand the point re. mixed runtimes, but that's an edge-case which I think is probably outside scope of the PR here. From what I know I don't think it's possible to infer that either, as the CRI socke path is a configuration detail of the kubelet and not exposed in a Kubernetes resource.

murali-reddy: @axeal thanks for the explanation. If there is a way we would have liked to avoid extra flag, but looks like not much choice. @ep4eg apologies for the back and forth requests, please revert back to your last commit to keep the flag

ep4eg: done

murali-reddy:
thanks @ep4eg LGTM

murali-reddy:
@axeal If you get a chance please try the `master`

PR #106, opened 2017-08-06
Use IP-IP tunnelling for cross-subnet pod connectivity

murali-reddy:
- create a ipip tunnel interface from each node to nodes in the different subnet
- insert route to local routing table using tunnel device

Fixes #92

PR #1029, opened 2021-02-04
It appears as though this line is no longer relevant

marcriddle:
As a result of the work added in https://github.com/cloudnativelabs/kube-router/pull/106, nodes that are not layer 2 adjacent will use IPIP tunnels to communicate to each other. While testing a cluster spanning multiple networks, I found that things worked properly, but was concerned about this line in the docs. Looking at git history, this line predates kube-router's ability to run across multiple networks using IPIP tunnels. I suspect this line is no longer correct, and should be removed from the docs.


murali-reddy:
Thanks for the PR. Make sense to remove that statement.

PR #1031, opened 2021-02-11
metrics: Add build_info metrics

mrueg:
This should add build_info metrics that can be consumed by prometheus to identify which version of kube-router is running.


murali-reddy:
@mrueg Changes look good. But was wondering what could be use-case where build info as prom metric can be useful. That to from daemonset?

mrueg:
> @mrueg Changes look good. But was wondering what could be use-case where build info as prom metric can be useful. That to from daemonset?
It makes it easy to visualize the kube-router version in grafana through prometheus data source.

PR #1039, opened 2021-02-25
use iptables-save and iptables-restore for npc sync

murali-reddy:
replace all individual iptables command that are run during full network policies sync into `iptables-save` and `iptables-restore`

This PR is intend to replace #1010

In #1010 I tried to refactor the code so that on pod add/delete/update kube-router performs iptables, ipset updates that are only constrained to corresponding pod. However after spendig time refactoring code it truned out it is not right approach.

In general kube-router should stick to `level-triggered` sync (Please see [1] [2]) i.e.) consider whole current state and desired state as a whole. 

Please use the image `cloudnativelabs/kube-router-git:amd64-iptables-save-restore` if any testing.

[1] https://speakerdeck.com/thockin/kubernetes-controllers-are-they-loops-or-events?slide=38
[2] https://speakerdeck.com/thockin/edge-vs-level-triggered-logic

### DIFF
--- a/pkg/controllers/netpol/pod.go
+++ b/pkg/controllers/netpol/pod.go
@@ -64,6 +64,10 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(networkPoliciesInfo []
 		// add rule to log the packets that will be dropped due to network policy enforcement
 		comment := "\"rule to log dropped traffic POD name:" + podName + " namespace: " + podNamespace + "\""
 		args := []string{"-A", podFwChainName, "-m", "comment", "--comment", comment, "-m", "mark", "!", "--mark", "0x10000/0x10000", "-j", "NFLOG", "--nflog-group", "100", "-m", "limit", "--limit", "10/minute", "--limit-burst", "10", "\n"}
+		// This used to be AppendUnique when we were using iptables directly, this checks to make sure we didn't drop unmarked for this chain already
+		if strings.Contains(npc.filterTableRules.String(), strings.Join(args, " ")) {
+			return nil
+		}
 		npc.filterTableRules.WriteString(strings.Join(args, " "))
 
 		// add rule to DROP if no applicable network policy permits the traffic
--- a/pkg/controllers/netpol/pod.go
+++ b/pkg/controllers/netpol/pod.go
@@ -3,10 +3,8 @@ package netpol
 import (
 	"crypto/sha256"
 	"encoding/base32"
-	"fmt"
 	"strings"
 
-	"github.com/coreos/go-iptables/iptables"
 	"github.com/golang/glog"
 	api "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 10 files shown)


NaPs:
Hi @murali-reddy,

We tested it today on a development cluster with ~10 nodes, ~300 pods, ~60 namespaces and ~60 netpols. After deploying this image, iptable sync time went from ~30-60s to < 1s. Yay!

image
image

No issue for now, but your image is still running on our cluster and I'll update this ticket if I found something.

Thank you for that, we are looking forward for it release now! :-)

pkg/controllers/netpol/pod.go
```diff
@@ -62,34 +60,20 @@ func (npc *NetworkPolicyController) syncPodFirewallChains(networkPoliciesInfo []
...
-
 	dropUnmarkedTrafficRules := func(podName, podNamespace, podFwChainName string) error {
 		// add rule to log the packets that will be dropped due to network policy enforcement
-		comment := "rule to log dropped traffic POD name:" + podName + " namespace: " + podNamespace
-		args := []string{"-m", "comment", "--comment", comment, "-m", "mark", "!", "--mark", "0x10000/0x10000", "-j", "NFLOG", "--nflog-group", "100", "-m", "limit", "--limit", "10/minute", "--limit-burst", "10"}
-		err = iptablesCmdHandler.AppendUnique("filter", podFwChainName, args...)
-		if err != nil {
-			return fmt.Errorf("Failed to run iptables command: %s", err.Error())
-		}
+		comment := "\"rule to log dropped traffic POD name:" + podName + " namespace: " + podNamespace + "\""
```
aauren, line 65: Going from AppendUnique to just adding the rules to the list, results in these rules getting added twice for every pod that contains both an Ingress and Egress policy which essentially drops all traffic, since it checks for the presence of a mark that it later drops, so when it traverses this a second time, traffic is dropped.

aauren: Fixed in the commit that I just pushed.

murali-reddy:
thanks @NaPs for cofirming this PR indeed reduces sync time by significant margin.

pkg/controllers/netpol/policy.go
```diff
@@ -262,55 +241,44 @@ func (npc *NetworkPolicyController) processIngressRules(policy networkPolicyInfo
...
 				policy.name + " namespace " + policy.namespace
-			if err := npc.appendRuleToPolicyChain(iptablesCmdHandler, policyChainName, comment, "", targetDestPodIPSetName, "", ""); err != nil {
+			if err := npc.appendRuleToPolicyChain(policyChainName, comment, "", targetDestPodIPSetName, "", ""); err != nil {
 				return err
 			}
 		}
 
 		if len(ingressRule.srcIPBlocks) != 0 {
 			srcIPBlockIPSetName := policyIndexedSourceIPBlockIPSetName(policy.namespace, policy.name, i)
-			srcIPBlockIPSet, err := npc.ipSetHandler.Create(srcIPBlockIPSetName, utils.TypeHashNet, utils.OptionTimeout, "0")
```
aauren, line 272: We lost this in the conversion and then defaulted to TypeHashIP on all of our sets in the RefreshSet function: https://github.com/cloudnativelabs/kube-router/pull/1039/files#diff-7e4d6faaa20a55d353749c53d06bb4c9697a4498614bcd00997391b1a30af9f3R229

This causes all subnets to expand when sent to ipset which, for large subnets (subnets with addresses > 65,536) forces an error and stops the NetworkPolicyController from syncing.

aauren: Fixed this with the commit I just pushed

aauren:
We've tested this internally in a fairly large dev cluster and have so far only seen a few bugs, all of which have now ben resolved.

Merging.

PR #1042, opened 2021-03-06
fix ipv6 eor conflict

kikimo:
This patch add an option to disable ipv6 configuration when bgp graceful restart is enabled in kube-router.

kube-router enable both ipv4-unicast and ipv6-unicast when bgp graceful restart is enabled. This can cause trouble when kube-router is peering with a switch without ipv6 enabled. According to **RFC 4724 4.1**:

>// Once the session between the Restarting Speaker and the Receiving
// Speaker is re-established, ...snip... it MUST defer route
// selection for an address family until it either (a) receives the 
// End-of-RIB marker from all its peers (excluding the ones with the 
// "Restart State" bit set in the received capability and excluding the 
// ones that do not advertise the graceful restart capability) or (b) 
// the Selection_Deferral_Timer referred to below has expired.

The switch without ipv6 enabled will fail to pass checks in condition (a) and kube-router will fall through to condition (b), that means it will wait until `Selection_Deferral_Timmer` timeout before it can send the first bgp update message, this is a big performance problem for system that requires high qos when kube-router is restarted. Following is the detail analysis:

we assume that:

1. kube-router is peering with a switch with ipv6 support disabled
2. bgp graceful restart is enabled in kube-router
 
the checks in `kube-router/vendor/github.com/osrg/gobgp/pkg/server/fsm.go:1330-1346` will faild to set `GracefulRestart.State.Enabled` and `MpGracefulRestart.State.Received` of ipv6 route family to true:

```go
...
gr, ok := fsm.capMap[bgp.BGP_CAP_GRACEFUL_RESTART]
if fsm.pConf.GracefulRestart.Config.Enabled && ok {
    state := &fsm.pConf.GracefulRestart.State
    state.Enabled = true
    cap := gr[len(gr)-1].(*bgp.CapGracefulRestart)
    state.PeerRestartTime = uint16(cap.Time)

    for _, t := range cap.Tuples {
        n := bgp.AddressFamilyNameMap[bgp.AfiSafiToRouteFamily(t.AFI, t.SAFI)]
        for i, a := range fsm.pConf.AfiSafis {
            if string(a.Config.AfiSafiName) == n {
                fsm.pConf.AfiSafis[i].MpGracefulRestart.State.Enabled = true
                fsm.pConf.AfiSafis[i].MpGracefulRestart.State.Received = true
                break
            }
        }
    }
...
```

therefore the `p.recvedAllEOR()` check in `kube-router/vendor/github.com/osrg/gobgp/pkg/server/server.go:1485:1501` will fail:

```go
else {
// RFC 4724 4.1
// Once the session between the Restarting Speaker and the Receiving
// Speaker is re-established, ...snip... it MUST defer route
// selection for an address family until it either (a) receives the
// End-of-RIB marker from all its peers (excluding the ones with the
// "Restart State" bit set in the received capability and excluding the
// ones that do not advertise the graceful restart capability) or (b)
// the Selection_Deferral_Timer referred to below has expired.
allEnd := func() bool {
    for _, p := range s.neighborMap { 
        if !p.recvedAllEOR() {
            return false    
        }
    }
    return true
}()
```

thus kube-router will fall through to wait for the `Selection_Deferral_Timmer` to timeout.

### DIFF
--- a/pkg/controllers/routing/bgp_peers.go
+++ b/pkg/controllers/routing/bgp_peers.go
@@ -183,7 +183,7 @@ func (nrc *NetworkRoutingController) syncInternalPeers() {
 }
 
 // connectToExternalBGPPeers adds all the configured eBGP peers (global or node specific) as neighbours// connectToExternalBGPPeers adds all the configured eBGP peers (global or node specific) as neighbours
-func connectToExternalBGPPeers(server *gobgp.BgpServer, peerNeighbors []*gobgpapi.Peer, bgpGracefulRestart bool, bgpGracefulRestartDeferralTime time.Duration,
+func connectToExternalBGPPeers(server *gobgp.BgpServer, peerNeighbors []*gobgpapi.Peer, bgpGracefulRestart bool, bgpGracefulRestartIpv6 bool, bgpGracefulRestartDeferralTime time.Duration,
 	bgpGracefulRestartTime time.Duration, peerMultihopTTL uint8) error {
 	for _, n := range peerNeighbors {
 
@@ -214,7 +214,7 @@ func connectToExternalBGPPeers(server *gobgp.BgpServer, peerNeighbors []*gobgpap
 					},
 					MpGracefulRestart: &gobgpapi.MpGracefulRestart{
 						Config: &gobgpapi.MpGracefulRestartConfig{
-							Enabled: true,
+							Enabled: bgpGracefulRestartIpv6,
 						},
 					},
 				},
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -92,6 +92,7 @@ type NetworkRoutingController struct {
 	bgpFullMeshMode                bool
 	bgpEnableInternal              bool
 	bgpGracefulRestart             bool
+	bgpGracefulRestartIpv6         bool
 	bgpGracefulRestartTime         time.Duration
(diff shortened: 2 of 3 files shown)


aauren:
@kikimo thanks for the PR, I appreciate your work on this. I especially appreciate the level of detail that you went into on describing the issue as well as the code references.

However, this feels like a very nuanced flag to be adding to kube-router (i.e. deciding that you only don't want a single capability to be set on a single address family). If we added options with this level of granularity for all such possibilities it would make the kube-router interface too complicated for our users.

I'm not an expert on all of the nuance of BGP, but it strikes me that there should be some more general way to handle this issue on the kube-router/gobgp side. i.e. if the switch isn't IPv6 capable it seems like the speaker should not wait for an EOR marker on that address family.

What happens if a user is deploying kube-router into a mixed environment where sometimes they are connecting to IPv6 capable switches and other times they are connecting to only IPv4 capable switches, but they desire graceful restart on all address families? A global flag would turn off IPv6 graceful restart even to the switches that were capable of supporting it.

I'm going to close this PR for now as I don't think that this would be acceptable for kube-router. Maybe you could propose an alternative approach that doesn't involve adding a global flag, but handles this situation in an issue? Then we could discuss the approach and hopefully land on a solution that:
1) Handles all situations intelligently
2) Does not introduce a global flag

If you'd like you can copy and paste the description of the PR here into the issue to kick it off and then suggest some alternatives.

PR #1043, opened 2021-03-16
Local make or build dependency problem solving.

cloudnativer:
"All the dependencies are vendored already, so just run make or go build -o kube-router kube-router.go to build."This is where I read the build instructions <a href='https://github.com/cloudnativelabs/kube-router/blob/master/docs/developing.md'>on  this page </a>.

I tried to download the latest version of v1.1.1 source code package, and then built it locally. I found that it failed. I found that not all the dependent packages were in vendor, which led me to fail to compile successfully in a localized environment. However, in fact, in some network environments, pulling dependent packets online may also time out due to poor network quality, and finally cannot continue to compile. The requirement of localized compilation is still more important.

So I updated all the dependencies required by kube-router and gobgp into this branch and fixed makefile. Using the updated code of this branch, you can run "make" in the localization environment to build successfully.


Dockerfile
```diff
@@ -9,11 +9,10 @@ RUN apk add --no-cache \
       conntrack-tools \
       curl \
       bash && \
-    mkdir -p /var/lib/gobgp && \
-    mkdir -p /usr/local/share/bash-completion && \
-    curl -L -o /usr/local/share/bash-completion/bash-completion \
```
mrueg, line 14: @murali-reddy is there a runtime need for this bash-completion or would it be safe to remove and minimize?

aauren:
In general we've been moving away from vendored dependencies with the advent of Go modules, so the documentation there is out of date.

IMO it would be better to remove the vendor directory entirely rather than increasing the amount of vendored dependencies as it just adds extra weight to the project and we've had several instances where people mistakenly submit PRs that change the vendored dependency rather than making the change at the upstream project.

I think its fair to assume that if you can pull this project successfully, then a builder should be able to pull down the module dependencies and deal with backoffs. Otherwise builders may also manage the dependencies themselves.

Makefile
```diff
@@ -61,9 +61,9 @@ endif
 test: gofmt ## Runs code quality pipelines (gofmt, tests, coverage, etc)
 ifeq "$(BUILD_IN_DOCKER)" "true"
 	$(DOCKER) run -v $(PWD):/go/src/github.com/cloudnativelabs/kube-router -w /go/src/github.com/cloudnativelabs/kube-router $(DOCKER_BUILD_IMAGE) \
-	    sh -c 'CGO_ENABLED=0 go test -v -timeout 30s github.com/cloudnativelabs/kube-router/cmd/kube-router/ github.com/cloudnativelabs/kube-router/pkg/...'
```
mrueg, line 64: I think we should look into dropping the vendor folder completely.

cloudnativer:
> In general we've been moving away from vendored dependencies with the advent of Go modules, so the documentation there is out of date.

If the document is out of date, I suggest updating it to avoid misunderstanding.

cloudnativer:
In some countries and regions with poor network environment, although we can successfully pull GitHub code, it is difficult to successfully access most of the dependencies in go modules. This is because some countries and regions set some IP firewall restrictions on the Internet boundary, which makes some sites inaccessible. Developers in these countries and regions often need to think of various ways to operate network agents, which is really very troublesome. If we can keep the vendor folder and solve the dependency, then these developers will be more convenient to complete the construction and compilation.

cloudnativer:
I see that vendor folder has been removed from the <a href="https://github.com/cloudnativelabs/kube-router/tree/drop-vendor">drop-vendor</a> branch

aauren:
Closing in favor of #1052

PR #1046, opened 2021-03-18
fix(npc): sync npc on pod label changes

### LINKED ISSUE #1035: NPC doesn't resync when pod's label changed
**What happened?**

when pod's label changed to a value match networkpolicy's rule, it still deny because NPC doesn't call `OnPodUpdate` 
 and ipset/iptables doesn't changed

An e2e-case `should allow ingress access from updated pod`  failed in my cluster, I use kube-router only for networkpolicy, with my own CNI plugins.

1. Step 1
deploy kube-router in a k8s cluster which has cni-plugin deployed.  kube-router has args as:
```
        - --run-router=false
        - --run-firewall=true
        - --run-service-proxy=false
        - --bgp-graceful-restart=true
```
2. Step 2
create pod B with label :  app=server ; 
create a networkpolicy with spec: 
```
spec:
  egress:
  - {}
  ingress:
  - from:
    - podSelector:
        matchLabels:
          app: client
  podSelector:
    matchLabels:
      app: server
  policyTypes:
  - Ingress
  - Egress
```
create a pod A without label,  pod A cannot ping pod B because of this networkpolicy
3. Step 3
label pod A with app=client  , pod A still cannot ping pod B , 
4. Step 4
we call `ipset list `  on the node pod B is running at , we can see there are no update of ipsets. 

 - Kube-Router Version (`kube-router --version`): [e.g. 1.0.1]
 - Kubernetes Version (`kubectl version`) : [e.g. 1.18.3]
 - Kubernetes Deployment Type: [e.g. EKS, GKE, Kops, Kubeadm, etc.]
 - Kube-Router Deployment Type: [e.g. DaemonSet, System Service]
 - Cluster Size: [e.g. 200 Nodes]

**Additional context**

aauren:
fixes #1035

Now that we're not proceeding with the partial syncs that were in #1010 we should still bring this fix into the next release.

### DIFF
--- a/pkg/controllers/netpol/pod.go
+++ b/pkg/controllers/netpol/pod.go
@@ -3,6 +3,7 @@ package netpol
 import (
 	"crypto/sha256"
 	"encoding/base32"
+	"reflect"
 	"strings"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


murali-reddy:
thanks. LGTM

PR #1048, opened 2021-03-20
Fix overly eager IPVS updating in network services controller

### LINKED ISSUE #1026: Spilo Endpoint updates force constant resyncs of ipvs
**What happened?**
I'm running Spilo in my cluster, which updates annotation on the endpoints object:

```
 annotations:
    acquireTime: "2021-01-21T11:18:59.523373+00:00"
    leader: harbor-db-0
    optime: "158075120872"
    renewTime: "2021-01-25T09:58:10.851366+00:00"
    transitions: "7"
    ttl: "30"
```

Thos annotations are updated every 5 - 10 seconds which force a full ipvs resync (which takes ~500ms) this leads to a high load from the kube-router process (50 - 80% of one core)

Changes that don't affect the ipvs tables should not trigger a resync.

Have a bunch of service and update an annotation of the endpoints object, each update triggers a ipvs resync

 - Kube-Router Version (`kube-router --version`): v1.1.1 (after upgrade from ancient 0.2.5)
 - Kube-Router Parameters: [e.g. --run-router --run-service-proxy --enable-overlay --overlay-type=full etc.]:
 ```
        - "--run-router=false"
        - "--run-firewall=true"
        - "--enable-overlay=false"
        - "--run-service-proxy=true"
        - "--ipvs-sync-period=1m"
        - "--masquerade-all=true"
        - "--enable-pod-egress=false"
        - "--kubeconfig=/var/lib/kube-router/kubeconfig"
        - "--enable-cni=false"
        - "--metrics-port=8080"
        - "--metrics-path=/metrics"
        - "-v=2"
```
 - Kubernetes Version (`kubectl version`) : [e.g. 1.18.3] 1.17.17 (in the process of updating to 1.20)
 - Cloud Type: [e.g. AWS, GCP, Azure, on premise] on premise
 - Kubernetes Deployment Type: [e.g. EKS, GKE, Kops, Kubeadm, etc.] manual
 - Kube-Router Deployment Type: [e.g. DaemonSet, System Service] DaemonSet
 - Cluster Size: [e.g. 200 Nodes] 34 nodes

Screenshot 2021-01-25 at 11 17 11

### PR DESCRIPTION
**Pull request by ananace**
Switches the endpoint map comparison in OnEndpointsUpdate from being a DeepEqual, to instead checking that all services exist, and that their associated endpoints are similar.

Ordering is no longer considered important in regards to the IPVS update check.

Should fix #1026 - got an image up on `ananace/kube-router-test` that seems to be doing well in regards to that

Not sure if there's a better way to do this, as I'm not really a Go coder - but the issue seemed simple enough to fix

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -813,6 +813,48 @@ func (nsc *NetworkServicesController) publishMetrics(serviceInfoMap serviceInfoM
 	return nil
 }
 
+func endpointsListsEquivalent(a, b []endpointsInfo) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	for _, epA := range a {
+	        valid := false
+		for _, epB := range b {
+			if epA == epB {
+				valid = true
+				break
+			}
+		}
+
+		if !valid {
+		    return false
+		}
+	}
+	return true
+}
+
+func endpointsMapsEquivalent(a, b endpointsInfoMap) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	// Endpoint maps are considered compatible if the lists of services includes same endpoints
+	// Order is not considered.
+	for k, valA := range a {
+		valB, ok := b[k]
+		if !ok {
+			return false
+		}
+
+		if !endpointsListsEquivalent(valA, valB) {
+			return false
+		}
+	}
+
+	return true
+}
+
 // OnEndpointsUpdate handle change in endpoints update from the API server
 func (nsc *NetworkServicesController) OnEndpointsUpdate(ep *api.Endpoints) {
 
@@ -850,7 +892,7 @@ func (nsc *NetworkServicesController) OnEndpointsUpdate(ep *api.Endpoints) {
 	newServiceMap := nsc.buildServicesInfo()
 	newEndpointsMap := nsc.buildEndpointsInfo()
 
-	if len(newEndpointsMap) != len(nsc.endpointsMap) || !reflect.DeepEqual(newEndpointsMap, nsc.endpointsMap) {
+	if !endpointsMapsEquivalent(newEndpointsMap, nsc.endpointsMap) {
 		nsc.endpointsMap = newEndpointsMap
 		nsc.serviceMap = newServiceMap
 		klog.V(1).Infof("Syncing IPVS services sync for update to endpoint: %s/%s", ep.Namespace, ep.Name)
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -813,24 +813,26 @@ func (nsc *NetworkServicesController) publishMetrics(serviceInfoMap serviceInfoM
 	return nil
 }
 
-func endpointsListsEquivalent(a, b []endpointsInfo) bool {
+// TODO Move to utils
+func unsortedListsEquivalent(a, b []endpointsInfo) bool {
 	if len(a) != len(b) {
 		return false
 	}
 
-	for _, epA := range a {
-	        valid := false
-		for _, epB := range b {
-			if epA == epB {
-				valid = true
-				break
-			}
-		}
+	values := make(map[interface{}]int)
+	for _, val := range a {
+		values[val] = 1
+	}
+	for _, val := range b {
+		values[val] = values[val] + 1
+	}
 
-		if !valid {
-		    return false
+	for _, val := range values {
+		if val == 1 {
+			return false
 		}
 	}
+
 	return true
 }
 
@@ -839,15 +841,13 @@ func endpointsMapsEquivalent(a, b endpointsInfoMap) bool {
 		return false
 	}
 
-	// Endpoint maps are considered compatible if the lists of services includes same endpoints
-	// Order is not considered.
-	for k, valA := range a {
-		valB, ok := b[k]
+	for key, valA := range a {
+		valB, ok := b[key]
 		if !ok {
 			return false
 		}
 
-		if !endpointsListsEquivalent(valA, valB) {
+		if !unsortedListsEquivalent(valA, valB) {
 			return false
 		}
 	}

### CODE REVIEW

File: pkg/controllers/proxy/network_services_controller.go [hunk 1]
```diff
@@ -793,6 +793,42 @@ func (nsc *NetworkServicesController) publishMetrics(serviceInfoMap serviceInfoM
...
 
+func epCompatible(a, b []endpointsInfo) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	for _, epA := range a {
+		for _, epB := range b {
+			if epA == epB {
+				return true
```
aauren, line 804: This would stop after the first match without considering the others right?

i.e. say `a` contains `["foo","bar"]` endpoints and `b` contains `["foo", "baz"]`. The first iteration of this nested loop would see `"foo" == "foo"` and then return true, without considering that `"bar" != "baz"`

You should invert the condition here so that if ever `epA != epB` then `return false` otherwise if you make it all the way to the end of the function `return true`

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -793,6 +793,42 @@ func (nsc *NetworkServicesController) publishMetrics(serviceInfoMap serviceInfoM
 	return nil
 }
 
+func epCompatible(a, b []endpointsInfo) bool {
```
aauren, line 796: I would name these functions something different here that is more descriptive of what they do. Maybe `comparable` or `equivalent`? I'd look around in the Golang API and see what these types of functions are usually called.

Additionally, it would probably be good to move these functions into a utils file or something. I know that you're just following what's already been done, but these controllers are getting really lengthy and its hard to follow the logic. So we should probably start splitting simple functions like this out.

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -812,7 +848,7 @@ func (nsc *NetworkServicesController) OnEndpointsUpdate(ep *api.Endpoints) {
 	newServiceMap := nsc.buildServicesInfo()
 	newEndpointsMap := nsc.buildEndpointsInfo()
 
-	if len(newEndpointsMap) != len(nsc.endpointsMap) || !reflect.DeepEqual(newEndpointsMap, nsc.endpointsMap) {
+	if !compatible(newEndpointsMap, nsc.endpointsMap) {
```
aauren, line 851: Just checking here, you don't need to do something similar for the DeepEqual in `OnServiceUpdate` below?

aauren (changes requested):

pkg/controllers/proxy/network_services_controller.go
ananace: The [`buildServicesInfo()` method](https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/proxy/network_services_controller.go#L1308-L1386) doesn't shuffle the result like [`buildEndpointsInfo()`](https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/proxy/network_services_controller.go#L1419-L1437) does, so the DeepEqual there should hopefully do better in that comparison

pkg/controllers/proxy/network_services_controller.go
ananace: Absolutely true, that logic mistake probably happened as I split out the endpoint comparison to a separate method - same as the missing early exit on that loop.

pkg/controllers/proxy/network_services_controller.go
ananace: I've not used Go much, but wouldn't such a method have to - by definition - be a public one in order to be extracted to a separate utility file?
Which would be a bit strange as it's a rather single-purpose method to check if two endpoint maps inside the update are compatible.

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -793,6 +793,48 @@ func (nsc *NetworkServicesController) publishMetrics(serviceInfoMap serviceInfoM
...
+				valid = true
+				break
+			}
+		}
+
+		if !valid {
+		    return false
+		}
+	}
+	return true
```
aauren, line 814: @ananace thanks so much for your work!

Sorry, that it has taken me so long to come back around to this. However, I think we still need one more change here. Unlike map (hash) keys, lists (slices) aren't guaranteed to be unique, so it's not enough here to only evaluate the size and make sure that all of the endpoints in A exist in B, as A may have some duplicate entries and still be the same size. We should be complete here and also iterate over B and make sure that all of the endpoints in B also exist in A.

While I think that the logic you have here should be sufficient for 99.9% of the use-cases I can think up, and I truly don't want to double-iterate here, working on this project has taught me that there are always edge cases and bugs like this can be hard to track down.

So two options here:
1) Add another function to make the slices have unique values (usually in Go this is done by transferring them to hash keys) before comparing them (if you make them a map you could re-use your existing function below)
2) Duplicate the range function you have here, this time making b the outside loop and a the inside loop

I think option 1 would likely be more performant and would allow you to re-use logic.

aauren (changes requested):
As soon as the final comment is addressed, I'll review and should be able to merge.

pkg/controllers/proxy/network_services_controller.go
aauren: Yes, it would have to be public, and while its a single use now, this seems like the type of thing that will doubtless be useful later. https://github.com/cloudnativelabs/kube-router/blob/master/pkg/utils/service.go would be a good spot to put this.

pkg/controllers/proxy/network_services_controller.go
aauren: I think that you are likely right. After the two corrections above are made, I'll test it out and double-check in one of my clusters.

pkg/controllers/proxy/network_services_controller.go
ananace: As I'm not very used to Go, what would be the best way to set this up, as the `endpointInfo` structs can't be used for an `interface{}` method - if I want to put it as a generic method.
Considering the data in question is marked as internal, should it be moved to be public as well?

aauren (approved):
LGTM! Thanks for all of your work on this!

PR #1052, opened 2021-03-22
Drop vendor, update dependencies, upgrade docker images

mrueg:
* Remove the vendor folder
* Update golang dependencies
*  Update the docker images to use latest go 1.15.x
*  Fix a broken symlink in documentation


go.mod
```diff
@@ -2,46 +2,42 @@ module github.com/cloudnativelabs/kube-router
...
+	github.com/vishvananda/netns v0.0.0-20210104183010-2eb08e3e575f
+	golang.org/x/net v0.0.0-20210316092652-d523dce5a7f4
+	google.golang.org/grpc v1.36.0
+	k8s.io/api v0.20.5
+	k8s.io/apimachinery v0.20.5
+	k8s.io/client-go v0.20.5
+	k8s.io/cri-api v0.20.5
 )
 
 go 1.13
```
aauren, line 43: @mrueg should this version here get bumped as well? To 1.15?

aauren (approved):

go.mod
mrueg: I think we should still be fine with go 1.13, not we're not relying on anything that is more recent (this is a minimum version)

aauren (approved):

murali-reddy:
LGTM

Issue #1058, opened 2021-04-07
Is kube-router compatible with netfilter based iptables (i.e. CentOS8/RHEL8 w/o full support for legacy iptables)?


ghglza:
The kernels of CentOS8 and RHEL8 have switched the iptables kernel interface to netfilter.
The iptables CLI tool of these distros supports some sort of backward compatibility, but there is no iptables-legacy version available (like in e.g. some Ubuntu distros).
Is the kube-router CNI plugin compatible with this version of the netfilter based interface?

I have seen that other CNI plugins use a [wrapper script](https://github.com/kubernetes-sigs/iptables-wrappers) around calls of iptables that seems to select the correct kernel interface and command line switches, but I see no references to this script in kube-router's sources.
Does this mean kube-router (esp. its kube-proxy replacement) is not nftables-ready -- or can it handle the new interface, and does this by other means than the wrapper script?

**Describe the solution you'd like**
Please add documentation concerning the state of compatiblity with netfilter based iptables, esp. under CentOS8 and RHEL8

murali-reddy:
I believe it should be compatible but need to test and confirm. Kube-router image is already built with `alpine:3.12` which includes both legacy and nft version of iptables as noted in https://github.com/cloudnativelabs/kube-router/issues/592

ghglza:
Thanks!
So then I suppose the question boils down to: does kube-router recognize when the kernel uses nftables based IP tables, and use the nft version of iptables in this case?
And, has nobody up to now performed compatibility tests of kube-router under CentOS8/RHEL8?

aauren:
Thanks for reaching out and for pointing out the upstream wrapper script.

I don't know that anyone has yet tested under CentOS8/RHEL8. For us, we run kube-router locally on the node, so we don't really have the issue of a potential mismatch. However, I'm aware that we are in the minority.

I don't have a way to test the wrapper functionality as it doesn't work for our use-case, but if you have access to CentOS8/RHEL8 and wanted to test, I added the PR #1061 for this purpose and even built a test container if you want to give it a shot and let us know how it works: aauren/kube-router-git:514ea9c85

If you (or someone else from the community) is able to test and give us a 👍 or 👎 we could probably move on this for the next release.

aauren:
Closing this issue in favor of for further tracking of this issue: #1069

PR #1061, opened 2021-04-08
feat(iptables): add upstream wrapper script

aauren:
This adds the upstream iptables-wrapper script from upstream to intelligently select which version of iptables to use based upon the host system as well as where other iptables entries have been added already (so that they don't become out of order).

See https://github.com/kubernetes-sigs/iptables-wrappers for more information

Fixes: #1058


aauren:
FYI what's failing here is the fact that the arm64 Alpine container doesn't appear to carry the binary `iptables-nft`. I've reached out to @mrueg on how to fix this.

rkojedzinszky:
When is it expected to be merged?

mrueg (approved):

PR #1073, opened 2021-04-21
skip hostnetwok pod's from enforcing network policies

### LINKED ISSUE #1067: host egress traffic blocked when default deny egress network policy applied with pods in hostNetwork
**What happened?**
When [default deny egress](https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-deny-all-egress-traffic) is applied in a namespace, and a host has pods in hostNetwork then traffic is blocked for host as well 

Host should be able to send the egress traffic. Network policy rule should only apply to the pods

- deploy network policy https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-deny-all-egress-traffic in a namespace
- deploy pods with hostNetwork in the same namespace
- check host is able to send outbound traffic

 this issue exists possibly before 1.0.0 release as well

murali-reddy:
Fixes #1067 

existing KUBE-POD-FW rules for any host network pods will be automatically cleaned up on first sync.

### DIFF
--- a/pkg/controllers/netpol/pod.go
+++ b/pkg/controllers/netpol/pod.go
@@ -37,7 +37,7 @@ func (npc *NetworkPolicyController) newPodEventHandler() cache.ResourceEventHand
 func (npc *NetworkPolicyController) OnPodUpdate(obj interface{}) {
 	pod := obj.(*api.Pod)
 	if pod.Spec.HostNetwork {
-		klog.V(2).Info("Ignoring update to hostNetwork pod: %s/%s", pod.Namespace, pod.Name)
+		klog.V(2).Infof("Ignoring update to hostNetwork pod: %s/%s", pod.Namespace, pod.Name)
 		return
 	}
 	klog.V(2).Infof("Received update to pod: %s/%s", pod.Namespace, pod.Name)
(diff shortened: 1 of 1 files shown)


pkg/controllers/netpol/pod.go
```diff
@@ -36,6 +36,10 @@ func (npc *NetworkPolicyController) newPodEventHandler() cache.ResourceEventHand
 // OnPodUpdate handles updates to pods from the Kubernetes api server
 func (npc *NetworkPolicyController) OnPodUpdate(obj interface{}) {
 	pod := obj.(*api.Pod)
+	if pod.Spec.HostNetwork {
+		klog.V(2).Info("Ignoring update to hostNetwork pod: %s/%s", pod.Namespace, pod.Name)
```
mrueg, line 40:
```suggestion
		klog.V(2).Infof("Ignoring update to hostNetwork pod: %s/%s", pod.Namespace, pod.Name)
```

aauren:
@murali-reddy Maybe you can explain a bit more, but I'm not sure how this would fix #1067 as this would only stop us from requesting a full sync when the pod's spec is set to HostNetwork.

However, any subsequent pod update would cause the full sync to get requested which would eventually run `syncPodFirewallChains()` (https://github.com/cloudnativelabs/kube-router/blob/1fb082004495700f8554f1a9a744fae12c587835/pkg/controllers/netpol/pod.go#L62) which calls `getIngressNetworkPolicyEnabledPods()` and `getEgressNetworkPolicyEnabledPods()` which list ALL pods (including hostNetwork'ed ones) and it will then add jump targets for their pod IP address which would be the node's IP address which would result in network policies incorrectly targeting the host correct?

murali-reddy:
@aauren updated `getIngressNetworkPolicyEnabledPods()` and `getEgressNetworkPolicyEnabledPods()` to skip host networked pod a well

aauren:
LGTM!

PR #1075, opened 2021-04-28
Mention k0sproject in getting started guide

soider:
k0s project uses kube-router as default solution for the CNI, so I'd like to mention it in a list of the ways to check kube-router in action.

Signed-off-by: Mikhail Sakhnov <mikhail@skhnv.me>


aauren:
LGTM! Thanks for submitting this!

PR #1078, opened 2021-05-04
Make pod firewall nflog rate-limiting configurable

bnu0:
For compliance reasons, it might be desired to tune the nflog rate-limits for packets that are rejected from pod firewalls. This PR makes it possible to tune the limit/burst parameters, or remove limiting entirely if you know what you are doing.

The defaults remain at `burst=10`, `limit=10/minute` as before.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -84,6 +84,8 @@ Usage of kube-router:
       --peer-router-passwords strings                 Password for authenticating against the BGP peer defined with "--peer-router-ips".
       --peer-router-passwords-file string             Path to file containing password for authenticating against the BGP peer defined with "--peer-router-ips". --peer-router-passwords will be preferred if both are set.
       --peer-router-ports uints                       The remote port of the external BGP to which all nodes will peer. If not set, default BGP port (179) will be used. (default [])
+      --pod-fw-rejects-log-burst int                  The number of rejected packets logged from pod firewalls to nflog before the limit defined with "--pod-fw-rejects-log-limit" is used to ratelimit. (default 10)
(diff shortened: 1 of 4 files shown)


bnu0:
The tests failed, because the test i added back in #1068 was flaky. The order of destroying sets was based on map iteration, which isn't predictable. I fixed that here.

aauren:
@bnu0 I definitely understand the motivation for this PR. However, I'm hesitant to add more CLI flags for kube-router for something this obscure (at least IMO).

One of the benefits to kube-router is that it is easy to understand and configure when compared to the other offerings available. So we try to police how many flags we add to the interface so as to only highlight major features and try to choose sane defaults rather than potentially overwhelming the end-user with configuration options for every knob that could be turned.

I'll likely wait and see if @murali-reddy has any opinions on whether or not to include this feature.

aauren:
Also, thanks for the stabilizing that unit test, I actually cherry-picked that commit and pushed it to master ahead of this PR so regardless of the action here, unit tests will be more stable in the future: https://github.com/cloudnativelabs/kube-router/commit/fce90b07ece374e4e73011e2f99759bf1df7dee7

If you can, please rebase your PR so that commit gets taken out.

aauren:
Thanks for creating this, and sorry that it has taken so long to come back around to this. However, adding 2 additional parameter flags isn't worth what this feature brings to the table.

I definitely see how it could be useful in certain situations, but given that its been posted for almost a year and no one else has chimed in, I don't think that it would be widely used.

Maybe in the future, if we create more ways of configuring kube-router, maybe expose some conf like mechanism for more advanced options or some such, then we could revisit this.

Issue #1081, opened 2021-05-12
kube router generates incorrect ipsets


phin1x:
**What happened?**
after upgrading from kube router 1.1.1 to 1.2.2 the network communication on some nodes of our cluster suddenly stops after some time. we found out that the ipset for the local ips got messed up, all entries are gone. we checked the ipsets on the working nodes and there the ipset kube-router-local-ips has three or four members (depending on the node type).

**What did you expect to happen?**
kube router should not discard the members in the ipset kube-router-local-ips

**How can we reproduce the behavior you experienced?**
We cannot say how this problem can be reproduced, since from our point of view it appears out of nowhere.

** Screenshots / Architecture Diagrams / Network Topologies **

![Screenshot 2021-05-12 131044](https://user-images.githubusercontent.com/46140201/117968404-7059e200-b326-11eb-9136-8f2f79637684.png)

** System Information (please complete the following information):**
 - Kube-Router Version 1.2.2
 - Kube-Router Parameters: --run-router=true --run-firewall=true --run-service-proxy=true --kubeconfig=/var/lib/kube-router/kubeconfig --metrics-port=9912
 - Kubernetes Version (`kubectl version`) : 1.20.5
 - Cloud Type: on premise
 - Kubernetes Deployment Type: custom setup
 - Kube-Router Deployment Type: DaemonSet
 - Cluster Size: 4

** Logs, other output, metrics **
No logs were recorded at the time the error occurred

**Additional context**

aauren:
@phin1x Thanks for reporting. Unfortunately, if we can't reproduce it there may not be much to do here. I can maybe give you more information in case you have a node that is currently in this state, or if it happens again.

The `kube-router-local-ips` ipset is populated by this method: [getAllLocalIPs](https://github.com/cloudnativelabs/kube-router/blob/bd5ee4f7082512b89a5dc220a3b5e7c2e88afa0c/pkg/controllers/proxy/network_services_controller.go#L2221)

This method essentially lists all of your Linux network interfaces (i.e. `ip link`) on the node kube-router is running on and if their name doesn't contain `dummy`, `kube`, or `docker` it will add all of their IPv4 addresses to this ipset.

If there is an error while it is attempting to get the IP addresses you should see a message like the following in your kube-router logs: `Error syncing ipvs svc iptables rules to permit traffic to service VIP's: failed to get local IPs: ...`. It is emitted at the error level, so it should be present no matter what.

[...]

PR #1088, opened 2021-05-19
Add Locking Around ipset Invocations

aauren:
@murali-reddy @mrueg 

Potential fix for #1081 

I'm marking this as a WIP as I'm not 100% sure that this is what's causing a problem, but at this point I highly suspect that there is contention between ipset invocations within the different controllers. Especially since most of the periodic syncs happen right on top of each other (by default the 5m period syncs are on a ticker that end up being scheduled within milliseconds of each other).

I suspect that what happens is that on some starts when the timing of the periodic syncs is incredibly unlucky, the controllers end up executing something like this:
1. NPC takes a snapshot of the system’s ipset state: https://github.com/cloudnativelabs/kube-router/blob/96675e620be3e0c152d5b873146cc2f7f54c9429/pkg/controllers/netpol/policy.go#L83
2. NPC continues to process but is no longer holding a lock on ipset since it doesn’t have any commands in process
3. NSC/NRC makes some changes to an ipset: https://github.com/cloudnativelabs/kube-router/blob/4306e5d47cba468f24b6e259a5514e883172d00f/pkg/controllers/proxy/network_services_controller.go#L672
4. NPC restores its now stale ipset rules back to the system which does not contain the changes from step 3: https://github.com/cloudnativelabs/kube-router/blob/96675e620be3e0c152d5b873146cc2f7f54c9429/pkg/controllers/netpol/policy.go#L136

Unfortunately, this collision is really difficult to track down, because even a restart of one kube-router process can cause the collision to no longer be in that magical window where it causes problems.

I inserted some debug into each of the control loops and I can see that during the periodic syncs they all execute right on top of each other, but in order to be within one of the NPC controller's ipset save/restore portion requires precise timing:
```
I0519 19:47:21.469114   33414 network_services_controller.go:412] ---- Performing periodic sync of ipvs services
I0519 19:47:21.469165   33414 network_services_controller.go:437] ---- START ipvs sync
I0519 19:47:21.473170   33414 network_routes_controller.go:289] ---- START periodic sync of routes
I0519 19:47:21.478918   33414 service_endpoints_sync.go:34] ---- START ipvs services sync
I0519 19:47:21.727247   33414 network_routes_controller.go:338] ---- END periodic sync routes took 254.079823ms
I0519 19:47:21.737685   33414 network_policy_controller.go:182] ---- PERIODIC policy sync
I0519 19:47:21.737728   33414 network_policy_controller.go:218] ---- START sync of iptables with version: 1621453641737721625
I0519 19:47:21.895080   33414 ipset.go:480] ---- START ipset restore
I0519 19:47:22.280165   33414 service_endpoints_sync.go:31] ---- END ipvs services sync took 801.24877ms
I0519 19:47:22.303753   33414 network_services_controller.go:465] ---- END ipvs sync
I0519 19:47:22.446412   33414 ipset.go:491] ---- END ipset restore
I0519 19:47:22.529935   33414 iptables.go:34] ---- START iptables restore
I0519 19:47:22.581375   33414 iptables.go:53] ---- END iptables restore
I0519 19:47:22.610633   33414 network_policy_controller.go:215] ---- END sync iptables took 872.905003ms
```

The main thing that doesn't make sense to me at this point, which is why this is still a WIP, is that I'm not sure why missing an update causes such a big impact. It seems to me that the end result of this would be some stale data for a while until another update came in and caused the controller to fix the issue, and not missing data, or bad data. However, the impact that we see is disruption to services and sometimes the node itself.

Also, this isn't a very elegant solution. There are probably better ways to handle this.

### DIFF
--- a/pkg/cmd/kube-router.go
+++ b/pkg/cmd/kube-router.go
@@ -72,6 +72,7 @@ func CleanupConfigAndExit() {
 // Run starts the controllers and waits forever till we get SIGINT or SIGTERM
 func (kr *KubeRouter) Run() error {
 	var err error
+	var ipsetMutex sync.Mutex
 	var wg sync.WaitGroup
 	healthChan := make(chan *healthcheck.ControllerHeartbeat, 10)
 	defer close(healthChan)
@@ -140,7 +141,8 @@ func (kr *KubeRouter) Run() error {
 	}
 
 	if kr.Config.RunRouter {
-		nrc, err := routing.NewNetworkRoutingController(kr.Client, kr.Config, nodeInformer, svcInformer, epInformer)
+		nrc, err := routing.NewNetworkRoutingController(kr.Client, kr.Config,
+			nodeInformer, svcInformer, epInformer, &ipsetMutex)
 		if err != nil {
 			return errors.New("Failed to create network routing controller: " + err.Error())
 		}
@@ -162,7 +164,7 @@ func (kr *KubeRouter) Run() error {
 
 	if kr.Config.RunServiceProxy {
 		nsc, err := proxy.NewNetworkServicesController(kr.Client, kr.Config,
-			svcInformer, epInformer, podInformer)
+			svcInformer, epInformer, podInformer, &ipsetMutex)
 		if err != nil {
 			return errors.New("Failed to create network services controller: " + err.Error())
 		}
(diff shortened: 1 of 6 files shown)


aauren:
We've been running this patch on multiple clusters for a while, and all of the symptoms of #1081 have gone away while running this patch. Its difficult to know definitively if this 100% solves #1081 as this issue was so transient for us, sometimes being resolved just by restarting the kube-router instance. However, given that the problem used to evidence itself at least once in a cluster per 24-hour period, and it has not resurfaced since we applied this patch over the last 5 days I'm going to call it good.

As for the lack of elegance for this patch, @murali-reddy said that he is willing to refactor this locking mechanism in a later release. For now we've agreed to merge this as it should resolve a significant problem for a large enough group of users.

Issue #1056, opened 2021-04-07
Network policies are applied for pods that are not running (completed jobs)


zerkms:
**What happened?**

I deployed an application and found out that I cannot connect its ports. All looked fine but still - connection was actively refused. After some debugging I have found that on that node there are 2 pods with the same IP assigned: one is the pod I'm talking about, and the other is a pod from a job that has successfully completed 15 hours ago.

Both have `10.51.5.17` IP assigned.

The job does have network policies configured, the application I just deployed - does not. Yet, there are iptables rules applied for the job still, even though it's completed.

```
Chain KUBE-POD-FW-AIWVY7TMPWP7PXQH (3 references)
    pkts      bytes target     prot opt in     out     source               destination
    3087   201342 ACCEPT     all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* rule for stateful firewall for pod */ ctstate RELATED,ESTABLISHED
       0        0 ACCEPT     all  --  *      *       0.0.0.0/0            10.51.5.17           /* rule to permit the traffic traffic to pods when source is the pod's local node */ ADDRTYPE match src-type LOCAL
       0        0 KUBE-NWPLCY-SB6L4XPQ7W7HDE22  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* run through nw policy default-ingress-deny */
       0        0 NFLOG      all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* rule to log dropped traffic POD name:backup-sync-adm-1617704100-hmrzs namespace: adm-backup-hq */ mark match ! 0x10000/0x10000 limit: avg 10/min burst 10 nflog-group 100
       0        0 REJECT     all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* rule to REJECT traffic destined for POD name:backup-sync-adm-1617704100-hmrzs namespace: adm-backup-hq */ mark match ! 0x10000/0x10000 reject-with icmp-port-unreachable
       0        0 MARK       all  --  *      *       0.0.0.0/0            0.0.0.0/0            MARK and 0xfffeffff

```

**What did you expect to happen?**

I suspect that if assigning the same IP is possible - then rules from the stopped pod should be cleaned up.

**How can we reproduce the behavior you experienced?**
1. Schedule enough jobs on the same node so that IPs wrapped around and started from the beginning, also - make sure you apply some network policy in those jobs namespace
[...]

PR #1090, opened 2021-05-23
Make Pod NetworkPolicy Handling More Efficient / Correct

aauren:
@murali-reddy @mrueg 

This has a couple of changes in it that make kube-router's NetworkPolicy handling better:
1. Don't process policy for pod statuses that are `Completed`, `Succeeded`, or `Failed`. This changes kube-router to match what other network plugins do (e.g. https://github.com/projectcalico/libcalico-go/blob/9bbd69b5de2b6df62f4508d7557ddb67b1be1dc2/lib/backend/k8s/conversion/conversion.go#L157-L171 thanks @chendotjs). Additionally, this fixes the issue reported in #1056 where errors were caused because the CNI will re-use the Pod IP address of pods in these phases which can improperly apply NetworkPolicy to pods that should be without it.
2. Don't process new pods (i.e. ones going through `AddFunc()`) if they are missing their PodIP, are in an invalid phase ^^^, or are HostNetwork, as NetworkPolicy can't be properly applied to these pods anyway. This optimizes kube-router's NPC controller to be more efficient and avoid early full syncs that wouldn't have worked correctly.
3. Still request a full sync on pod update if the updated pod has `HostNetwork=true`. It is important to still process this event as the pod may have changed from `HostNetwork=false -> HostNetwork=true` in which case we need to remove NetworkPolicy that was previously applied to the Pod's previous Pod IP.
4. Refactor/Consolidate a lot of items into the utils package so that it can be more easily re-used and tested (also added unit testing for all of the above ^^^)
5. Now includes testing for changed `PodIPs` array and `HostIP` to cover all of the bases when testing if kube-router should request a full sync off of a pod update.

### DIFF
--- a/pkg/controllers/netpol/pod.go
+++ b/pkg/controllers/netpol/pod.go
@@ -36,10 +36,6 @@ func (npc *NetworkPolicyController) newPodEventHandler() cache.ResourceEventHand
 // OnPodUpdate handles updates to pods from the Kubernetes api server
 func (npc *NetworkPolicyController) OnPodUpdate(obj interface{}) {
 	pod := obj.(*api.Pod)
-	if pod.Spec.HostNetwork {
-		klog.V(2).Infof("Ignoring update to hostNetwork pod: %s/%s", pod.Namespace, pod.Name)
-		return
-	}
 	klog.V(2).Infof("Received update to pod: %s/%s", pod.Namespace, pod.Name)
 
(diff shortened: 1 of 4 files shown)


murali-reddy:
thanks @aauren for the tidying up all the pod life-cycle cases. Changes LGTM. 

I will help with testing the PR and report if any issues.

PR #1092, opened 2021-05-27
.github: Enable dependabot to manage dependencies

mrueg:
See also: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates

@aauren @murali-reddy this enables dependabot to manage and updates golang dependencies once per week. It will send PRs with suggested updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "monthly"


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
aauren, line 6: Weekly seems a little too frequent here. We don't cut releases weekly, so no one will really be using weekly updates and I don't know that I want to review weekly dependency updates. Monthly would probably be sufficient as it is a little closer to our release cadence.

mrueg: Changed it to monthly.

aauren (approved):

aauren:
@mrueg do you know if there is a way to have dependbot group the dep updates into 1 PR rather than having a PR per dep?

PR #1098, opened 2021-06-01
fix(goreleaser): add version and date to published binaries

aauren:
@mrueg 

This should have been included in #1031 as without this, published binaries made by goreleaser don't print build version/date correctly.


mrueg:
Thanks!

PR #1110, opened 2021-06-27
use wait option of iptables-restore when available

murali-reddy:
update alpine image to 3.14 which has latest version of `iptables-restore` command with `--wait` option.

Also check if the `iptables-restore` has the wait option (to cover previous vesion of command with no --wait option) and use it

### DIFF
--- a/pkg/utils/iptables.go
+++ b/pkg/utils/iptables.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os/exec"
+	"strings"
 )
 
 // SaveInto calls `iptables-save` for given table and stores result in a given buffer.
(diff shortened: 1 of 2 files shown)


pkg/utils/iptables.go
```diff
@@ -32,15 +33,30 @@ func Restore(table string, data []byte) error {
 	if err != nil {
 		return err
 	}
-	args := []string{"iptables-restore", "-T", table}
+	var args []string
+	args = []string{"iptables-restore", "--help"}
```
aauren, line 37: @murali-reddy Can you think of any easy way that we could just cache this on startup so that we don't have the extra command call for each call to `Restore()`? Maybe make it it's own function and then call it at the beginning of the NPC initialization?

aauren (approved):
LGTM

pkg/utils/iptables.go
murali-reddy: Good suggestion. Incorporated.

PR #1133, opened 2021-07-26
remove IPVS metrics

### LINKED ISSUE #734: Prometheus service metrics are never removed
* Kubernetes version: 1.13.2 (installed via kubeadm)
* Kube-router version v0.3.1, built on 2019-05-12T09:22:07+0000, go1.10.8

Daemonset container args:
```
- args:
  - --run-router=true
  - --run-firewall=true
  - --run-service-proxy=true
  - --metrics-port=8080
```

We've been running kube-router for a few weeks and have noticed that service metrics do not appear to be removed from the prometheus metrics page, which is causing excessive tsdb writes with our prometheus server. Restarting the kube-router process brought our tsdb writes back to a normal level.

```
$ kubectl get svc -n example
NAME                         TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)     AGE
example-deploy                ClusterIP   10.97.203.41    <none>        80/TCP   15m
$ curl 192.168.100.102:8080/metrics --silent | grep example-deploy
kube_router_service_bps_in{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 13
kube_router_service_bps_out{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 2817
kube_router_service_bytes_in{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 104739
kube_router_service_bytes_out{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 2.0942055e+07
kube_router_service_cps{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 0
kube_router_service_packets_in{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 1686
kube_router_service_packets_out{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 1938
kube_router_service_pps_in{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 0
kube_router_service_pps_out{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 0
kube_router_service_total_connections{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 39
$
$ ...
$ kubectl delete svc -n example example-deploy
$ ...
$ 
$ kubectl get svc -n example
NAME                         TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
$
$ ...
$
$ curl 192.168.100.102:8080/metrics --silent | grep example-deploy
kube_router_service_bps_in{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 13
kube_router_service_bps_out{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 2817
kube_router_service_bytes_in{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 104739
kube_router_service_bytes_out{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 2.0942055e+07
kube_router_service_cps{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 0
kube_router_service_packets_in{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 1686
kube_router_service_packets_out{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 1938
kube_router_service_pps_in{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 0
kube_router_service_pps_out{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 0
kube_router_service_total_connections{port="80",protocol="tcp",service_name="example-deploy",service_vip="10.97.203.41",svc_namespace="example"} 39
```

bhcleek:
Remove metrics for IPVS services when the IPVS service is deleted so
that the number of metrics does not grow without bound.

Fixes #734

### DIFF
--- a/pkg/controllers/proxy/service_endpoints_sync.go
+++ b/pkg/controllers/proxy/service_endpoints_sync.go
@@ -548,6 +548,23 @@ func (nsc *NetworkServicesController) cleanupStaleIPVSConfig(activeServiceEndpoi
 					ipvsServiceString(ipvsSvc), err.Error())
 				continue
 			}
+
+			labelValues, ok := nsc.metricsMap[key]
+			if !ok {
+				continue
+			}
+
+			metrics.ServiceBpsIn.DeleteLabelValues(labelValues...)
+			metrics.ServiceBpsOut.DeleteLabelValues(labelValues...)
+			metrics.ServiceBytesIn.DeleteLabelValues(labelValues...)
+			metrics.ServiceBytesOut.DeleteLabelValues(labelValues...)
+			metrics.ServiceCPS.DeleteLabelValues(labelValues...)
+			metrics.ServicePacketsIn.DeleteLabelValues(labelValues...)
+			metrics.ServicePacketsOut.DeleteLabelValues(labelValues...)
+			metrics.ServicePpsIn.DeleteLabelValues(labelValues...)
+			metrics.ServicePpsOut.DeleteLabelValues(labelValues...)
+			metrics.ServiceTotalConn.DeleteLabelValues(labelValues...)
+			metrics.ControllerIpvsServices.Dec()
 		} else {
 			dsts, err := nsc.ln.ipvsGetDestinations(ipvsSvc)
 			if err != nil {
--- a/pkg/controllers/proxy/service_endpoints_sync.go
+++ b/pkg/controllers/proxy/service_endpoints_sync.go
@@ -63,6 +63,11 @@ func (nsc *NetworkServicesController) syncIpvsServices(serviceInfoMap serviceInf
 		syncErrors = true
 		klog.Errorf("Error cleaning up stale IPVS services and servers: %s", err.Error())
 	}
+	err = nsc.cleanupStaleMetrics(activeServiceEndpointMap)
+	if err != nil {
+		syncErrors = true
+		klog.Errorf("Error cleaning up stale metrics: %s", err.Error())
+	}
 	err = nsc.syncIpvsFirewall()
 	if err != nil {
 		syncErrors = true
@@ -548,24 +553,6 @@ func (nsc *NetworkServicesController) cleanupStaleIPVSConfig(activeServiceEndpoi
 					ipvsServiceString(ipvsSvc), err.Error())
 				continue
 			}
-
-			labelValues, ok := nsc.metricsMap[key]
-			if !ok {
-				continue
-			}
-
-			metrics.ServiceBpsIn.DeleteLabelValues(labelValues...)
-			metrics.ServiceBpsOut.DeleteLabelValues(labelValues...)
-			metrics.ServiceBytesIn.DeleteLabelValues(labelValues...)
-			metrics.ServiceBytesOut.DeleteLabelValues(labelValues...)
-			metrics.ServiceCPS.DeleteLabelValues(labelValues...)
-			metrics.ServicePacketsIn.DeleteLabelValues(labelValues...)
-			metrics.ServicePacketsOut.DeleteLabelValues(labelValues...)
-			metrics.ServicePpsIn.DeleteLabelValues(labelValues...)
-			metrics.ServicePpsOut.DeleteLabelValues(labelValues...)
-			metrics.ServiceTotalConn.DeleteLabelValues(labelValues...)
-			metrics.ControllerIpvsServices.Dec()
-			delete(nsc.metricsMap, key)
 		} else {
 			dsts, err := nsc.ln.ipvsGetDestinations(ipvsSvc)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controllers/proxy/service_endpoints_sync.go
```diff
@@ -548,6 +548,23 @@ func (nsc *NetworkServicesController) cleanupStaleIPVSConfig(activeServiceEndpoi
 					ipvsServiceString(ipvsSvc), err.Error())
 				continue
 			}
+
+			labelValues, ok := nsc.metricsMap[key]
```
aauren, line 552: I'm always leary about storing string states in the control loop structure as it has the potential to bloat quickly on some cluster edge cases and then cause memory leaks.

In this case, I think that your use here is justified, as you have access to some information on publish that you don't have here. That being said, I'd like to be more careful that this map never has the chance to grow unbound. So I'd like to see two changes.

1. Delete the items in the map when you're done with them (i.e. in this step)
2. I see a small edge case here that the loop that you're embedding this into may not accurately reflect the current state of `nsc.metricsMap`. This loop is based upon the current ipvs services that it found on the node. However, something could happen where the map and the set of current ipvs services on the node become different and there may be some entries in the map that won't ever get cleaned. So I think it would be better if this logic was in its own loop.

Instead of nesting this here, abstract this to its own function, with `activeServiceEndpointMap` as a parameter? Then you could call it from `syncIpvsServices()` directly after `cleanupStaleVIPs()`. In this function you would loop over `nsc.metricsMap`, look for a matching entry in `activeServiceEndpointMap` (the key structures *should* match if I'm reading the code correctly), if you don't find a match, do the `DeleteLabelValues` calls along with deleting the `nsc.metricsMap` entry as well.

aauren (changes requested):

aauren:
@bhcleek thanks for looking into this issue. I'd recommend waiting maybe a day or two to see if @murali-reddy has any other recommended changes before proceeding with the items that I put in there (just so you don't waste any development time).

If you don't hear anything back after a day or two, just go for the changes that I put in there, and I'll merge it after another quick review.

pkg/controllers/proxy/service_endpoints_sync.go
bhcleek: Good catch on the map key deletion.

I can move this out and call as you suggest, but a lot of this function will need to get duplicated there in order to get the services, check excluded cidrs, derive the key for the service, etc. Are you sure that's what you want?

If it is what you want, then it seems to me that the new function should be called directly after `cleanupStaleIPVSConfig` or else the metrics would be deleted too soon (i.e. potentially before the service itself is actually deleted). WDYT?

aauren: Maybe you're seeing something that I'm not, but I think you can just do something like:
```
if !nsc.MetricsEnabled {
  return
}

for key, values := range nsc.metricsMap {
  if _, ok := activeServiceEndpointMap[key]; !ok {
    <delete metric labels and entry from map>
  }
}
```

This should work because the key structure of your map and `activeServiceEndpointMap` should be the same: https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/proxy/service_endpoints_sync.go#L120 https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/proxy/service_endpoints_sync.go#L145

And you're right, it should be after `cleanupStaleIPVSConfig` sorry, I got that wrong on my previous comment.

bhcleek: I think you're right about the implementation; I was thinking that we'd have to handle iterating over the IPVS services and the differences with FWMark, but I agree with your approach. New commit is coming up.

bhcleek:
@aauren I think I've addressed all of your concerns. PTAL

aauren (changes requested):

pkg/controllers/proxy/service_endpoints_sync.go
```diff
@@ -575,3 +580,25 @@ func (nsc *NetworkServicesController) cleanupStaleIPVSConfig(activeServiceEndpoi
 	}
 	return nil
 }
+
+func (nsc *NetworkServicesController) cleanupStaleMetrics(activeServiceEndpointMap map[string][]string) error {
```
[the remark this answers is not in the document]
bhcleek: Yeah, I was torn on this. I wasn't sure if you preferred consistency with the other similar functions or not.

Change coming up.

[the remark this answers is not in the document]
bhcleek: PTAL

aauren (approved):
Thanks @bhcleek! I appreciate all of your work. 🙂 

@murali-reddy said that he would try to make some time to take a look at this tomorrow, if not I'll merge in the next few days.

murali-reddy:
LGTM as well. Sorry about the delay and thanks for the PR @bhcleek

bhcleek:
Thank you for the reviews @aauren and @murali-reddy 🙇

PR #1137, opened 2021-07-31
Add a default route 0.0.0.0/8 import policy deny rule

### LINKED ISSUE #584: Default route from eBGP peer, pulls kernel default upon withdraw/shutdown

lucasmundim:
This pull request protects Kube-router from receiving the default route from a misconfigured switch, affecting the node's reachability.

fixes #584 and #586

### DIFF
--- a/pkg/controllers/routing/bgp_policies.go
+++ b/pkg/controllers/routing/bgp_policies.go
@@ -32,6 +32,11 @@ func (nrc *NetworkRoutingController) AddPolicies() error {
 		klog.Errorf("Failed to add `servicevipsdefinedset` defined set: %s", err)
 	}
 
+	err = nrc.addDefaultRouteDefinedSet()
+	if err != nil {
+		klog.Errorf("Failed to add `defaultroutedefinedset` defined set: %s", err)
+	}
+
 	iBGPPeerCIDRs, err := nrc.addiBGPPeersDefinedSet()
 	if err != nil {
 		klog.Errorf("Failed to add `iBGPpeerset` defined set: %s", err)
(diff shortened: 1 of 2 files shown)


aauren (commented):
Code wise, this looks fine to me. Have you had a chance to test this both in a regular cluster and with a peer advertising `0.0.0.0/0`?

Also fixes #586

lucasmundim:
> Code wise, this looks fine to me. Have you had a chance to test this both in a regular cluster and with a peer advertising `0.0.0.0/0`?

Yes, we are running a fork with this for about six months in our production cluster. We had set up a test for this with a peer advertising 0.0.0.0/0, first with the original version to ensure that it breaks, and then with the forked one and confirm that it works and does not introduce other effects you pointed out on the slack channel.
```
from: @aauren 
> You’d have to test it to make sure that gobgp doesn’t incorrectly interpret 0.0.0.0/0 to include all possible subnets inside of that, but rather that one specifically. i.e. it might interpret a policy that has 0.0.0.0/0 to also include something like 172.16.0.1/32 which would be bad.
```

We confirmed that this does not happen.

aauren (approved):
@lucasmundim thanks for submitting and giving context. LGTM!

I'll leave open for another day or so to give @murali-reddy a chance to take a look, and then merge.

PR #1145, opened 2021-08-02
Fix containerd CVE-2021-32760

aauren:
FYI @mrueg / @murali-reddy

* Removes usages of <containerd-1.5.4 as it suffers from the unix file permissions problem. We don't use this portion of containerd, so I don't think that its super pressing, but we should fix none-the-less
* Updates go.mod to refer to Go 1.16 which is what we currently compile with


go.mod
```diff
@@ -2,10 +2,11 @@ module github.com/cloudnativelabs/kube-router
 
 require (
 	github.com/aws/aws-sdk-go v1.38.70
-	github.com/containerd/containerd v1.5.2 // indirect
+	github.com/containerd/containerd v1.5.4 // indirect
```
mrueg, line 5: 1.5.5 is out, do you want to update it to this version?

go.mod
```diff
@@ -30,4 +31,6 @@ require (
 	k8s.io/klog/v2 v2.9.0
 )
 
-go 1.13
+replace github.com/containerd/containerd => github.com/containerd/containerd v1.5.4 // CVE-2021-32760
```
mrueg, line 34: I wonder if this replacement is needed?

aauren: It appears to be needed, without this, all of the old containerd versions come back into `go.sum`

go.mod
aauren: Updated to 1.5.5

go.mod
```diff
@@ -2,13 +2,14 @@ module github.com/cloudnativelabs/kube-router
 
 require (
 	github.com/aws/aws-sdk-go v1.38.70
-	github.com/containerd/containerd v1.5.2 // indirect
+	github.com/containerd/containerd v1.5.5 // indirect
```
mrueg, line 5: It seems like your suggestions with 1.5.4 is the better choice. 1.5.5 updates protobuf which breaks linting as it deprecated some functions we use and you probably don' t want to update that one in this change.

aauren: reverted

go.mod
```diff
@@ -2,13 +2,14 @@ module github.com/cloudnativelabs/kube-router
...
-	github.com/containerd/containerd v1.5.2 // indirect
+	github.com/containerd/containerd v1.5.4 // indirect
 	github.com/containernetworking/cni v0.8.1
 	github.com/containernetworking/plugins v0.9.1
 	github.com/coreos/go-iptables v0.6.0
+	github.com/docker/distribution v2.7.1+incompatible // indirect
 	github.com/docker/docker v20.10.7+incompatible
 	github.com/docker/go-connections v0.4.0 // indirect
-	github.com/golang/protobuf v1.4.3
+	github.com/golang/protobuf v1.5.0
```
mrueg, line 12: You will need to revert this back to 1.4.3 otherwise the linting will fail.

aauren: Doh! Alright. I think I have it now, and ran `make lint` manually on my desktop before pushing this.

mrueg (approved):

PR #1153, opened 2021-08-13
feat(bgp): add custom BGP import rejection policy support via node annotation

### LINKED ISSUE #586: eBGP import policy options
Where a kube-router has eBGP sessions with upstream ToR, a goBGP Export policy exists, but no Import policy.  This results in a default accept all.

In most cases, the ToR would be configured to deny sending anything.  However, if a mistake were to happen and the ToR starts sending BGP updates, kube-router has no control to do anything other than accept and install these, unless it has been configured after deploy.

## Suggested enhancement behaviour
- Import policy which denies all from eBGP neighbors (default false)
- boolean `--peer-router-deny-in`

The Import policy created on deploy doesn't have to be a complex set of policies.  I would imagine most deployments would want to deny routes from eBGP neighbors by default anyway, because I expect most deployments rely on the nodes routing table for things reachable outside the cluster, and iBGP for things inside the cluster.

## Use case
Helps avoid situations like in https://github.com/cloudnativelabs/kube-router/issues/584

lucasmundim:
fixes #586 adding another option for protecting required routes.

### DIFF
--- a/docs/bgp.md
+++ b/docs/bgp.md
@@ -173,6 +173,19 @@ kubectl annotate node <kube-node> "kube-router.io/node.bgp.communities=65535:652
 kubectl annotate node <kube-node> "kube-router.io/node.bgp.communities=no-export"
 ```
 
+### Custom BGP Import Policy Reject
+
+Kube-router accepts by default all routes advertised by it's neighbors.
+
+If the bgp session with one neighbor dies, gobgp deletes all routes received by it.
+
+If one of the received routes is needed for this node to function properly (eg: custom static route), it could stop working.
+
+In the following example we add custom prefixes that'll be set on a custom import policy reject rule via annotation, protecting the node from losing required routes:
+```
+kubectl annotate node <kube-node> "kube-router.io/node.bgp.customimportreject=10.0.0.0/16, 192.168.1.0/24"
+```
+
 ## BGP listen address list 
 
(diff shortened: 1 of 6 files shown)


aauren:
Thanks for submitting this @lucasmundim. We're still focusing on bug fixes for the 1.3 release for a bit, but as we look towards features for 1.4 we'll take a look then. I've added it to the 1.4 project that we use to sort features and requests for the next release.

aauren:
@lucasmundim Thanks for this work! Apologies that this took so long to get back to!

Everything looked pretty much spot on and I think that this fits well with the intention behind the bug fix reported in #586. I've gone ahead and rebased it and pushed it back to your branch.

Additionally, in my testing, I found that it wasn't blocking subnets of the ones defined in `customimportreject`. For instance, if someone added the subnet `10.100.0.0/16` to the annotation, it would block `10.100.0.0/16`, but it would allow things like `10.100.1.0/24` or `10.100.1.1/32`.

I don't think that was your intention, so I added another commit that rejects all subnets of subnets that are defined in the annotation.

If you can, please take a look and let me know what you think. Assuming you're good with the change, we can merge this.

lucasmundim:
@aauren Hi, I agree with your suggestion as this should be the expected behavior. Thanks.

aauren:
Thanks!

PR #1165, opened 2021-09-01
Makefile: Update golangci-lint to 1.42.0

mrueg:
Update to a more recent version of golangci-lint and see if it caught more linting issues.


aauren:
It looks like we need to change something else with the linting process. From the job:
```
level=warning msg="[runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive."
```

Also, I need a rebase after the gobgp PR was merged.

Makefile
```diff
@@ -45,7 +45,7 @@ FILE_ARCH=x86-64
 DOCKER_ARCH=
 endif
 $(info Building for GOARCH=$(GOARCH))
-all: test kube-router container ## Default target. Runs tests, builds binaries and images.
+all: lint test kube-router container ## Default target. Runs tests, builds binaries and images.
```
mrueg, line 48:
```suggestion
all: lint test kube-router container ## Default target. Runs lints code, runs tests, builds binaries and images.
```

Makefile
```diff
@@ -18,7 +18,7 @@ MAKEFILE_DIR=$(dir $(realpath $(firstword $(MAKEFILE_LIST))))
 UPSTREAM_IMPORT_PATH=$(GOPATH)/src/github.com/cloudnativelabs/kube-router/
 BUILD_IN_DOCKER?=true
 DOCKER_BUILD_IMAGE?=golang:1.16.7-alpine3.14
-DOCKER_LINT_IMAGE?=golangci/golangci-lint:v1.27.0
+DOCKER_LINT_IMAGE?=golangci/golangci-lint:v1.42.0
```
mrueg, line 21:
```suggestion
DOCKER_LINT_IMAGE?=golangci/golangci-lint:v1.42.1
```

pkg/controllers/netpol/policy.go
```diff
@@ -164,129 +158,124 @@ func (npc *NetworkPolicyController) processIngressRules(policy networkPolicyInfo
...
+				if err := npc.appendRuleToPolicyChain(policyChainName, comment, "", namedPortIPSetName,
+					eps.protocol, eps.port, eps.endport); err != nil {
 					return err
 				}
 			}
 		}
 
-		// case where nether ports nor from details are speified in the ingress rule
-		// so match on all ports, protocol, source IP's
+		// case where nether ports nor from details are specified in the ingress rule so match on all ports, protocol,
```
mrueg, line 234:
```suggestion
		// case where neither ports nor from details are specified in the ingress rule so match on all ports, protocol,
```

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -1055,287 +1088,68 @@ func (ln *linuxNetworking) prepareEndpointForDsr(containerID string, endpointIP
 
 	hostNetworkNamespaceHandle, err := netns.Get()
 	if err != nil {
-		return errors.New("Failed to get namespace due to " + err.Error())
+		return fmt.Errorf("failed to get namespace due to %v", err)
```
mrueg, line 1091: Why not create a new error object here?

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -1055,287 +1088,68 @@ func (ln *linuxNetworking) prepareEndpointForDsr(containerID string, endpointIP
...
-	return nil
+	return ln.configureContainerForDSR(vip, endpointIP, containerID, pid, hostNetworkNamespaceHandle)
 }
 
 // The same as prepareEndpointForDsr but using CRI instead of docker.
-func (ln *linuxNetworking) prepareEndpointForDsrWithCRI(runtimeEndpoint, containerID, endpointIP, vip string) (err error) {
+func (ln *linuxNetworking) prepareEndpointForDsrWithCRI(runtimeEndpoint, containerID, endpointIP, vip string) error {
 
-	// FIXME: its possible switch namespaces may never work safely in GO without hacks.
+	// Its possible switch namespaces may never work safely in GO without hacks.
```
mrueg, line 1120:
```suggestion
	// It's possible switch namespaces may never work safely in GO without hacks.
```

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -1960,23 +1784,26 @@ func (ln *linuxNetworking) ipvsAddService(svcs []*ipvs.Service, vip net.IP, prot
...
 	h := fnv.New32a()
 	_, err := h.Write([]byte(ip + "-" + protocol + "-" + port))
 	if err != nil {
 		return 0, err
 	}
-	return h.Sum32() & 0x3FFF, err
+	return h.Sum32() & maxFwMarkBitSize, err
 }
 
 // ipvsAddFWMarkService: creates a IPVS service using FWMARK
```
mrueg, line 1796:
```suggestion
// ipvsAddFWMarkService: creates an IPVS service using FWMARK
```

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -2257,7 +2093,8 @@ func (ln *linuxNetworking) setupRoutesForExternalIPForDSR(serviceInfoMap service
 		for _, externalIP := range svc.externalIPs {
 			// Verify the DSR annotation exists
 			if !svc.directServerReturn {
-				klog.V(1).Infof("Skipping service %s/%s as it does not have DSR annotation\n", svc.namespace, svc.name)
+				klog.V(1).Infof("Skipping service %s/%s as it does not have DSR annotation\n",
```
mrueg, line 2096: Do we need the newline at the end of the string here?

pkg/controllers/proxy/utils.go
```diff
@@ -0,0 +1,158 @@
...
+		return fmt.Errorf("failed to get activeNetworkNamespace due to %v", err)
+	}
+	klog.V(2).Infof("Current network namespace after netns. Set to container network namespace: %s",
+		activeNetworkNamespaceHandle.String())
+	_ = activeNetworkNamespaceHandle.Close()
+
+	// TODO: fix boilerplate `netns.Set(hostNetworkNamespaceHandle)` code. Need a robust
+	// way to switch back to old namespace, pretty much all things will go wrong if we dont switch back
+
+	// create a ipip tunnel interface inside the endpoint container
```
mrueg, line 63:
```suggestion
	// create an ipip tunnel interface inside the endpoint container
```

pkg/controllers/proxy/utils.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// disable rp_filter on all interface
+	err = ioutil.WriteFile("/proc/sys/net/ipv4/conf/kube-tunnel-if/rp_filter",
+		[]byte(strconv.Itoa(0)), 0640)
+	if err != nil {
+		attemptNamespaceResetAfterError(hostNetworkNamespaceHandle)
+		return fmt.Errorf("failed to disable rp_filter on kube-tunnel-if in the endpoint container")
+	}
+
+	err = ioutil.WriteFile("/proc/sys/net/ipv4/conf/eth0/rp_filter", []byte(strconv.Itoa(0)), 0640)
```
mrueg, line 132: Is this assuming eth0 always exists?

mrueg: Should those paths be extracted into variables?

pkg/controllers/routing/bgp_peers.go
```diff
@@ -303,7 +300,8 @@ func (nrc *NetworkRoutingController) newNodeEventHandler() cache.ResourceEventHa
 			node := obj.(*v1core.Node)
 			nodeIP, err := utils.GetNodeIP(node)
 			if err != nil {
-				klog.Errorf("New node received, but we were unable to add it as we were couldn't find it's node IP: %v", err)
+				klog.Errorf(
+					"New node received, but we were unable to add it as we were couldn't find it's node IP: %v", err)
```
mrueg, line 304:
```suggestion
					"New node received, but we were unable to add it as we were couldn't find its node IP: %v", err)
```

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -230,14 +248,20 @@ func (nrc *NetworkRoutingController) Run(healthChan chan<- *healthcheck.Controll
...
 	// enable netfilter for the bridge
 	if _, err := exec.Command("modprobe", "br_netfilter").CombinedOutput(); err != nil {
-		klog.Errorf("Failed to enable netfilter for bridge. Network policies and service proxy may not work: %s", err.Error())
+		klog.Errorf("Failed to enable netfilter for bridge. Network policies and service proxy may "+
+			"not work: %s", err.Error())
 	}
-	if err = ioutil.WriteFile("/proc/sys/net/bridge/bridge-nf-call-iptables", []byte(strconv.Itoa(1)), 0640); err != nil {
-		klog.Errorf("Failed to enable iptables for bridge. Network policies and service proxy may not work: %s", err.Error())
+	if err = ioutil.WriteFile(
+		"/proc/sys/net/bridge/bridge-nf-call-iptables", []byte(strconv.Itoa(1)), 0640); err != nil {
```
mrueg, line 255: should those paths be extracted?

pkg/controllers/routing/network_routes_controller_test.go
```diff
@@ -1709,10 +1711,8 @@ func Test_routeReflectorConfiguration(t *testing.T) {
 				if testcase.expectedClusterID != testcase.nrc.bgpClusterID {
 					t.Errorf("Node suppose to have cluster id '%s' but got %s", testcase.expectedClusterID, testcase.nrc.bgpClusterID)
 				}
-			} else {
-				if err == nil {
-					t.Fatal("misconfigured BGP server not suppose to start")
-				}
+			} else if err == nil {
+				t.Fatal("misconfigured BGP server not suppose to start")
```
mrueg, line 1715:
```suggestion
				t.Fatal("misconfigured BGP server is not supposed to start")
```

pkg/controllers/proxy/utils.go
aauren: Good catch on this one! That seems like it would be a long-standing bug.

aauren: Actually, it turns out that all of this happens from inside the container. In this case the interface inside the container is always `eth0`. At this point, I'm not entirely sure what chooses this, but if I read the CNI spec correctly, it gets specified by the component that calls the CNI plugin to setup the device (https://www.cni.dev/docs/spec/#add-add-container-to-network-or-apply-modifications), which in the case of Kubernetes would be the `kubelet`.

However, I can't seem to find out what component of the kubelet chooses this name or how it is set. I've reached out to the SIG-NET Slack channel to see if they have an answer. After ~30 some min searching through the k8s code base, I can't seem to find it.

pkg/controllers/proxy/network_services_controller.go
aauren: `fmt.Errorf()` actually returns a new error object, it just makes formatting a bit easier and more intuitive. Furthermore, when I eventually enable the `errorlint` linter and change all of these `%v` references it will automatically handle error wrapping which is the correct way to stack errors in GoLang so that they are comparable.

pkg/controllers/proxy/utils.go
aauren: I have added a TODO here to come back to this one. @thockin very helpfully provided in Slack that this is, in fact, determined by the container runtime. So the internal interface name itself could shift at some point in the future.

It looks like we can count on docker and containerd (https://github.com/containerd/containerd/blob/cf600abecc27200a3c0e1415cd1f6c325eb05edf/pkg/cri/server/helpers.go#L89) at least to name this eth0 for now, but this could shift at any point in the future or with a different runtime.

pkg/controllers/routing/network_routes_controller.go
aauren: I found that there was, in fact, a utils method responsible for doing this. I have updated all instances of proc setting to call into that utils method for consistency and consolidated all of the proc paths there as constants as well.

aauren:
All review comments have been fixed. Thanks for the review @mrueg!

Merging as this is a pretty complex refactor and I don't want to have to keep rebasing as master drifts.

PR #1166, opened 2021-09-06
only enable nf_call_ip{,6}tables for kube-bridge

sysedwinistrator:
The bridge CNI plugin requires traffic traversing its bridge interface to be passed to ip{,6}tables. kube-router currently ensures this by setting the sysctl `net.bridge.bridge-nf-call-ip{,6}tables=1` (the kernel's default), which forces traffic traversing **any** bridge on the node to be filtered through ip{,6}tables.

That sysctl is disabled by default on some distros because it leads to issues with VMs and other applications which expect bridges (other than `kube-bridge` of cause) to be just plain L2 bridges. The per-bridge setting in `/sys/class/net/other-bridge/bridge/nf_call_ip{,6}tables` will also be ignored when the sysctl is enabled.
However, when the sysctl is disabled, the filtering can still be enabled per bridge. I propose that kube-router enables `nf-call-ip{,6}tables` only for the `kube-bridge` interface and does not touch the sysctl anymore. The per-bridge setting will be applied if it's disabled; and if it's enabled, we don't need to do anything anyway.

A limitation of the proposed solution is that the `kube-bridge` interface has to exist for kube-router to be able to configure it. If kube-router fails to create `kube-bridge`, it leaves the creation to the bridge CNI plugin, which does not handle this setting and relies on the sysctl already being enabled. So in the case where the sysctl is disabled and kube-router somehow fails to create the `kube-bridge` interface, one would end up with an incorrectly configured `kube-bridge`.

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -232,11 +232,11 @@ func (nrc *NetworkRoutingController) Run(healthChan chan<- *healthcheck.Controll
 	if _, err := exec.Command("modprobe", "br_netfilter").CombinedOutput(); err != nil {
 		klog.Errorf("Failed to enable netfilter for bridge. Network policies and service proxy may not work: %s", err.Error())
 	}
-	if err = ioutil.WriteFile("/proc/sys/net/bridge/bridge-nf-call-iptables", []byte(strconv.Itoa(1)), 0640); err != nil {
+	if err = ioutil.WriteFile("/sys/class/net/kube-bridge/bridge/nf_call_iptables", []byte(strconv.Itoa(1)), 0644); err != nil {
 		klog.Errorf("Failed to enable iptables for bridge. Network policies and service proxy may not work: %s", err.Error())
 	}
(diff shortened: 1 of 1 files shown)


aauren:
I think I agree with the sentiment that we should be trying to only control our specific bridge. However, I think that we should check if the sysfs path for `kube-bridge` exists first. If it does, then set it. If it does not, then we should revert to the current /proc/sys so that the project follows the principle of least surprise. This would also mitigate the condition you mentioned at the end as well.

In a recent PR I consolidated all of the `/proc/sys` handling to be handled using a common function set so that handling was similar in all of kube-router's controllers: https://github.com/cloudnativelabs/kube-router/blob/master/pkg/utils/sysctl.go This also allows us to better track which paths of `/proc/sys` we're changing.

We should probably just extend this to handle sysfs as well and then fold this functionality into similar methods there. Are you able to rebase your work on https://github.com/cloudnativelabs/kube-router/pull/1168 and then implement the functionality described above? ^^^

aauren:
Closing due to lack of response.

PR #1168, opened 2021-09-11
fix(sysctl): revert is fatal check for some conditions

aauren:
@mrueg @murali-reddy 

I was a little quick on this one. I added a check for `IsFatal()` (which returns false when the path doesn't exist) to some of the error checks for `utils.SetSysctl()`. However, for some of the conditions, this just stops us from logging messages (as the error handling doesn't results in only a logging statement), which isn't helpful.

For the cases of DSR, we probably want the condition to be fatal even if `utils.SetSysctl()` isn't able to find the path (which was the previous behavior).


aauren:
Also, found one last hard-coded reference to `KUBE-ROUTER-HAIRPIN` that I removed along with some old TODOs that went along with the `KUBE-ROUTER-HAIRPIN` refactor work I did before.

mrueg (approved):

PR #1170, opened 2021-09-27
feat(pod.go): add conntrack invalid state drop

### LINKED ISSUE #709: Drop packets in INVALID state
Should we implement?

https://github.com/anfernee/kubernetes/commit/a07169bcadf3c35ff456c17bed970ccbca0963cb

https://github.com/kubernetes/kubernetes/issues/74839

@murali-reddy

aauren:
@mrueg @murali-reddy

This drops invalid state traffic for pods in order to fix #709

This is very similar to what Calico does here: https://github.com/projectcalico/felix/blob/b91f95da817c28385498afb6679e9588ea1ca984/rules/endpoints.go#L495

And seems to be recommended by the netfilter folks in order to avoid an under-documented "feature" where timed out conntrack entries are no longer NAT'd and private IP space is leaked: https://bugzilla.netfilter.org/show_bug.cgi?id=693#c6

I took the clue from Calico and implemented it on a per-pod chain basis so that we only affect our own traffic. Additionally, I also used their ordering, so that this rule will come in directly after accepting `RELATED` and `ESTABLISHED` traffic (note logic is reversed, each subsequent statement is inserted at the top of the chain).

For now I made this DROP unconditional, which is different than Calico. However, I believe that after looking through their issues the only reason why they make it conditional, is because they also allow users to add `doNotTrack` to certain pods/nodes which then bypasses conntrack completely and will result in all states for those flows as being `INVALID`. As such, I don't think that we need to wrap this in a condition yet. I added a comment on there to say as much for future posterity.

FYI @NaPs


aauren:
I deployed this to a very small local test environment and it looks to work fine.

This is what the rules look like:
```
Chain KUBE-POD-FW-MD35BSM33OICKD3E (7 references)
    0     0 ACCEPT     all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* rule for stateful firewall for pod */ ctstate RELATED,ESTABLISHED
    0     0 DROP       all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* rule to drop invalid state for pod */ ctstate INVALID
    0     0 ACCEPT     all  --  *      *       0.0.0.0/0            172.20.1.14          /* rule to permit the traffic traffic to pods when source is the pod's local node */ ADDRTYPE match src-type LOCAL
    0     0 KUBE-NWPLCY-DEFAULT  all  --  *      *       172.20.1.14          0.0.0.0/0            /* run through default egress network policy  chain */
    0     0 KUBE-NWPLCY-DEFAULT  all  --  *      *       0.0.0.0/0            172.20.1.14          /* run through default ingress network policy  chain */
    0     0 NFLOG      all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* rule to log dropped traffic POD name:toolbox-8dck6 namespace: development */ mark match ! 0x10000/0x10000 limit: avg 10/min burst 10 nflog-group 100
    0     0 REJECT     all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* rule to REJECT traffic destined for POD name:toolbox-8dck6 namespace: development */ mark match ! 0x10000/0x10000 reject-with icmp-port-unreachable
    0     0 MARK       all  --  *      *       0.0.0.0/0            0.0.0.0/0            MARK and 0xfffeffff
    0     0 MARK       all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* set mark to ACCEPT traffic that comply to network policies */ MARK or 0x20000
```

One of the limitations of this approach is that it would only work when the NetworkPolicy controller was activated for kube-router. i.e. people that put kube-router in only proxy or router mode would not have this change. However, I think that while it could be argued that this change is relevant to the proxy side as well, at the end of the day, this IS a policy change as we are dropping packets that violate a condition.

However, given that since #1108 these rules should be applied to every pod regardless of whether or not they contain a NetworkPolicy, as long as kube-router is in NetworkPolicy mode, this change should take effect.

aauren:
@murali-reddy any thoughts on this?

pkg/controllers/netpol/pod.go
```diff
@@ -190,6 +190,17 @@ func (npc *NetworkPolicyController) setupPodNetpolRules(pod podInfo, podFwChainN
 		"-m", "addrtype", "--src-type", "LOCAL", "-d", pod.ip, "-j", "ACCEPT", "\n"}
 	npc.filterTableRules.WriteString(strings.Join(args, " "))
 
+	// ensure statefull firewall drops INVALID state traffic from/to the pod
+	// For full context see: https://bugzilla.netfilter.org/show_bug.cgi?id=693
+	// The NAT engine ingores any packet with state INVALID, because there's no reliable way to determine what kind of
```
mrueg, line 195:
```suggestion
	// The NAT engine ignores any packet with state INVALID, because there's no reliable way to determine what kind of
```

mrueg (approved):

pkg/controllers/netpol/pod.go
aauren: Fixed! Thanks for the spelling catch. I wonder why our misspell linter didn't catch that though.

PR #1178, opened 2021-10-26
fix bug when adding ip rule for fwmark

kailunshi:
This fixes a bug when the string representation of the fwmark is contained by an existing ip rules. 

for example, if existing ip rules:
32764:	from all fwmark 0x8ef lookup 78
new fwmark to be added: 0x8e
result: the ip rule add command won't be executed.


murali-reddy:
@kailunshi thanks for finding the bug and submitting PR

LGTM

PR #1185, opened 2021-11-08
fix(bgp): use PeerState_ESTABLISHED logic like function name(#1184)

### LINKED ISSUE #1184: isPeerEstablished function is a bug
**What happened?**
isPeerEstablished function is a bug, it has the opposite result when call this function https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/routing/network_routes_controller.go#L552-L556 , isPeerEstablished function is like https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/routing/network_routes_controller.go#L620-L637:

```go
func (nrc *NetworkRoutingController) isPeerEstablished(peerIP string) (bool, error) {
        var peerDisconnected bool
	peerFunc := func(peer *gobgpapi.Peer) {
		if peer.Conf.NeighborAddress == peerIP {
			if peer.State.SessionState != gobgpapi.PeerState_ESTABLISHED {
				peerDisconnected = true
			}
		}
	}
	err := nrc.bgpServer.ListPeer(context.Background(), &gobgpapi.ListPeerRequest{
		Address: peerIP,
	}, peerFunc)
	if err != nil {
		return false, fmt.Errorf("unable to list peers to see if tunnel & routes need to be removed: %v", err)
	}
	return peerDisconnected, nil
}

```

should return peerConnected logic like the function name. fix in pr  https://github.com/cloudnativelabs/kube-router/pull/1185

lx1036:
fix #1184

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -618,12 +618,10 @@ func (nrc *NetworkRoutingController) injectRoute(path *gobgpapi.Path) error {
 }
 
 func (nrc *NetworkRoutingController) isPeerEstablished(peerIP string) (bool, error) {
-	var peerDisconnected bool
+	var peerConnected bool
 	peerFunc := func(peer *gobgpapi.Peer) {
-		if peer.Conf.NeighborAddress == peerIP {
-			if peer.State.SessionState != gobgpapi.PeerState_ESTABLISHED {
-				peerDisconnected = true
-			}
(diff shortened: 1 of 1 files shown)


aauren:
@lx1036 Good catch on this one!

What's odd is that I swear that we had done some pretty extensive testing on this in our labs. But I just put it through its paces again and can definitely confirm that the previous logic did not clean up tunnels the way it was supposed to. Your logic implements the function correctly.

Thanks for contributing!

PR #1186, opened 2021-11-09
fix(ecmp vip): simplify nodeHasEndpointsForService() code

lx1036:
make code more readable in nodeHasEndpointsForService() func.

### DIFF
--- a/pkg/controllers/routing/ecmp_vip.go
+++ b/pkg/controllers/routing/ecmp_vip.go
@@ -516,14 +516,8 @@ func (nrc *NetworkRoutingController) nodeHasEndpointsForService(svc *v1core.Serv
 
 	for _, subset := range ep.Subsets {
 		for _, address := range subset.Addresses {
-			if address.NodeName != nil {
-				if *address.NodeName == nrc.nodeName {
-					return true, nil
-				}
-			} else {
-				if address.IP == nrc.nodeIP.String() {
-					return true, nil
-				}
+			if (address.NodeName != nil && *address.NodeName == nrc.nodeName) || address.IP == nrc.nodeIP.String() {
+				return true, nil
 			}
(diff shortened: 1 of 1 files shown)


mrueg (approved):

pkg/controllers/routing/ecmp_vip.go
```diff
@@ -516,14 +516,8 @@ func (nrc *NetworkRoutingController) nodeHasEndpointsForService(svc *v1core.Serv
...
 		for _, address := range subset.Addresses {
-			if address.NodeName != nil {
-				if *address.NodeName == nrc.nodeName {
-					return true, nil
-				}
-			} else {
-				if address.IP == nrc.nodeIP.String() {
-					return true, nil
-				}
+			if (address.NodeName != nil && *address.NodeName == nrc.nodeName) || address.IP == nrc.nodeIP.String() {
```
aauren, line 519: These two sets of logic aren't logically the same.

If `address.NodeName != nil` and `address.NodeName != nrc.nodeName` logic will continue without ever processing if `address.IP == nrc.nodeIP.String()`.

So with the original logic, if the values were set to:
```
address.NodeName := foo
nrc.nodeName := bar
address.IP := 10.0.0.1
nrc.nodeIP := 10.0.0.1
```
The end result would be `return false, nil`

vs. with your logic it would match the `||` condition and `return true, nil`

aauren (changes requested):

pkg/controllers/routing/ecmp_vip.go
lx1036: yes, you are right. these two sets of logic are not the same. I will see if close the pr later.

aauren:
Closing because of lack of follow up.

PR #1187, opened 2021-11-09
fix(service type): use k8s service type constant

lx1036:
use k8s service type raw constant instead of custom constant.


aauren:
Closing because of lack of follow up.

Issue #1188, opened 2021-11-10
[bgp policy]compare bgp defined set which is not sorted


lx1036:
**What happened?**
reflect.DeepEqual two []string, which is not sorted, like in https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/routing/bgp_policies.go#L246-L248 , may not happen as expected:

```
func (nrc *NetworkRoutingController) addiBGPPeersDefinedSet() ([]string, error) {
	iBGPPeerCIDRs := make([]string, 0)
	if !nrc.bgpEnableInternal {
		return iBGPPeerCIDRs, nil
	}

	// Get the current list of the nodes from the local cache
	nodes := nrc.nodeLister.List()
	for _, node := range nodes {
		nodeObj := node.(*v1core.Node)
		nodeIP, err := utils.GetNodeIP(nodeObj)
		if err != nil {
			klog.Errorf("Failed to find a node IP and therefore cannot add internal BGP Peer: %v", err)
			continue
		}
		iBGPPeerCIDRs = append(iBGPPeerCIDRs, nodeIP.String()+"/32")
	}

	var currentDefinedSet *gobgpapi.DefinedSet
	err := nrc.bgpServer.ListDefinedSet(context.Background(),
		&gobgpapi.ListDefinedSetRequest{DefinedType: gobgpapi.DefinedType_NEIGHBOR, Name: "iBGPpeerset"},
		func(ds *gobgpapi.DefinedSet) {
			currentDefinedSet = ds
		})
	if err != nil {
		return iBGPPeerCIDRs, err
	}
	if currentDefinedSet == nil {
		iBGPPeerNS := &gobgpapi.DefinedSet{
			DefinedType: gobgpapi.DefinedType_NEIGHBOR,
			Name:        "iBGPpeerset",
			List:        iBGPPeerCIDRs,
		}
		err = nrc.bgpServer.AddDefinedSet(context.Background(), &gobgpapi.AddDefinedSetRequest{DefinedSet: iBGPPeerNS})
		return iBGPPeerCIDRs, err
	}

	if reflect.DeepEqual(iBGPPeerCIDRs, currentDefinedSet.List) {
		return iBGPPeerCIDRs, nil
	}

```

reflect.DeepEqual([]string{"192.168.0.0", "192.169.0.0"}, []string{"192.169.0.0", "192.168.0.0"}) is not true, is false. the slice should be sorted.

[...]

PR #1189, opened 2021-11-10
fix(bgp policy): sort the slice items before deep equal(#1188)

lx1036:


aauren (approved):
LGTM

Thanks for the contribution!

PR #1198, opened 2021-12-01
Update Security Deps

aauren:
@murali-reddy / @mrueg 

This would have to be merged after #1101 is merged (and I have a chance to rebase). But this should update the deps for containerd and image-spec which have security vulnerabilities associated with them. I don't believe that either of the vulns affect us, but better to just update anyway.


go.sum
```diff
@@ -96,6 +101,8 @@ github.com/containerd/console v1.0.1/go.mod h1:XUsP6YE/mKtz6bxc+I8UiKKTP04qjQL4q
 github.com/containerd/console v1.0.2/go.mod h1:ytZPjGgY2oeTkAONYafi2kSj0aYggsf8acV1PGKCbzQ=
 github.com/containerd/containerd v1.5.4 h1:uPF0og3ByFzDnaStfiQj3fVGTEtaSNyU+bW7GR/nqGA=
```
mrueg, line 102: I wonder if this one should be removed here somehow?

aauren: Fixed! Thanks for the feedback.

aauren:
I noticed while I was doing a long running test of this build in one of our clusters that there was an enormous amount of logs saying stuff like:
```
E1202 15:40:14.701298   18525 service_endpoints_sync.go:373] failed to find FW mark for the service: no key matching tcp:10.0.0.1:8080 was found in fwMarkMap
```

Looking into this, I found that there was an error condition that shouldn't have been treated as an error condition. So I have fixed that in this PR as well.

PR #1208, opened 2021-12-06
feat(nsc): only hairpin endpoints on local node

### LINKED ISSUE #1196: kube-router-hairpin in nat table cannot be updated in real time
当我删除一个pod然后重新拉起后，nat 表中的 KUBE-ROUTER-HAIRPIN 链会删除原来pod和svc的条目，但不会生成新的 pod和svc 的规则；除非iptables -t nat -F 后，systemctl restart kube-router 重启服务后会生成新的规则，但是我们希望它能实时动态更新Chain KUBE-ROUTER-HAIRPIN   

kubernetes:1.22.4    kube-router:1.3.2      --hairpin-mode=true     hairpinMode:true

aauren:
Fixes #1196 (Addendum)

@murali-reddy @mrueg 

We should only add hairpinning rules for pods that are local to the endpoint.

Given that the definition of hairpinning is when a pod talks to itself via its service endpoint, these flows must be local in order to be valid. However, the way we implement hairpinning right now is that we add all pods and all endpoints to the `KUBE-ROUTER-HAIRPIN` chain, even ones that don't have a local endpoint.

This creates a needless number of execs since these are still handled via single iptables rules. It also creates additional time that the kernel processes rules in the data path.

When I deployed this change to our test cluster with ~300 pods running on it to a node with relatively few pods running on it, the number of hairpin rules went from ~350 to ~5.

FYI @z461948964

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -1353,7 +1353,18 @@ func (nsc *NetworkServicesController) syncHairpinIptablesRules() error {
 	// Generate the rules that we need
 	for svcName, svcInfo := range nsc.serviceMap {
 		if nsc.globalHairpin || svcInfo.hairpin {
+			// If this service doesn't have any active & local endpoints on this node, then skip it as only local
+			// endpoints matter for hairpinning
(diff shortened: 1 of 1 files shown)


mrueg (approved):

murali-reddy:
LGTM

PR #1216, opened 2021-12-09
Remove Duplicate Iptables Sync Metrics

aauren:
@mrueg @murali-reddy 

@MikeSpreitzer pointed out that these metrics are already present in the histogram type as *_count and *_sum and these two added metrics just add duplicates. I've also proved out in my own environments that these metric values are identical to the ones already carried in the histogram.


mrueg (approved):

murali-reddy:
LGTM

PR #1224, opened 2021-12-17
Gobgp v3.0.0 Upgrade

aauren:
@mrueg @murali-reddy 

Adding this as a WIP right now because GoBGP 3.X is not an official release yet.

When it does release, it will carry with it breaking changes for several APIs, but notable ones are:
* `MonitorTable()` no longer exists and is replaced by the `WatchEvent()`
* Protobuf is upgraded from APIv1 to APIv2 (https://go.dev/blog/protobuf-apiv2).


aauren:
Removed the WIP - Going through final testing and then we should be able to merge this soon.

go.mod
```diff
@@ -6,18 +6,18 @@ require (
 	github.com/containernetworking/plugins v1.1.1
 	github.com/coreos/go-iptables v0.6.0
 	github.com/docker/docker v20.10.16+incompatible
-	github.com/golang/protobuf v1.5.2
 	github.com/moby/ipvs v1.0.1
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.19.0
-	github.com/osrg/gobgp v0.0.0-20211201041502-6248c576b118
+	github.com/osrg/gobgp/v3 v3.0.0-rc3
```
mrueg, line 12: Can you upgrade this to v3.2.0?

mrueg:
Can you update https://github.com/cloudnativelabs/kube-router/blob/master/Makefile#L22 as well with a matching version of gobgp?

go.mod
aauren: Done.

aauren:
> Can you update https://github.com/cloudnativelabs/kube-router/blob/master/Makefile#L22 as well with a matching version of gobgp?

@mrueg  Not sure if I did this correctly or not, since I didn't use a SHA and am not quite certain what SHA I would use. Maybe that's not needed now that it's on an official version? Check and see what you think.

go.mod
```diff
@@ -6,27 +6,29 @@ require (
...
 	github.com/stretchr/testify v1.7.1
 	github.com/vishvananda/netlink v1.1.1-0.20210330154013-f5de75959ad5
 	github.com/vishvananda/netns v0.0.0-20211101163701-50045581ed74
 	golang.org/x/net v0.0.0-20220526153639-5463443f8c37
 	google.golang.org/grpc v1.46.2
-	k8s.io/api v0.23.7
-	k8s.io/apimachinery v0.23.7
-	k8s.io/client-go v0.23.7
+	google.golang.org/protobuf v1.27.1
+	k8s.io/api v0.24.1
```
mrueg, line 21: Do we want to upgrade to 0.24.x here? If so, we should also update the cri-api dependency.

Makefile
```diff
@@ -185,9 +185,9 @@ gobgp:
 ifeq "$(BUILD_IN_DOCKER)" "true"
 	$(DOCKER) run -v $(PWD):/go/src/github.com/cloudnativelabs/kube-router -w /go/src/github.com/cloudnativelabs/kube-router $(DOCKER_BUILD_IMAGE) \
 	    sh -c \
-	    'CGO_ENABLED=0 GOARCH=$(GOARCH) GOOS=linux go install github.com/osrg/gobgp/cmd/gobgp@$(GOBGP_VERSION) && if [ ${GOARCH} != $$(go env GOHOSTARCH) ]; then PREFIX=linux_${GOARCH}; fi && cp $$(go env GOPATH)/bin/$${PREFIX}/gobgp .'
+	    'CGO_ENABLED=0 GOARCH=$(GOARCH) GOOS=linux go install github.com/osrg/gobgp/v3/cmd/gobgp@$(GOBGP_VERSION) && if [ ${GOARCH} != $$(go env GOHOSTARCH) ]; then PREFIX=linux_${GOARCH}; fi && cp $$(go env GOPATH)/bin/$${PREFIX}/gobgp .'
```
mrueg, line 188: Thanks, yes that's what I had in mind!

mrueg:
One small nitpick, then this would be good to go.

go.mod
aauren: This has been updated.

aauren:
Addressed review feedback and finished local testing.

PR #1232, opened 2022-01-14
change to account for internet headers also when setting MSS

noillir:


aauren:
@noillir Thanks for putting this together! I'll try to find a slot to test it out in the next week or so.

aauren:
Got a chance to test this, it looks good to me. 

For posterity the numbers in this PR are calculated like this:
```
|--- Overlay Net IPIP --- | --- DSR IPIP --- | --- TCP Headers ---| TCP content    |
| 20 bytes                | 20 bytes         | 40 bytes           | Max MSS size   |
```

The overlay network may or may not exist, but this is compensated for by the logic in `utils.GetMTUFromNodeIP()`. This logic stays the same regardless of whether or not `--auto-mtu` is set to `true` or `false`. Then the change in this PR removes another 60 bytes (DSR IPIP headers + TCP Headers) in order to find the correct TCP MSS size.

So in the end the TCP MSS will be one of the following:
* When overlay networking is disabled: `node IP interface MTU - 60 bytes`
* When overlay networking is enabled: `node IP interface MTU - 80 bytes`

The only thing that `--auto-mtu` does is adjust the MTU on the `kube-bridge` and pod interfaces for the pod IPIP network.

PR #1244, opened 2022-02-04
prometheus metrics: add option to specify listen address

jpiper:
In the situation that you have multiple interfaces/IP addresses, we want to be able to specify which one we want to expose the prometheus metrics on.


aauren (changes requested):
Thanks for submitting. This seems reasonable. There is one piece of feedback that needs implementing, and it would also be helpful if you could squash your commits as these are all related to the same functionality and should be one atomic commit.

jpiper:
Yes! Sorry I haven’t had the chance to reply, I should be able to look at your comments this week

aauren:
@jpiper Just checking in again, are you still interested in this PR?

aauren:
Closing due to lack of updates

PR #1249, opened 2022-02-13
Some ipv6 fixes

arianvp:
Decides to uses ip6tables and ipset based on whether the NodeIP is an ipv6 address.
Analogous to how NetworkRoutingController does it.

No support for dual stack yet

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -985,8 +985,13 @@ func (nrc *NetworkRoutingController) startBgpServer(grpcServer bool) error {
 	}
 
 	if grpcServer {
-		nrc.bgpServer = gobgp.NewBgpServer(
-			gobgp.GrpcListenAddress(nrc.nodeIP.String() + ":50051" + "," + "127.0.0.1:50051"))
+		var addr string
+		if nrc.isIpv6 {
+			addr = "[" + nrc.nodeIP.String() + "]:50051" + "," + "[::1]:50051"
+		} else {
+			addr = nrc.nodeIP.String() + ":50051" + "," + "127.0.0.1:50051"
+		}
+		nrc.bgpServer = gobgp.NewBgpServer(gobgp.GrpcListenAddress(addr))
 	} else {
 		nrc.bgpServer = gobgp.NewBgpServer()
 	}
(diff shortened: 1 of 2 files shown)


pkg/controllers/routing/network_routes_controller.go
```diff
@@ -985,8 +985,13 @@ func (nrc *NetworkRoutingController) startBgpServer(grpcServer bool) error {
 	}
 
 	if grpcServer {
-		nrc.bgpServer = gobgp.NewBgpServer(
-			gobgp.GrpcListenAddress(nrc.nodeIP.String() + ":50051" + "," + "127.0.0.1:50051"))
+		var addr string
+		if nrc.isIpv6 {
+			addr = "[" + nrc.nodeIP.String() + "]:50051" + "," + "[::1]:50051"
```
arianvp, line 990: Random question by the way. Why does this need to listen on `nodeIP` on the first place; and not just on loopback?  `nodeIP` is definitely publicly routable in ipv6 and i'm curious why we're having to spawn a publicly accessible grpc server. Do the different `kube-router` daemonsets connect to eachother over grpc or something?

arianvp:
Haven't tested it on my cluster yet but am in the process of. 

Understand the concern that simply breaking as soon as an ipv4 address is added is not ideal.

It would work for my usecase as I have an ipv6 only internal network and am planning on doing address translation at the egde (https://www.jool.mx/en/intro-xlat.html#siit-dc) .  Wondering if it would make sense to address translate the ipv4 addresses in NetworkPolicy rules to their NAT-mapped counterparts (e.g. `1.2.3.4` => `64:ff9b:1:1::1.2.3.4/96`) when in Ipv6 mode. That would at least make ipv4 rules work on an ipv6 only cluster if some form of 4-to-6 translation is in use. 

 dual-stack in my opinion is not the way to go for Ipv6 deployments and sometimes bit surprised how much Kubernetes focuses on making it work. one is way better off by making it a concern at the gateway of your network. But I understand it's something that needs to be supported as it is something a compliant network plugin is supposed to handle according to kubernetes.

Let me get some tests done (and maybe implement the address mapping from ipv4 to ipv6) and see if it works well and then we can decide whether this is a patch I should just carry myself or it's something worth adopting as a stopgap solution.  The address mapping support would still be neat even if dual-stack works in the future as it's the way to go for ipv6-only deployments.

aauren:
That seems good. Comment back when you think you have it how you want it.

aauren:
Replaced by #1386

PR #1151, opened 2021-08-13
Routes Synchronization Routine [Fixes #509]

Rusox89:
I've separated the synchronization of routes with the host from the event watching on MonitorTable

PR #1262, opened 2022-03-15
Add Route Sync Routine

### LINKED ISSUE #509: Route table gone forever after network interface link down and up
I just found a weird behavior of kube-router that cause an outage in my cluster.

There was an event (not confirmed by AWS) that cause an host interface link down and then up.
After that event, all nodes inside cluster (including master node) lost a route table and it cause all pods communication to be failed.

You can easily reproduce this by provision a cluster, deploy some application and then just run `if down eth0 && if up eth0` with root account in any node.

After that, check route table `ip route` and all routes gone. To resolve that I have to manually restart kube-router and wait a couple minute to see route table come back again.

Please help.
Thanks

Environment:
Cloud: AWS
Kubernetes: 1.9.3
OS: Debian GNU/Linux 9 (stretch) 4.9.0-5-amd64
Docker: 17.3.2
Kube-router: 0.1.0

aauren:
@murali-reddy @mrueg 

Replaces #1151 and largely based on the original commit @Rusox89 made in that PR. Thanks for your work there!

I just rebased his work and patched up some stuff to fix a few unaccounted use-cases and bring it a little more in line with the rest of the kube-router code base.

Fixes #509, which is a rare situation where routes will get lost after an interface is brought down and back up, by introducing a route synchronization goroutine. This keeps track of what the kernel routing table should look like and syncs it from time to time.

I have deployed this in a test cluster and tested taking routes in / out and starting / stopping peers. It acts correctly in all use-cases.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -64,7 +64,7 @@ Usage of kube-router:
       --health-port uint16                            Health check port, 0 = Disabled (default 20244)
   -h, --help                                          Print usage information.
       --hostname-override string                      Overrides the NodeName of the node. Set this if kube-router is unable to determine your NodeName automatically.
-      --injected-routes-sync-period duration          The delay between route table synchronizations  (e.g. '5s', '1m', '2h22m'). Must be greater than 0. (default 15s)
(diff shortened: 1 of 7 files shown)


aauren:
Gave this quite a bit of testing in my development cluster. Along with the unit tests, I think that there can be a reasonable amount of confidence. Merging.

PR #1281, opened 2022-04-15
change package name from v1core to corev1

lucming:


lucming:
k8s.io/api/core/v1 this package we often use the alias as corev1 instead of v1core,it's better to rename it.

aauren:
I'm not opposed to renaming it, but I'm also not in favor of changing it just because. Can you give more information about the benefit of this change? Specifically what projects benefit from this changing and why this effort is worthwhile?

lucming:
> I'm not opposed to renaming it, but I'm also not in favor of changing it just because. Can you give more information about the benefit of this change? Specifically what projects benefit from this changing and why this effort is worthwhile?

It's just a matter of naming conventions, we can often see ${group}${version} in the code from k8s instead of ${version}${group}.

mrueg (approved):

aauren:
Currently waiting on the work in #1280 and and #1224 to go in first. Those PRs were raised first and would require a lot of rebasing after incorporating this PR.

lucming:
> ok，i will push this pr,you can tell me after those pr merged

aauren:
@lucming sorry that I came back around to this so late.

A lot happened with that work that I previously linked. If you're still interested in updating this, I'm still open to accepting it, but it would need to be targeted at the `prep-v2.0` branch which is where we're staging the active work that we're doing.

PR #1305, opened 2022-05-26
fix(proxy): compare slice to simplify code

lx1036:


aauren (changes requested):
Hey @lx1036 thanks for attempting to clean up the code a bit! I definitely appreciate the effort!

Unfortunately, these two implementations are not guaranteed to be equivalent. Despite using SliceStable for the endpoints, the ports aren't guaranteed to be unique (and as a matter of fact are very much likely not to be unique) and the endpoints themselves are randomized within the map which means the relying on the existing ordering isn't possible: https://github.com/lx1036/kube-router-1/blob/patch-2/pkg/controllers/proxy/network_services_controller.go/#L1228

I'm going to close this for now, as I don't see a lot of ways to simplify this code, but if you come up with another idea on how to rework this function post a comment here and I'll reopen it.

PR #1352, opened 2022-08-23
logging ipset/iptables commands

juchem:
This PR logs `ipset` and `iptables` commands at log level 9.

I've recently encountered an issue with Kubernetes v1.24.0 and
`kube-router` v1.5.1 where the node's network goes bonkers as soon as
`NetworkPolicyController::syncNetworkPolicyChains` is executed (last log
line is "Syncing network policy chains took 48.780838ms"). After that,
`ssh` freezes and `kubelet` becomes inacessible.

Debugging this issue proved to be quite daunting, specially due to the
lack of insight on which network policies are being applied.

Adding these logs have helped the debugging process immensely so far.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 _output
 _cache
 vendor
+.*.sw?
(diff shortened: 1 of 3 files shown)


aauren:
Looks like one or both of the files needs to be gofmt'd. Otherwise looks good.

juchem:
@aauren git format applied

PR #1380, opened 2022-10-19
feat(Makefile): make local builds behave like CI

aauren:
@mrueg 

`RUNTIME_BASE` and `BUILDTIME_BASE` are both consumed in the Dockerfile and handle setting the base containers used for both building and running kube-router. These are handled differently in CI using GitHub actions than they are for local builds.

This updates the Makefile so that local builds behave the same way that CI based builds in GitHub are when it comes to these bases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,10 @@ MAKEFILE_DIR=$(dir $(realpath $(firstword $(MAKEFILE_LIST))))
 UPSTREAM_IMPORT_PATH=$(GOPATH)/src/github.com/cloudnativelabs/kube-router/
 BUILD_IN_DOCKER?=true
 DOCKER_BUILD_IMAGE?=golang:1.19.2-alpine3.16
+## These variables are used by the Dockerfile as the bases for building and creating the runtime container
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -91,7 +95,8 @@ container: kube-router gobgp multiarch-binverify ## Builds a Docker container im
 	    echo "Using qemu to build non-native container"; \
 	    $(DOCKER) run --rm --privileged $(QEMU_IMAGE) --reset -p yes; \
 	fi
-	$(DOCKER) build -t "$(REGISTRY_DEV):$(subst /,,$(IMG_TAG))" -f Dockerfile --build-arg ARCH="$(DOCKER_ARCH)" .
+	$(DOCKER) build -t "$(REGISTRY_DEV):$(subst /,,$(IMG_TAG))" -f Dockerfile --build-arg ARCH="$(DOCKER_ARCH)" . \
```
mrueg, line 98: Thanks, can you change the order of the build-args so they appear before the scope?

mrueg (changes requested):

mrueg (approved):

PR #1382, opened 2022-10-20
fix invalid MTU in CNI config file

tamihiro:
This PR prevents troubles in the following rare scenario.
When upgrading from pre v1.1.0 version, install-cni container in a pod writes NodeIP's link MTU (e.g. 65536) to the CNI config file even when it's larger than kube-bridge's max MTU (e.g. 65535).
Then the main container tries but fails to set kube-bridge's MTU to 65536, leaving kube-bridge's MTU unchanged (e.g. 1500).
But because the CNI config file (/etc/cni/net.d/10-kuberouter.conflist) keeps its MTU as 65536, and because the node's CRI uses this value, it fails to create a veth pair and consequently can no longer create a container.
This patch, in the above scenario, rewrites kube-bridge's MTU as 1500 in the CNI config file.


aauren (approved):
Thanks @tamihiro

PR #1384, opened 2022-10-20
.github: Update github actions

mrueg:
@aauren This should remove the warnings in CI and update the used actions to the latest version.

Context: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


aauren (approved):
Thanks @mrueg!

PR #1385, opened 2022-10-21
.github: Allow tagging RCs without updating :latest

mrueg:
@aauren I believe this should allow you to tag release candidates without updating :latest


aauren (approved):
Thanks @mrueg!

Issue #1371, opened 2022-10-11
Mismatched local address when peering with external router on difference network


pmint93:
**What happened?**

kube-router v1.5 did not peer with external router on difference network (`Mismatched local address`)
My cluster setup has worker nodes join 2 difference networks (see figure bellow) and kube-router was configured to peer with external router https://github.com/cloudnativelabs/kube-router/blob/master/docs/bgp.md#peering-outside-the-cluster

**What did you expect to happen?**

Peer state is Established

**How can we reproduce the behavior you experienced?**

1. Step 1: Setup k8s cluster with workers join 2 difference networks as diagram bellow
2. Step 2: Install kube-router with params `--peer-router-ips` `--peer-router-asns` to peer with external router
3. Step 3: Check BGP neighbors on kube-router `gobgp neighbor` to see that the peer state did not Established

**Screenshots / Architecture Diagrams / Network Topologies**

```
                      ----|-----------------|--------------- network A (10.21.136.0/24)
                      workerA             workerB                 
--------|-----------------|-----------------|--------------- network B (10.99.0.0/24)
    External Peer
```

**System Information (please complete the following information):**

 - Kube-Router Version: v1.5.0
 - Kube-Router Parameters: `--peer-router-ips=<EXTERNAL_PEER_IP>`
 - Kubernetes Version: v1.22.10
 - Cloud Type: on premise
 - Kubernetes Deployment Type: custom
 - Kube-Router Deployment Type: DaemonSet
 - Cluster Size: 2

**Logs, other output, metrics**

```
time="2022-10-10T09:45:09Z" level=info msg="Mismatched local address" Addr=10.99.0.6 BindInterface= Configured addr=10.21.136.204 Key=10.99.0.2 Topic=Peer
```

```
Peer         AS Up/Down State       |#Received  Accepted
10.99.0.2 64513   never Active      |        0         0
10.99.0.3 64513   never Active      |        0         0
```

**Additional context**

This PR seem to be the reason https://github.com/cloudnativelabs/kube-router/pull/777

aauren:
@pmint93 - Thanks for doing some leg work on tracking down the issue. I agree with you, that #777 is likely is stopping you from peering on all interfaces.

[...]

PR #1392, opened 2022-11-02
Support for kube-router.io/peer.localips annotation

### LINKED ISSUE #1391: v1.5.0 external bgp connection fails
**What happened?**
Upgraded from v1.4 to v1.5, and external bgp peer connections failed to estabilish.

Expected external bgp sessions to come up as before.

1. Have a node with two (or more L3) interfaces, one said to be k8s backbone, and one facing an external router
2. Annotate the node to have the external bgp configuration
3. Restart kube-router on node
4. Observe `Mismatched local address` log messages in kube-router's log

**Additional context**
I suspect 184976a5369920ec176da961e61de739726be5fb causes the change. Perhaps,  it would be helpful to specify a localaddress for peers via annotations the same way as other parameters.

rkojedzinszky:
Fixes #1391

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -47,6 +47,7 @@ const (
 	pathPrependRepeatNAnnotation     = "kube-router.io/path-prepend.repeat-n"
 	peerASNAnnotation                = "kube-router.io/peer.asns"
 	peerIPAnnotation                 = "kube-router.io/peer.ips"
+	peerLocalIPAnnotation            = "kube-router.io/peer.localips"
 	//nolint:gosec // this is not a hardcoded password
 	peerPasswordAnnotation             = "kube-router.io/peer.passwords"
 	peerPortAnnotation                 = "kube-router.io/peer.ports"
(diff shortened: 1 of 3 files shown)


aauren:
Relates to #1371

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -1130,9 +1131,38 @@ func (nrc *NetworkRoutingController) startBgpServer(grpcServer bool) error {
...
 
+		// Get Global Peer Router LocalIP configs
+		var peerLocalIPs []string
+		nodeBGPPeerLocalIPs, ok := node.ObjectMeta.Annotations[peerLocalIPAnnotation]
+		if !ok {
+			klog.Infof("Could not find BGP peer local ip info in the node's annotations. Assuming node IP.")
+		} else {
+			peerLocalIPs = stringToSlice(nodeBGPPeerLocalIPs, ",")
+			err = func() error {
+				for _, s := range ipStrings {
```
aauren, line 1142: Did you mean to pass this as a parameter to the anonymous function?

aauren (changes requested):

pkg/controllers/routing/network_routes_controller.go
rkojedzinszky: The anonymous function is there just to check if the list element is empty string or can be parsed as an IP address. The wrong list was being processed there, fixed now.

aauren:
This was released as part of `v1.5.3`: https://github.com/cloudnativelabs/kube-router/releases/tag/v1.5.3

PR #1402, opened 2022-11-29
Dont set LocalAddress for external bgp peers by default

### LINKED ISSUE #1401: Set bgp local-address implicitly for internal peers only
**Is your feature request related to a problem? Please describe.**
A feature request.

#778 was about that internal peers could not establish neighborship, and #777 fixed it by setting local address to nodeIP for _all_ peers, including external ones. This caused issues with existing setups (#1371). Additional fix has been implemented (#1392), howewer that needs the configuration be even more complex.

I suggest relaxing #777 by setting localaddress for internal peers only. That would bring the old behavior back, while still providing the fix for #778. It would make configuration much simpler in many cases. In other cases, one can still set localaddress explicitly with #1392.

rkojedzinszky:
Closes #1401


aauren:
Leaving comments on #1401

PR #1419, opened 2023-01-10
Extend bgp sent metric

KSauter:
The bgp sent metric was only counted for pod routed.
VIP routes, advertised via bgp, weren't.

This will also count for VIP routes advertisements and withdraws.

To distinguish between the three types, a new label was added.


mrueg (approved):

aauren (approved):

aauren:
@KSauter - Thanks for clearing up a couple of problems there. Sorry that it took so long to get around to reviewing it.

PR #1438, opened 2023-02-06
use JoinHostPort for GRPC listen address

whooo:
Fixes the "too many colons in address" issue when the primary IP is an IPv6 address.

Tested on an ipv6-only cluster


aauren (approved):
Thanks for finding this!

PR #1439, opened 2023-02-07
add generation of router id based on hash of primary IP

whooo:
When enabled, generate the router id by hashing the primary IP. With this no explicit router id has to be provided on IPv6-only clusters.

Tested on a (small) IPv6-only cluster.

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -2,8 +2,10 @@ package routing
 
 import (
 	"context"
+	"encoding/binary"
 	"errors"
 	"fmt"
+	"hash/fnv"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


pkg/controllers/routing/network_routes_controller.go
```diff
@@ -1366,9 +1368,17 @@ func NewNetworkRoutingController(clientset kubernetes.Interface,
 	}
 	nrc.isIPv6Capable = len(nrc.nodeIPv6Addrs) > 0
 
-	if kubeRouterConfig.RouterID != "" {
+	switch {
+	case kubeRouterConfig.RouterID == "generate":
```
aauren, line 1372: Can you add some information to the documentation about this option for setting the router ID?

Probably needs to go here: https://github.com/cloudnativelabs/kube-router/blob/master/pkg/options/options.go#L201 (this will necessitate updating the options here: https://github.com/cloudnativelabs/kube-router/blob/master/docs/user-guide.md)

aauren (changes requested):
Thanks for your continued updates of the IPv6 stuff @whooo!

The code changes look fine, but we should update the docs for that option too.

pkg/controllers/routing/network_routes_controller.go
whooo: Done

PR #1452, opened 2023-03-12
Fix Service VIP Prefix Set Build

### LINKED ISSUE #1442: [v2.0.0 Issue]: Changing a service to PreferDualStack does not advertise IPv6 routes
When modifying an existing service of type LoadBalancer to PreferDualStack, the IPv6 route is not advertised.

1. Run an nginx container ```kubectl run nginx --image=nginx```
2. Create service
```apiVersion: v1
kind: Service
metadata:
  name: nginx
spec:
  selector:
    run: nginx
  ports:
    - protocol: TCP
      port: 80
      name: http
  type: LoadBalancer
  externalTrafficPolicy: Local
```
3. Edit service and add `ipFamilyPolicy: PreferDualStack`
4. Confirm no IPv6 route is advertised
5. Delete service and create with `ipFamilyPolicy: PreferDualStack`
6. Confirm IPv6 route is advertised

**System Information:**
 - Kube-Router Version (`kube-router --version`): 2.0.0-rc1
 - Kube-Router Parameters: Deployed with Cilium```        - "--enable-ipv6=true"
        - "--run-router=true"
        - "--run-firewall=false"
        - "--run-service-proxy=false"
        - "--bgp-graceful-restart=true"
        - "--enable-cni=false"
        - "--enable-pod-egress=false"
        - "--enable-ibgp=true"
        - "--enable-overlay=true"
        - "--peer-router-ips="
        - "--peer-router-asns="
        - "--cluster-asn="
        - "--advertise-cluster-ip=false"
        - "--advertise-external-ip=true"
        - "--advertise-loadbalancer-ip=true"```
 - Kubernetes Version (`kubectl version`) : 1.26.1
 - Cloud Type: on premise
 - Cluster Size: 9 nodes

aauren:
@mrueg @rwagoner

When a single IP family's set looks to be equal, switch to continue instead of return so that other families can still be evaluated as those might have changes.

This should fix #1442

### DIFF
--- a/pkg/controllers/routing/bgp_policies.go
+++ b/pkg/controllers/routing/bgp_policies.go
@@ -213,7 +213,7 @@ func (nrc *NetworkRoutingController) addServiceVIPsDefinedSet() error {
 			return currentPrefixes[i].IpPrefix < currentPrefixes[j].IpPrefix
 		})
 		if reflect.DeepEqual(advIPPrefixList, currentPrefixes) {
-			return nil
+			continue
 		}
 		toAdd := make([]*gobgpapi.Prefix, 0)
 		toDelete := make([]*gobgpapi.Prefix, 0)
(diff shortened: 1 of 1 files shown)


aauren:
This PR now includes a lot of fixes. There were multiple places in the code that assumed certain things like BGP policy statements and VIPs would not be changed and so the code was additive only.

This addresses the ones that I found that broke my tests and the ones that were causing the issue on #1442.

PR #1459, opened 2023-03-30
Fix #1458 - check if endpoints subset changed

### LINKED ISSUE #1458: Log spam (and higher cpu usage) due to postgres-operator
**What happened?**
zalando/postgres-operator users the annotations of the main db service service as a keep alive token. This causes and endpoints update for kube-router. 
It will spam the output with lines like:

```
I0330 09:11:58.326271       1 network_services_controller.go:1106] Skipping service name:live-bitbucket-database namespace:atlassian as there is no cluster IP
```

etc. every second

Do not consider updates to annotations relevant, since they are not used in building the endpoint information anyways. This means they can be safely ignored

1. Create a service
2. Update the services annotations
3. Kube-router will recheck all the services on all nodes

 - Kube-Router Version (`kube-router --version`): [e.g. 1.0.1] 1.5.3
 - ```
        - "--run-router=false"
        - "--run-firewall=false"
        - "--enable-overlay=false"
        - "--run-service-proxy=true"
        - "--ipvs-sync-period=1m"
        - "--masquerade-all=true"
        - "--enable-pod-egress=false"
        - "--enable-cni=false"
        - "--metrics-port=8081"
        - "--metrics-path=/metrics"
        - "-v=2"
  ```
 - Kubernetes Version (`kubectl version`) : [e.g. 1.18.3] 1.24.12
 - Cloud Type: [e.g. AWS, GCP, Azure, on premise] on premise
 - Kubernetes Deployment Type: [e.g. EKS, GKE, Kops, Kubeadm, etc.] manual
 - Kube-Router Deployment Type: [e.g. DaemonSet, System Service] DaemonSet
 - Cluster Size: [e.g. 200 Nodes] 10

Thoro:
Adds a check, so that Endpoints are only updated if the subset is changed - ignoring changes to annotations.


Thoro:
Leads to massive CPU load reduction if zalando/postgres-operator or similar is in use:
Screenshot 2023-03-30 at 12 09 24

aauren:
@Thoro - Thanks for looking into this. I think that this is a good change, and one that is fairly straight forward. We already do similar things on other objects, like assessing pod changes to see if they are network policy relevant, before processing them.

At this point, we are not adding any more feature or enhancements to the v1.X line of kube-router. Can you please rebase your commit on the prep-v2.0 branch and make sure that you don't have any commits on it other than your own? Once that is done, and it passes CI, we should be able to merge this.

aauren:
(you'll also need to see if GitHub will allow you to change this PR to merge into the `prep-v2.0` branch as well. If it won't, then you'll need to close this PR and open a new one)

Thoro:
Redid the PR: #1461

PR #1470, opened 2023-04-24
fix(dsr): CRI runtime/v1alpha2 -> runtime/v1

### LINKED ISSUE #1464: DSR api v1alpha2.RuntimeService outdated on 1.26 with crio
**What happened?**
I build a baremetal cluster, use metallb to provide loadbalancer, and deploy a redis-cluster with external access, redis pod needs to access others via loadbalancer, however, when cluster initialize, redis pod access it self via loadbalancer.

so iptables trace like
`PACKET: 2 693ce7cf OUT=kube-bridge LL=0x0 ffffff8a2affffff9bffffffbf7effffffb8ffffffd648ffffff886bffffff8cffffff8c0800 SRC=10.244.2.233 DST=10.244.2.233 LEN=60 TOS=0x0 TTL=63 ID=59297DF SPORT=42404 DPORT=80 SYN`
and tcpdump like
`10:40:29.714577 IP (tos 0x0, ttl 64, id 58914, offset 0, flags [DF], proto TCP (6), length 60)
    10.244.2.233.50746 > xxxxxxxxxxx.80: Flags [S], cksum 0x30de (incorrect -> 0x1b7a), seq 284959392, win 64240, options [mss 1460,sackOK,TS val 3218622182 ecr 0,nop,wscale 7], length 0
10:40:29.714654 IP (tos 0x0, ttl 63, id 58914, offset 0, flags [DF], proto TCP (6), length 60)
    10.244.2.233.50746 > 10.244.2.233.80: Flags [S], cksum 0x1be8 (incorrect -> 0x3070), seq 284959392, win 64240, options [mss 1460,sackOK,TS val 3218622182 ecr 0,nop,wscale 7], length 0`

packet lost, can't access, so I want to test DSR, but DSR can't work.

see this https://github.com/cri-o/cri-o/issues/6554
use v1.RuntimeService api

1. configure kube-router with DSR
2. annotate service
3. service can't access

 - Kube-Router Version (`kube-router --version`): [1.5.3]
 - Kube-Router Parameters: [
        - --run-router=true
        - --run-firewall=true
        - --run-service-proxy=true
        - --bgp-graceful-restart=true
        - --service-external-ip-range=xxxxxxxx/xx
        - --runtime-endpoint=unix:///var/run/crio/crio.sock]
 - Kubernetes Version (`kubectl version`) : [e.g. 1.26.3]
 - Kubernetes Deployment Type: [Kubeadm]
 - Kube-Router Deployment Type: [DaemonSet]
 - Cluster Size: [3 Nodes]

Error setting up IPVS services for service external IP's and load balancer IP's: failed to setup DSR endpoint xxxxxxxx: unable to setup DSR receiver inside pod: failed to prepare endpoint 10.244.2.233 to do DSR due to: rpc error: code = Unimplemented desc = unknown service runtime.v1alpha2.RuntimeService

**Additional context**
There's any ways to deal with pod access it self via loadbalancer?

aauren:
@mrueg 

Updates the cri-api to use `runtime/v1` instead of `runtime/v1alpha2` which has been deprecated for a bit now.

History of this and how it affects the project:
* cri-o
  * Added support for v1 and v1alpha2 starting in `v1.21.0` which was released `~2 years ago`
  * Removed support for v1alpha2 starting in `v1.26.0` which was released `~4 months ago`
  * Last release that only contained v1alpha2 (and not v1 additionally) was `v1.20.9` which was released `~9 months ago` and is no longer supported
* containerd
  * Added support for v1 and v1alpha2 starting in `v1.6.0` which was released `~1 year 2 months ago`
  * Removed support for v1alpha2 starting in `v1.7.0` which was released `~1 month ago`
  * Last release that only contained v1alplha2 (and not v1 additionally) was `v1.5.18` which was released `~2 months ago` and is no longer supported
* docker
  * Is not affected as we don't use the CRI API to talk to Docker

While this kinda breaks the traditional major.minor.bugfix pattern, I'm considering cutting this as part of a v1.5.4 release of kube-router. I generally feel like v1.5 is only in support mode right now and really don't want to release and support a v1.6 release ahead of going to v2.0 once I finish out the NSC work I've been working on.

Its possible that given that the two deprecation dates, that there may still be some people out there running containerd from `v1.5.x` line given that it was only deprecated 2 months ago (https://containerd.io/releases/#support-horizon), and this would break DSR functionality for them. It might be surprising to have that happen when upgrading to a bugfix release.

All that said, this probably does need to be released sooner rather than later as we're currently breaking supported versions of cri-o and will begin breaking supported versions of containerd when 1.7.x gains adoption from major distros.

Let me know what you think @mrueg ?

Replaces #1468 

Fixes #1464


mrueg:
I think it makes sense to have it in particular if it applies as clean as this one. :+1:

aauren:
`v1.5.4` is building here: https://github.com/cloudnativelabs/kube-router/actions/runs/4791628826

PR #1471, opened 2023-04-24
feat(ci.yml): update github action versions

aauren:
@mrueg 

Hopefully this removes all of the warnings that we get about NodeJS 12 not being supported any longer, but at the very least it should bring us current with new versions of actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Set up Go 1.x
-      uses: actions/setup-go@v3
+      uses: actions/setup-go@v4
       with:
         go-version: ${{ env.GO_VERSION }}
       id: go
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -187,7 +187,7 @@ jobs:
       uses: actions/checkout@v3
 
     - name: Run GoReleaser
-      uses: goreleaser/goreleaser-action@v3
+      uses: goreleaser/goreleaser-action@v4
       with:
         version: latest
         args: release --rm-dist
```
mrueg, line 193:
```suggestion
        args: release --clean
```
Can you update this as well? It got deprecated.

mrueg:
I believe we need to take a look at 

${{ steps.extract_tag.outputs.tag }} and other steps keys as well, they got deprecated iirc

aauren:
> I believe we need to take a look at
>
> ${{ steps.extract_tag.outputs.tag }} and other steps keys as well, they got deprecated iirc

Hmm... do you remember how or what about it got deprecated? I see that `extract_tag` comes from an earlier step that is just a bash step. I don't think that anything that we're adding changed how that was extracted, as that step doesn't have a `uses` statement.

While we did upgrade the `build-push-action` from `@v3` to `@v4`, it looks like both versions accept the `tags` parameter: https://github.com/docker/build-push-action#inputs

PR #1474, opened 2023-05-08
netpol: Fix ipset only containing one IP when port name is used.

BoleynSu:


aauren:
Thanks for submitting @BoleynSu!

Looks good on the surface. I'll try to give this a test later this week.

aauren:
@BoleynSu Sorry that this took so long to come back to.

Got a chance to put this change through it's paces and went through the thread that you posted on the k3s side. Everything looks good. Thanks again for being willing to submit this upstream as well as in k3s!

PR #1490, opened 2023-06-24
Use alpine linux 3.18 for docker image

whooo:
The ipset version included in alpine 3.17 doesn't work on newer kernels.


aauren:
Merged here: https://github.com/cloudnativelabs/kube-router/commit/2838b9b54717283420513587bb227a505af031ea

PR #1496, opened 2023-07-02
fix(ci): checkout before go setup

aauren:
@mrueg 

In setup-go@v4 it changed to enabling caching by default which looks for a go modules file to check the cache. This means that the checkout needs to be processed before the setup-go action.


aauren:
Resolves CI warnings like this one: https://github.com/cloudnativelabs/kube-router/actions/runs/5437599174/jobs/9888266810#step:2:20

mrueg:
Thanks!

PR #1497, opened 2023-07-02
CI Add Caching

aauren:
@mrueg 

Add caching to the kube-router build, test, moq, & lint Makefile targets

With a primed cache I've found that this greatly increases the responsiveness of each of this make targets:

```
Standard: make kube-router  0.06s user 0.03s system 0% cpu 28.231 total
With Primed Cache: make kube-router  0.05s user 0.04s system 3% cpu 2.340 total

Standard: make test  0.18s user 0.07s system 0% cpu 36.072 total
With Primed Cache: make test  0.20s user 0.04s system 21% cpu 1.171 total

Standard: make lint  0.18s user 0.08s system 0% cpu 40.760 total
With Primed Cache: make lint  0.20s user 0.04s system 1% cpu 14.314 total
```

I _think_ that we might be able to gain some performance out of the `container` target as well by adding caching to the Dockerfile similar to: https://www.docker.com/blog/containerize-your-go-developer-environment-part-2/

However, I've never had much luck with that, and don't exactly know how to navigate the whole `Makefile` -> `docker build` -> `Makefile` chain that we have for that step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,8 @@ env:
   BUILDTIME_BASE: "golang:1.19.5-alpine3.17"
   RUNTIME_BASE: "alpine:3.17"
   GO_VERSION: "~1.19.5"
+  GO_CACHE: "/home/runner/.cache/go-build"
+  GO_MOD_CACHE: "/home/runner/go/pkg/mod"
 
 jobs:
   # Runs Golangci-lint on the source code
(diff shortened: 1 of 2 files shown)


aauren:
image

^^^ Above you can see the differences made in GitHub. All three jobs were run without a primed cache, and then ci-build-kube-router was run again once the cache was primed resulting in a decrease of runtime from ~2.5 minutes to ~30 seconds.

aauren:
From the documentation, it looks like GitHub only keeps build caches around for 7 days so not sure how big of an impact it will make all the time, but during quick iteration cycles it should be pretty beneficial.

mrueg (approved):

PR #1525, opened 2023-08-21
Add RISC-V 64 support

iggy:
Add goreleaser and Makefile entries to build/deploy riscv64 assets

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -10,6 +10,7 @@ builds:
   - arm64
   - ppc64le
   - s390x
+  - riscv64
   goarm:
   - 6
   - 7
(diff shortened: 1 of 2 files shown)


iggy:
I hopefully got the `FILE_ARCH` bit right. Fwiw, here is the fill output from `file`

```
/bin/busybox: ELF 64-bit LSB pie executable, UCB RISC-V, RVC, double-float ABI, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-riscv64.so.1, BuildID[sha1]=a87aae564a75c28b98fdba5cf1ebbc22390b97d8, stripped
```

mrueg (approved):

PR #1534, opened 2023-08-29
Partially revert riscv64 support

mrueg:
@iggy @aauren Since the alpine image we use does not have a riscv64 architecture, we can't build riscv64 images. We can still build a riscv64 binary with goreleaser though.

We'd need to wait for https://gitlab.alpinelinux.org/alpine/aports/-/issues/13269

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,7 +119,6 @@ jobs:
           linux/arm/v7
           linux/s390x
           linux/ppc64le
-          linux/riscv64
         push: true
         build-args: |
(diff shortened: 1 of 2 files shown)


iggy:
Makes total sense. I missed the fact that only the `edge` container image had riscv64 support. I've talked to the alpine devs about this already and basically they are waiting to get actual hardware that can reliably build the packages before they add riscv64 to a stable release. I'm working with the riscv64 team over there, so I'll keep tabs on the progress and revisit this when it gets into a stable release. Thanks for the quick turn around on the binary part (it's what I actually needed anyway).

Issue #1432, opened 2023-01-25
[v2.0.0 Issue]: Enable Dual-Stack Configuration with Proxy Controller


aauren:
We still need to retrofit the Proxy controller (enabled with the `--run-service-proxy` command-line option) to work with dual-stack addresses. This is implementation is needed so that kube-router's dual-stack implementation can be considered feature complete and a full v2.0.0 release can be released.

aauren:
Fixed by #1544

PR #1544, opened 2023-09-14
Add Dual-Stack Capabilities to NSC

aauren:
This is an initial stab at adding dual-stack capabilities to NSC which is the last controller needed to merge into master and do a release of v2.0.0.

I still need to do some additional testing, but I figured I would put this up in case anyone else in the community wanted to test it. It has so far been stable for me, but there are a LOT of changes contained in this MR and it was rough work getting all of the logic in line.

One of the things to specifically call out is that we now use `EndpointSlices` instead of `Endpoints` in order to get ALL the possible IP endpoints for a given service as `Endpoints` only returns what it considers the best IP for a given pod endpoint rather than all of the IPs it might have on it (especially those from another family).


aauren:
Fixes #1432

PR #1545, opened 2023-09-16
Add k3s Improvements

aauren:
@brandond @ChristianCiach

This work was taken from @brandond's fork of kube-router for k3s: https://github.com/brandond/kube-router/tree/metrics-var

The primary work done there was to:

* Improve the performance of the network policy controller (NPC) by reducing the number of shell calls made to ipset when syncing policies to help improve the performance problems reported by @ChristianCiach here: https://github.com/k3s-io/k3s/discussions/8321
* Improve observability in the NPC by adding metrics for number of iptables chain and number of ipsets that kube-router manages
* Improve observability in the NPC by adding metrics for save / restore of iptables / ipsets per family
* Added ability to swap out metrics handlers for projects that consume kube-router as a library

All the work here was thanks to @brandond and @ChristianCiach's troubleshooting and fixing

@brandond - I added a few changes on top of the ones that you had. I separated out the commits into a pretty granular set so that they would be easier for you to review. Let me know what you think of the changes that I added on top of yours. Let me also know if you'd prefer that I close this PR and allow you to submit it yourself. I'm fine either way.

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -253,7 +253,6 @@ func (npc *NetworkPolicyController) fullPolicySync() {
 	}
 
 	for ipFamily, iptablesSaveRestore := range npc.iptablesSaveRestore {
-		ipFamily := ipFamily
 		npc.filterTableRules[ipFamily].Reset()
 		saveStart := time.Now()
 		err := iptablesSaveRestore.SaveInto("filter", npc.filterTableRules[ipFamily])
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -265,7 +265,7 @@ func (npc *NetworkPolicyController) fullPolicySync() {
 				metrics.ControllerIptablesV6SaveTime.Observe(saveEndTime.Seconds())
 			}
 		}
-		klog.V(2).Infof("Saving %v iptables rules took %v", ipFamily, saveEndTime)
+		klog.V(1).Infof("Saving %v iptables rules took %v", ipFamily, saveEndTime)
 
 		if err != nil {
 			klog.Errorf("Aborting sync. Failed to run iptables-save: %v", err.Error())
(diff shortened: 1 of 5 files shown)


aauren:
@brandond - I also saw that k3s had modified the `ensureExplicitAccept` function: https://github.com/cloudnativelabs/kube-router/pull/1545/files#diff-61248fd525f15b9b270112c7242e93d27a1824b1c0f618fda2d6fd680c66bf6fR551-R562 but the changes didn't really make sense to me, so I omitted them from this PR during the rebase.

Here they are from your commit: https://github.com/brandond/kube-router/commit/079f06df7f5d81c621986fb3ecad6018fee708e7#diff-61248fd525f15b9b270112c7242e93d27a1824b1c0f618fda2d6fd680c66bf6fR498-R526

aauren:
> I had put the metrics in deferred anonymous functions because that seems to be the pattern for the metrics across the rest of the codebase. Completion messages, unlocks, and metrics, all seem to be handled with a defer. Is there any reason to not do that with these, other than the slight inconvenience of needing create some extra vars to capture loop variables?

The main reason why we defer other places is to be able to capture the entire runtime of the whole function which is usually some sort of sync loop. In this case, since we are only capturing a single operation (save or restore), we can emit it as soon as it occurs. So I don't think that it is necessary to keep the deferral logic for these metrics.

If that was something that was important though, a simple fix would be to add the endtime as a parameter to the function call similar to: https://goplay.tools/snippet/OvIF49w1J4-

aauren:
@brandond any further thoughts on this one before I merge it?

brandond:
No, looks great, thanks!

> I also saw that k3s had modified the ensureExplicitAccept function

This was due to preserve some legacy behavior that made flannel work when users have a deny policy. Apparently we were relying on behavior that accepted the packet, instead of returning it to hit the deny rule. See https://github.com/k3s-io/k3s/issues/7203#issuecomment-1495345235

I don't think we expected y'all to upstream this, although I guess we could rework the logic make it optional, and enable it by default on K3s? cc @rbrtbnfgl.

aauren:
> No, looks great, thanks!

Thank you! I really appreciate k3s doing the leg work on debugging this and fixing it! I also really like the extra metrics, those will be handy for users going forward to better track regressions if they happen in the future.

PR #1570, opened 2023-11-03
prometheus metrics: add option to specify listen address

jpiper:
In the situation that you have multiple interfaces/IP addresses, we want to be able to specify which one we want to expose the prometheus metrics on.

This is the follow up to #1244 with the PR feedback addressed.


aauren:
Thanks @jpiper!

I'll try to take a look this weekend and merge.

aauren:
Got a chance to put it through its paces. LGTM.

Thanks for the contribution!

PR #1582, opened 2023-12-03
fix(hairpin): set hairpin_mode for veth iface

aauren:
It used to be that the kubelet handled setting hairpin mode for us: https://github.com/kubernetes/kubernetes/pull/13628

Then this functionality moved to the dockershim:
https://github.com/kubernetes/kubernetes/pull/62212

Then the functionality was removed entirely:
https://github.com/kubernetes/kubernetes/commit/83265c9171f

Unfortunately, it was lost that we ever depended on this in order for our hairpin implementation to work, if we ever knew it at all. Additionally, I suspect that containerd and cri-o implementations never worked correctly with hairpinning.

Without this, the NAT rules that we implement for hairpinning don't work correctly. Because hairpin_mode isn't implemented on the virtual interface of the container on the host, the packet bubbles up to the kube-bridge. At some point in the traffic flow, the route back to the pod gets resolved to the mac address inside the container, at that point, the packet's source mac and destination mac don't match the kube-bridge interface and the packet is black-holed.

This can also be fixed by putting the kube-bridge interface into promiscuous mode so that it accepts all mac addresses, but I think that going back to the original functionality of enabling hairpin_mode on the veth interface of the container is likely the lesser of two evils here as putting the kube-bridge interface into promiscuous mode will likely have unintentional consequences.

With this fix, kube-router now passes the Kubernetes end-to-end conformance test for hairpinning traffic.

### DIFF
--- a/pkg/controllers/proxy/hairpin_controller.go
+++ b/pkg/controllers/proxy/hairpin_controller.go
@@ -1,0 +1,121 @@
+package proxy
+
+import (
+	"fmt"
+	"net"
+	"os"
+	"path"
+	"runtime"
+	"sync"
+	"time"
+
+	"github.com/cloudnativelabs/kube-router/v2/pkg/healthcheck"
+	"github.com/cloudnativelabs/kube-router/v2/pkg/utils"
+	"github.com/vishvananda/netns"
+	"k8s.io/klog/v2"
+)
+
+type hairpinController struct {
+	epC <-chan string
+	nsc *NetworkServicesController
+}
+
+func (hpc *hairpinController) Run(stopCh <-chan struct{}, wg *sync.WaitGroup,
+	healthChan chan<- *healthcheck.ControllerHeartbeat) {
+	defer wg.Done()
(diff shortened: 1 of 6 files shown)


pkg/controllers/proxy/hairpin_controller.go
```diff
@@ -0,0 +1,121 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to prepare endpoint %s to do direct server return due to %v",
+				endpointIP, err)
+		}
+	} else {
+		// We expect CRI compliant runtimes here
+		// ugly workaround, refactoring of pkg/Proxy is required
+		pid, err = hpc.nsc.ln.getContainerPidWithCRI(hpc.nsc.dsr.runtimeEndpoint, containerID)
+		if err != nil {
+			return fmt.Errorf("failed to prepare endpoint %s to do DSR due to: %v", endpointIP, err)
```
szuecs, line 86: you should use `fmt.Errorf("...: %w", ..., err)` instead of %v for error type propagation that you can later check with errors.Is or convert with errors.As

aauren:
Progress towards: #1596

PR #1584, opened 2023-12-08
Fix Ddd hostPID to Daemonsets

aauren:
Since #1582  was merged, we now rely upon having access to the host's PID namespace in order to detect which veth device a given container doing hairpinning is using. As such, we now add it to the daemonsets in the example and attempt to direct users upon certain errors.


aauren:
Progress towards: #1596

Issue #1588, opened 2023-12-13
kube-router breaks iptables beacause it insists on using iptables table "filter"


simonfelding:
**What happened?**
When I start kube-router, it creates the table "filter" and adds rules to it. This leads to the error:
`iptables v1.8.5 (nf_tables): table 'filter' is incompatible, use 'nft' tool.` when I use `iptables -L`.

It's entirely avoidable by just not using the "filter" table when the nf_tables backend is used, as nftables doesn't put special meaning to table names.

**What did you expect to happen?**
I expected kube-router to work without breaking iptables.

**How can we reproduce the behavior you experienced?**
1. install oracle linux 8
2. download k0s
3. `k0s install controller --single`
4. `watch iptables -L`

** System Information (please complete the following information):**
 - Kube-Router Version (`kube-router --version`): 1.6.0
 - Kube-Router Parameters: default
 - Kubernetes Version (`kubectl version`) : v1.28.4
 - Cloud Type: on premise
 - Kubernetes Deployment Type: k0s
 - Kube-Router Deployment Type: [e.g. DaemonSet, System Service]
 - Cluster Size: 1 node

simonfelding:
I realize this is kube-router v.2.0.0, but I looked through the kube-router code and didn't find any changes that would have solved this.

aauren:
Hey @simonfelding 

First off, I'm not sure if you intended this or not, but your issue description comes off a bit abrasive. The sentiment seems to be that we insisted on breaking compatibility for your use-case. While that may happen from time to time accidentally, we certainly never intend to make design mistakes in kube-router and we would appreciate if, in the future you would word your issue descriptions a bit more carefully so as to give us, and other open source projects, the benefit of the doubt in this respect.

In this case, we are not able to use any table other than the filter table because at its heart kube-router uses iptables the same as you're trying to do on your node. The incompatibility that you're experiencing is because the iptables that is included in the kube-router container by k0s is using iptables in legacy mode (https://wiki.nftables.org/wiki-nftables/index.php/Legacy_xtables_tools). You can see this by entering the kube-router container of k0s and looking at how it is linked:

```sh
k0s:~# iptables --version
iptables v1.8.9 (nf_tables)
k0s:~# which iptables
/sbin/iptables
k0s:~# ls -l /sbin/iptables
lrwxrwxrwx    1 root     root            23 Dec 13 20:42 /sbin/iptables -> /sbin/xtables-nft-multi
```

[...]

PR #1589, opened 2023-12-14
Pin iptables to 1.8.8-r1

simonfelding:
As per https://github.com/cloudnativelabs/kube-router/issues/1588, pin the iptables version to 1.8.8-r1 in order to support RHEL8-based distros.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,10 +10,11 @@ RUN apk add --no-cache make git \
     && make gobgp
 
 FROM ${RUNTIME_BASE}
-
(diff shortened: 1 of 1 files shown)


aauren:
@simonfelding - At this point, I don't think that we're willing to accept a pin of iptables in the main kube-router container, but if you want to rework this as an additional container based on the RHEL UBI image, like @mrueg suggested here: https://github.com/cloudnativelabs/kube-router/issues/1588#issuecomment-1863333452 then we are open to that change.

aauren:
Closing this for lack of response.

Issue #521, opened 2018-08-25
TCP Connection reset after connection was idle for more that 15 minutes


tanapoln:
My applications communicates via gRPC and the connections was not active all time. When a specific connection was idle for more than 15 minutes, it would get RST packet after try to write to that connection.

A tcpdump on client-side (a sender) shows that RST packet comes from remote IP but tcpdump on server-side (a receiver) show nothing about RST packet. I also notice that TTL of RST packet is not decrease (default TTL=64 in my server).

I suspect that it was not sent from remote but in a local itself. You can check with this tcpdump capture: [tcpdump.zip](https://github.com/cloudnativelabs/kube-router/files/2320742/tcpdump.zip)

I also wonder why there are retransmission and duplicate packets in tcpdump result but leave it later about that.

The problem I found can reproduce reliably, you can use attached Kubernetes YAML to test and see result: [grpc-test.yml.zip](https://github.com/cloudnativelabs/kube-router/files/2320741/grpc-test.yml.zip). I also tested using default `kubenet` and problem was not found.

The code of grpc-test do simply task: client send RPC to server every specific interval which can be configure via environment variable `INTERVAL_MS`. The connection will establish at the beginning when client was started.

### Environments:
Kubernetes: 1.9.3
Docker: 17.3.2
Linux Kernel: 4.4.121-k8s (Debian jessie)
Kube-router: 0.1.0

My cluster was provisioned using Kops with this command:
```
export NAME=testcluster.k8s.local
export KOPS_STATE_STORE=s3://my-testcluster-kops-xxx
kops create cluster --dns private --api-loadbalancer-type public --kubernetes-version=1.9.3 --zones ap-southeast-1b --node-count 1 --node-size t2.medium --master-size t2.medium --networking kube-router -t public ${NAME}
```

tanapoln:
Investigated and found a problem.

The default value of IPVS Connection timeout is 900 seconds, after increase a timeout that (via `ipvsadm --set`) a problem is gone.

I think this config value must be implemented inside kube-route due to a manual config is not persist across reboot.

znoodl:
@murali-reddy 
I found another way to fix this, https://success.docker.com/article/ipvs-connection-timeout-issue
> tcp_keepalive_time must less than ipvs(900 secs), and client socket must set SO_KEEPALIVE.

But it's not works for me, do you have some plans to support config in kube-router?
or I have to set it by myself.

roffe:
[...]

PR #1590, opened 2023-12-21
feat(NSC): add IPVS service timeouts

### LINKED ISSUE #911: TCP connection reset after some hours of idle
Might be related to #521, but setting IPVS timeout (sessionAffinityConfig.clientIP.timeoutSeconds) to 86400 does not help.

TCP connection from POD to service (in our case from application to PostgreSQL service, the same with Elasticsearch service) is reset after some hours of Idle.
Which timeout may affect us in this case?
We do not experience such issues on another cluster (there is Openshift with VXLAN/iptables based networking).

### LINKED ISSUE #1577: Getting RST on long-lived connections
**What happened?**
I don't know exactly but from time to time my apps receive RST from the kernel they are running on. It looks very much like this https://unix.stackexchange.com/questions/572276/l4-balancing-using-ipvs-drop-rst-packets-failover but there is no IP switching or anything

App does syn/ack, then works.... time passes here..... app sends `[P.]` and gets RST from kernel immediately. I have a feeling that some syncing/re-syncing mechanism in kube-router flashes state table (does ipvs have state table??)

Nothing like that, just a working socket

Screenshot from 2023-11-28 17-39-33

 - Kube-Router Version (`kube-router --version`): 1.6.0

aauren:
This is a feature that has been requested a few times over the years and would bring us closer to feature parity with other k8s network implementations for service proxy.

Adds the options:

```
      --service-tcp-timeout duration                  Specify TCP timeout for IPVS services in standard duration syntax (e.g. '5s', '1m'), default 0s preserves default system value (default: 0s)
      --service-tcpfin-timeout duration               Specify TCP FIN timeout for IPVS services in standard duration syntax (e.g. '5s', '1m'), default 0s preserves default system value (default: 0s)
      --service-udp-timeout duration                  Specify UDP timeout for IPVS services in standard duration syntax (e.g. '5s', '1m'), default 0s preserves default system value (default: 0s)
```

Keeps backwards compatibility by default defining them as `0s` which will keep the IPVS defaults in place (which is how these are currently set). The IPVS default appear to be the same on most systems, which should be:

```
TCP Timeout: 15 minutes
TCPFIN Timeout: 2 minutes
UDP Timeout: 5 minutes
```

Fixes #911 #521 - Without relying on the application to send keep-alives

Potentially resolves #1577 - But there isn't enough information in the ticket in order to know for sure.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -127,6 +127,9 @@ Usage of kube-router:
       --service-cluster-ip-range strings              CIDR values from which service cluster IPs are assigned (can be specified up to 2 times) (default [10.96.0.0/12])
       --service-external-ip-range strings             Specify external IP CIDRs that are used for inter-cluster communication (can be specified multiple times)
       --service-node-port-range string                NodePort range specified with either a hyphen or colon (default "30000-32767")
+      --service-tcp-timeout duration                  Specify TCP timeout for IPVS services in standard duration syntax (e.g. '5s', '1m'), default 0s preserves default system value (default: 0s)
+      --service-tcpfin-timeout duration               Specify TCP FIN timeout for IPVS services in standard duration syntax (e.g. '5s', '1m'), default 0s preserves default system value (default: 0s)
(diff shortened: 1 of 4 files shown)


aauren:
Progress towards: #1596

PR #1591, opened 2023-12-26
Respect service.kubernetes.io Labels

### LINKED ISSUE #979: Honor the service.kubernetes.io/service-proxy-name label
**Is your feature request related to a problem? Please describe.**

* [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0031-20181017-kube-proxy-services-optional.md)

The `service.kubernetes.io/service-proxy-name` was initially intended to off-load the kube-proxy from handling all services already handled by a service mesh. I think there will be more specialized K8s proxiers coming in the near future. People has requirement for special things and to get a (possibly odd) feature into the "real" K8s proxier (or Kube-router) is hard (if at all possible).

Kube-router should honour the `service.kubernetes.io/service-proxy-name` label when taking over the standard kube-proxy. This means; if the label is defined for a service Kube-router **must not** setup loadbalancing for that service. Example;

```
apiVersion: v1
kind: Service
metadata:
  name: mserver-no-proxy
  labels:
    service.kubernetes.io/service-proxy-name: istio
spec:
  selector:
    app: mserver-deployment
  ports:
  - port: 5001
    name: mconnect
```

**Describe alternatives you've considered**

You could also take a more defensive approach in Kube-router by (optionally) allowing a setting like `service.kubernetes.io/service-proxy-name: kube-router` to let kube-router handle _only_ some services, e.g. for DSR. Some administrations are reluctant to deviate from "standard" K8s, but special handling for special services may be ok.

**Additional context**

* [PR](https://github.com/kubernetes/kubernetes/pull/71355)
* [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0031-20181017-kube-proxy-services-optional.md)
* Related comment; https://github.com/kubernetes/kubernetes/issues/92369#issuecomment-650034563

This is easy to implement by filtering the service watch (https://github.com/sbezverk/nfproxy/pull/50).

aauren:
@mrueg 

This teaches kube-router about the following two kubernetes standard labels:

* `service.kubernetes.io/headless`
* `service.kubernetes.io/service-proxy-name`

It also resolves 2 of the e2e kubernetes testing failures that kube-router currently has and resolves #979

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -1612,24 +1612,24 @@ func (nsc *NetworkServicesController) cleanupMangleTableRule(ip string, protocol
 	args := []string{"-d", ip, "-m", protocol, "-p", protocol, "--dport", port, "-j", "MARK", "--set-mark", fwmark}
 	exists, err := iptablesCmdHandler.Exists("mangle", "PREROUTING", args...)
 	if err != nil {
-		return fmt.Errorf("Failed to cleanup iptables command to set up FWMARK due to " + err.Error())
+		return fmt.Errorf("failed to cleanup iptables command to set up FWMARK due to %v", err)
 	}
 	if exists {
 		klog.V(2).Infof("removing mangle rule with: iptables -D PREROUTING -t mangle %s", args)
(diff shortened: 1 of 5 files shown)


aauren:
Progress towards: #1596

Issue #1597, opened 2024-01-08
Fully Support Upstream Service Traffic Policies


aauren:
**Is your feature request related to a problem? Please describe.**

Upstream has defined a set of [traffic policies](https://kubernetes.io/docs/reference/networking/virtual-ips/#traffic-policies) that Kubernetes network providers should support.

kube-router currently supports `spec.externalTrafficPolicy`, but does not support `spec.internalTrafficPolicy` (which was introduced as stable in v1.26) at all.

Additionally, even when it applies `spec.externalTrafficPolicy` it does so by setting both Cluster IP and External IP to route locally or cluster, as opposed to only affecting the External traffic flows.

**Describe the solution you'd like**

As related to #1596 our goal is to make kube-router 100% conformant with upstream network implementation standards. To this end, kube-router should support `spec.internalTrafficPolicy` and fix it's implementation of `spec.externalTrafficPolicy`.

**Describe alternatives you've considered**

N/A

Issue #1596, opened 2024-01-06
kube-router Doesn't Currently Pass All Upstream e2e Integration Tests


aauren:
**What happened?**

kube-router doesn't currently pass all of the [e2e network integration tests](https://github.com/kubernetes/kubernetes/tree/master/test/e2e/network) provided by upstream. In order to be a fully compliant CNI we should be passing as many of these as is possible.

This issue is to track the progress of fixing these errors in kube-router.

**What did you expect to happen?**

kube-router in a basic kops cluster should be able to pass the e2e network tests.

**How can we reproduce the behavior you experienced?**

The upstream tests can be run using [kubetest2](https://github.com/kubernetes-sigs/kubetest2/blob/master/README.md) and the kops deployer / tester as follows:

```sh
time kubetest2 kops \
  -v 2 \
  --test \
  --cloud-provider=aws \
  --cluster-name=${NAME} \
  --kops-binary-path=${KOPS_ROOT}/.build/dist/$(go env GOOS)/$(go env GOARCH)/kops \
  --kubernetes-version=v1.27.8 \
  --test=kops \
  -- \
  --test-package-bucket=k8s-release-dev --test-package-dir=ci --test-package-marker=latest.txt \
  --parallel 25 \
  --skip-regex="\[Slow\]|\[Serial\]|\[Disruptive\]|\[Flaky\]|\[Feature:.+\]|\[HPA\]|Dashboard|RuntimeClass|RuntimeHandler|KubeProxy" \
  --focus-regex="\[sig-network\]"
```

The above will run the `sig-network` labeled tests against an existing cluster. `kubetest2` can also be configured to create the cluster from scratch as well.

https://testgrid.k8s.io/kops-network-plugins#kops-aws-cni-kuberouter - Is the automated testing harness that the kops project runs and shows a historical list of test results. However, it should be kept in mind that:

* This won't be up to date with any changes that have been made since the v1.6.0 version of kube-router
* In order to weed out recurring errors, tests that are known to fail [are not run as part of this suite for kube-router](https://github.com/kubernetes/kops/blob/master/tests/e2e/pkg/tester/skip_regex.go#L99-L102)

**Screenshots / Architecture Diagrams / Network Topologies **

Failing tests as of v2.0.1:

```
  [FAIL] [sig-network] Services [It] should allow pods to hairpin back to themselves through services
[...]

PR #1598, opened 2024-01-11
Feature: Support All Upstream Traffic Policies

aauren:
Adds support for `spec.internalTrafficPolicy` (introduced in k8s v1.26) and fixes `spec.externalTrafficPolicy` to behave according to upstream indicated behavior: https://kubernetes.io/docs/reference/networking/virtual-ips/#traffic-policies

At this point, we still don't handle terminating endpoints as specified in the documentation. However, I think that this is a good bit of work, so its probably best to handle that in a future PR.

Relates to #1597 - However, #1597 won't be completely finished until we implement the specialized draining procedure and pass all traffic policy related e2e tests.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -51,9 +51,13 @@ issues:
     - text: "G306:"
       linters:
         - gosec
-    # Exlude tests from long line linter
+    # Exclude tests from long line linter
     - linters:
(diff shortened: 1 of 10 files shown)


aauren:
Also related to #1596

PR #1601, opened 2024-01-28
feat(cni): add binaries to host when missing

### LINKED ISSUE #1600: kube-router Should Own CNI Plugin Installation Process

In order for kube-router to better align with other Kubernetes networking providers and to improve the end-user experience, kube-router should bundle and install CNI plugins to the host's system when they are missing.

This should be in an initContainer to limit the host mounts that kube-router carries during it's runtime.

Additionally, user's should be able to opt-out of this process by manipulating the daemonset or setting a flag or some such.

**Describe alternatives you've considered**

kube-router could continue not owning this process, but it puts more onus on the user to make sure that this is available and given that most networking providers appear to do this, it violates the principle of least surprise for normal Kubernetes users.

**Additional context**

Antrea has kindly provided a link to their implementation for this process here: 
https://github.com/antrea-io/antrea/blob/main/build/images/scripts/install_cni

aauren:
@mrueg @AlexisDucastel

This PR does the following:

* Adds the cni-plugins package to the kube-router container
* Creates a script which will run in the initContainer and install either of the CNI plugins that kube-router is known to use
  * If the plugins are already found on the user's host filesystem and are executable in the standard location (`/opt/cni/bin`) then it will noop
* Updates all of the DaemonSets that have the kube-router initContainer to execute the `cni-install` script when it is present in the container (that way it won't block anyone that updates their daemonsets using an old container)

Fixes #1600


mrueg (approved):

Makefile
```diff
@@ -228,6 +230,14 @@ multiarch-binverify:
...
 	@[ `file kube-router gobgp| cut -d, -f2 |grep -cw "$(FILE_ARCH)"` -eq 2 ]
 
+cni-download:
+	@echo Downloading CNI Plugins for $(GOARCH)
+		curl -L -o cni-plugins-$(GOARCH).tgz \
+			https://github.com/containernetworking/plugins/releases/download/$(CNI_VERSION)/cni-plugins-linux-$(GOARCH)-$(CNI_VERSION).tgz
+		mkdir -p cni-download
+		tar -xf cni-plugins-$(GOARCH).tgz -C cni-download
+		rm -f cni-plugins-$(GOARCH).tgz
+
```
aauren, line 240: @mrueg after trying a couple of approaches, this is the way I decided to go. Let me know if you have any thoughts.

mrueg: Yeah that seems fine, if you wanna avoid cURL, you could use ADD with a url as well.

https://docs.docker.com/engine/reference/builder/#add

aauren:
This has been verified to be working by an external tester. Merging.

PR #1612, opened 2024-03-01
Make kube router healthchecknodeport aware

aauren:
The big change here is that kube-router now implements the NodePort HealthCheck feature that is best described here: https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-type-loadbalancer

Blurb from the commit for this feature is:

>NodePort Health Check has long been part of the Kubernetes API, but
kube-router hasn't implemented it in the past. This is meant to be a
port that is assigned by the kube-controller-manager for LoadBalancer
services that have a traffic policy of `externalTrafficPolicy=Local`.

>When set, the k8s networking implementation is meant to open a port and
provide HTTP responses that inform parties external to the Kubernetes
cluster about whether or not a local endpoint exists on the node. It
should return a 200 status if the node contains a local endpoint and
return a 503 status if the node does not contain a local endpoint.

>This allows applications outside the cluster to choose their endpoint in
such a way that their source IP could be preserved.

Additionally, in order to become more compliant with other network providers and the upstream e2e tests, we now also add endpoints to load balancers for EndpointSlices that are either `Ready` or `Serving` as opposed to just `Ready`. This allow traffic to route to legitimate service endpoints that are in the `Terminating` state.

This is a departure from previous functionality, so there may be some impact to services that were relying on the previous `Ready` only interpretation for endpoint adding.

Fixes: #1597

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -243,6 +243,7 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
 		klog.Fatalf("error cleaning up old/bad masquerade rules: %s", err.Error())
 	}
 
+	// Run the hairpin controller
 	wg.Add(1)
 	go nsc.hpc.Run(stopCh, wg, healthChan)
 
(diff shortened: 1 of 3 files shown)


aauren:
Related to: #1596

PR #1618, opened 2024-03-18
dependabot: Group kubernetes dependencies

mrueg:
@aauren this will reduce the number of PRs that dependabot makes for the Kubernetes deps.


aauren:
Thanks @mrueg !

Issue #1616, opened 2024-03-18
Failed to setup policy routing required for DSR due to open /etc/iproute2/rt_tables: no such file or directory


s11fsch:
**What happened?**
I have a kubeadm cluster with currently two nodes that should connect over the internet. The service routing between them does not work (times out). Looking in the kube-router pod log I see:
```
E0318 07:56:22.845315  109570 service_endpoints_sync.go:99] Error setting up necessary policy based routing configuration needed for direct server return: Failed setup PBR for DSR due to: failed to setup policy routing required for DSR due to open /etc/iproute2/rt_tables: no such file or directory
```

**What did you expect to happen?**
The routing should work and no error should occur.

**How can we reproduce the behavior you experienced?**
1. Set-up two node cluster using kubeadm
2. Install kube-router: https://github.com/cloudnativelabs/kube-router/blob/master/docs/kubeadm.md
3. See kube-router logs

**System Information (please complete the following information):**
 - Kube-Router Version (`kube-router --version`): v2.1.0, built on 2024-03-02T21:50:00+0000, go1.21.7
 - Kube-Router Parameters: `--run-router=true --run-firewall=true --run-service-proxy=true --bgp-graceful-restart=true --kubeconfig=/var/lib/kube-router/kubeconfig`
 - Kubernetes Version (`kubectl version`) : `Client Version: v1.29.1`, `Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3`, `Server Version: v1.29.3`
 - Cloud Type: on premise
 - Kubernetes Deployment Type: Kubeadm
 - Kube-Router Deployment Type: DaemonSet
 - Cluster Size: 2 Nodes

**Logs, other output, metrics**
`k logs kube-router-kqhsj -n kube-system`
```
I0318 07:46:22.494016  109570 version.go:66] Running /usr/local/bin/kube-router version v2.1.0, built on 2024-03-02T21:50:00+0000, go1.21.7
I0318 07:46:22.595860  109570 kube-router.go:137] Metrics port must be over 0 and under 65535 in order to be enabled, given port: 0
I0318 07:46:22.595900  109570 kube-router.go:139] Disabling metrics for kube-router, set --metrics-port properly in order to enable
[...]

PR #1627, opened 2024-03-18
doc(DSR): add /etc/iproute2/rt_tables caveat

aauren:
Add documentation to improve user DSR experience about mounting `/etc/iproute2/rt_tables` into the container when kube-router runs as a daemonset.

Based upon feedback from #1616 from @s11fsch


s11fsch (approved):
Looks good to me

mrueg (approved):

PR #1642, opened 2024-04-10
install iptables-legacy so the iptables-wrapper can use it

### LINKED ISSUE #1641: container image v2.1.0 is missing the iptables-legacy binaries
**What happened?**

Kube-router v2.1.0 uses iptables with the nftables backend even when the host is using iptables with the legacy backend.

Since the iptables-wrapper is installed in the container image I expected kube-router to use the same backend as the host system.

1. Have a host system with iptables-legacy and a Kubernetes cluster. 
2. Run `iptables --version` on the host. Example output: `iptables v1.8.9 (legacy)`
3. Repeat the previous command in a kube-router container (`kubectl --namespace kube-system exec daemonsets/kube-router -- iptables --version`). Output: `iptables v1.8.10 (nf_tables)`

**System Information (please complete the following information):**
 - Kube-Router Version (`kube-router --version`): 2.1.0

**Additional context**

When I downgrade kube-router to v2.0.1 the correct iptables backend is used.

According to the iptables-wrapper documentation the `iptables-legacy` package needs to be installed in containers that use an Alpine 3.19 base image: <https://github.com/kubernetes-sigs/iptables-wrappers?tab=readme-ov-file#building-a-container-image-that-uses-iptables>

elchenberg:
* Installs iptables-legacy so the iptables-wrapper can use it.
* No longer installs the ip6tables package because it is already provided by the iptables package.
* Adds a check for the iptables-nft and iptables-legacy binary before running the iptables-wrapper-installer.sh.
* Removes nsswitch.conf creation because the file already exists in the Alpine 3.19 base image.

Please let me know what you think. :)

Closes #1641.


aauren:
Thanks! @mrueg are you able to take a look at this?

Dockerfile
```diff
@@ -42,14 +42,17 @@ COPY --from=builder /build/cni-download /usr/libexec/cni
...
 # This is necessary because of the bug reported here: https://github.com/flannel-io/flannel/pull/1340/files
 # Basically even under QEMU emulation, it still doesn't have an ARM kernel in-play which means that calls to
-# iptables-nft will fail in the build process. The sanity check here only makes sure that we are not using
-# iptables-1.8.0-1.8.2. For now we'll manage that on our own.
-RUN /iptables-wrapper-installer.sh --no-sanity-check
-
-
-# Since alpine image doesn't contain /etc/nsswitch.conf, the hosts in /etc/hosts (e.g. localhost)
-# cannot be used. So manually add /etc/nsswitch.conf to work around this issue.
-RUN echo "hosts: files dns" > /etc/nsswitch.conf
```
mrueg, line 52: Do we still need this?

mrueg:
Changes make sense to me, similar to what was changed here: https://github.com/kubernetes-sigs/iptables-wrappers/pull/8

I also saw that https://github.com/kubernetes-sigs/iptables-wrappers now contains a go-binary. Do we want to sync this back into our repository (or simply fetch it from upstream)?

Dockerfile
aauren: No, we shouldn't need it, like @elchenberg mentioned it is now default this way via the Alpine base image.

aauren:
> I also saw that https://github.com/kubernetes-sigs/iptables-wrappers now contains a go-binary. Do we want to sync this back into our repository (or simply fetch it from upstream)?

Hmm... I agree and would like to install it the same way that upstream does it, but we would need to copy over the right arch for each of our builds if we were to adopt their go version of the iptables-wrapper. The easiest way to do that would be to have a container image that we could copy it from, but I don't see anywhere where they are building it.

Given all of that, it may be easiest to continue with this script we have for now. Can you think of an easy way to get the right version of the iptables-wrapper go script into our container @mrueg ?

elchenberg:
Regarding the go version of the iptables-wrapper, you could build it in the builder stage, for example:

```diff
diff --git a/Dockerfile b/Dockerfile
index 843a0484..ef2c8b39 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,15 @@ RUN apk add --no-cache make git tar curl \
     && make gobgp \
     && make cni-download
 
+WORKDIR /iptables-wrappers
+# This is the latest commit on the master branch.
+ENV IPTABLES_WRAPPERS_VERSION=f6ef44b2c449cca8f005b32dea9a4b497202dbef
+RUN git clone https://github.com/kubernetes-sigs/iptables-wrappers.git . \
+    && git checkout "${IPTABLES_WRAPPERS_VERSION}" \
+    && make build \
+    && test -x bin/iptables-wrapper \
+    && test -x iptables-wrapper-installer.sh
+
 FROM ${RUNTIME_BASE}
 
 RUN apk add --no-cache \
@@ -39,7 +48,8 @@ COPY --from=builder /build/cni-download /usr/libexec/cni
 # Use iptables-wrappers so that correct version of iptables-legacy or iptables-nft gets used. Alpine contains both, but
 # which version is used should be based on the host system as well as where rules that may have been added before
 # kube-router are being placed. For more information see: https://github.com/kubernetes-sigs/iptables-wrappers
-COPY build/image-assets/iptables-wrapper-installer.sh /
+COPY --from=builder /iptables-wrappers/bin/iptables-wrapper /
+COPY --from=builder /iptables-wrappers/iptables-wrapper-installer.sh /
 # This is necessary because of the bug reported here: https://github.com/flannel-io/flannel/pull/1340/files
 # Basically even under QEMU emulation, it still doesn't have an ARM kernel in-play which means that calls to
 # iptables-nft will fail in the build process. The sanity check here only makes sure that iptables-nft and iptables-legacy
```

If you want I can add it to this PR or open a new PR.

aauren:
> If you want I can add it to this PR or open a new PR.

Thanks @elchenberg!

That looks good to me. Go ahead and add it to this PR and we'll merge it in.

aauren:
@elchenberg - Any chance you have time to come back around to this? Or do you want me to add in your changes above?

elchenberg:
@aauren Feel free to add the changes. I would probably not be able to do it before the weekend.

aauren:
Thanks so much for the contribution @elchenberg!

I tested it on my system and it worked great for me. I'll combine this with a few other bug fixes that I've been working on in the next week or so and release a v2.1.1 soon.

Issue #1649, opened 2024-04-20
CLI Option `--cleanup-config` Is Not Working Correctly


aauren:
**What happened?**
@vladimirtiukhtin reported the following here: https://github.com/cloudnativelabs/kube-router/issues/1617#issuecomment-2066114076

>Unrelated to the subj, but might be worth to look at

```sh
# kube-router --cleanup-config
I0419 08:38:35.287507  288326 network_policy_controller.go:713] Cleaning up NetworkPolicyController configurations...
I0419 08:38:35.287552  288326 network_policy_controller.go:746] Successfully cleaned the NetworkPolicyController configurations done by kube-router
I0419 08:38:35.287575  288326 network_services_controller.go:1755] Cleaning up NetworkServiceController configurations...
I0419 08:38:35.289694  288326 network_services_controller.go:1762] ipvs definitions don't have names associated with them for checking, during cleanup we assume that we own all of them and delete all ipvs definitions
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x98 pc=0x180e7d2]

goroutine 1 [running]:
github.com/cloudnativelabs/kube-router/v2/pkg/controllers/proxy.(*NetworkServicesController).deleteHairpinIptablesRules(0xc000628010?, {0x2262d11?, 0x0?})
	/home/runner/work/kube-router/kube-router/pkg/controllers/proxy/network_services_controller.go:1436 +0x52
github.com/cloudnativelabs/kube-router/v2/pkg/controllers/proxy.(*NetworkServicesController).Cleanup(0xc00061f8b0?)
	/home/runner/work/kube-router/kube-router/pkg/controllers/proxy/network_services_controller.go:1779 +0x32b
github.com/cloudnativelabs/kube-router/v2/pkg/cmd.CleanupConfigAndExit()
	/home/runner/work/kube-router/kube-router/pkg/cmd/kube-router.go:71 +0x85
main.Main()
	/home/runner/work/kube-router/kube-router/cmd/kube-router/kube-router.go:65 +0x406
main.main()
	/home/runner/work/kube-router/kube-router/cmd/kube-router/kube-router.go:21 +0x13
```

PR #1650, opened 2024-04-21
Fix Cleanup Code for v2.x

aauren:
kube-router v2.X introduced the idea of iptables and ipset handlers that allow kube-router to be dual-stack capable. However, the cleanup logic for the various controllers was not properly ported when this happened. When the cleanup functions run, they often have not had their controllers fully initialized as cleanup should not be dependant on kube-router being able to reach a kube-apiserver.

As such, they were missing these handlers. And as such they either silently ended up doing noops or worse, they would run into nil pointer failures.

This corrects that, so that kube-router no longer fails this way and cleans up as it had in v1.X.

@vladimirtiukhtin - I tested this in my environment and found it worked correctly, but if you wouldn't mind checking as well, that would be helpful to ensure that there isn't something I'm missing. There should be a kube-router docker container that is built as part of this PR that you can find in the GitHub actions section if you don't want to build kube-router yourself.

Fixes: #1649

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -207,8 +207,23 @@ kube-router --master=http://192.168.1.99:8080/ --run-firewall=true --run-service
 Please delete kube-router daemonset and then clean up all the configurations done (to ipvs, iptables, ipset, ip routes
 etc) by kube-router on the node by running below command.
 
+### Docker
+
 ```sh
-docker run --privileged --net=host cloudnativelabs/kube-router --cleanup-config
+docker run --privileged --net=host \
+--mount type=bind,source=/lib/modules,target=/lib/modules,readonly \
+--mount type=bind,source=/run/xtables.lock,target=/run/xtables.lock,bind-propagation=rshared \
+cloudnativelabs/kube-router /usr/local/bin/kube-router --cleanup-config
+```
+
+### containerd
+
+```sh
+$ ctr image pull docker.io/cloudnativelabs/kube-router:latest
+$ ctr run --privileged -t --net-host \
+--mount type=bind,src=/lib/modules,dst=/lib/modules,options=rbind:ro \
+--mount type=bind,src=/run/xtables.lock,dst=/run/xtables.lock,options=rbind:rw \
+docker.io/cloudnativelabs/kube-router:latest kube-router-cleanup /usr/local/bin/kube-router --cleanup-config
 ```
 
 ## trying kube-router as alternative to kube-proxy
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -712,6 +712,26 @@ func (npc *NetworkPolicyController) cleanupStaleIPSets(activePolicyIPSets map[st
 func (npc *NetworkPolicyController) Cleanup() {
 	klog.Info("Cleaning up NetworkPolicyController configurations...")
 
+	if len(npc.iptablesCmdHandlers) < 1 {
+		iptablesCmdHandlers, ipSetHandlers, err := NewIPTablesHandlers(nil)
+		if err != nil {
+			klog.Errorf("unable to get iptables and ipset handlers: %v", err)
+			return
+		}
+		npc.iptablesCmdHandlers = iptablesCmdHandlers
+		npc.ipSetHandlers = ipSetHandlers
+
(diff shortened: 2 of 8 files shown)


vladimirtiukhtin:
@aauren it works. Not without errors though
```
# kube-router --cleanup-config
I0422 08:55:24.774859       7 network_policy_controller.go:713] Cleaning up NetworkPolicyController configurations...
I0422 08:55:25.432990       7 network_policy_controller.go:766] Successfully cleaned the NetworkPolicyController configurations done by kube-router
I0422 08:55:25.433718       7 network_services_controller.go:1755] Cleaning up NetworkServiceController configurations...
WARN[0000] Running modprobe ip_vs failed with message: `modprobe: can't change directory to '/lib/modules': No such file or directory`, error: exit status 1 
I0422 08:55:25.437538       7 network_services_controller.go:1762] ipvs definitions don't have names associated with them for checking, during cleanup we assume that we own all of them and delete all ipvs definitions
I0422 08:55:25.446883       7 network_services_controller.go:1790] Processing IPv4 hairpin rule cleanup
I0422 08:55:25.449415       7 network_services_controller.go:1798] Processing IPv6 hairpin rule cleanup
E0422 08:55:25.470290       7 network_services_controller.go:566] Failed to run iptables command: running [/sbin/iptables -t filter -X KUBE-ROUTER-SERVICES --wait]: exit status 4: iptables v1.8.10 (nf_tables):  CHAIN_DEL failed (Resource busy): chain KUBE-ROUTER-SERVICES
I0422 08:55:26.553989       7 network_services_controller.go:1822] Successfully cleaned the NetworkServiceController configuration done by kube-router
I0422 08:55:26.554088       7 network_routes_controller.go:897] Cleaning up NetworkRoutesController configurations
I0422 08:55:26.561566       7 pod_egress.go:73] Deleted iptables rule to masquerade outbound traffic from pods.
W0422 08:55:27.626388       7 network_routes_controller.go:947] Error deleting ipset: ipset v7.19: Set cannot be destroyed: it is in use by a kernel component
W0422 08:55:27.646544       7 network_routes_controller.go:947] Error deleting ipset: ipset v7.19: The set with the given name does not exist
I0422 08:55:27.646589       7 network_routes_controller.go:951] Successfully cleaned the NetworkRoutesController configuration done by kube-router
```

aauren:
So on my cluster, I was able to run without issues if I add in a bind mount for the `/lib/modules` directory (which is what the deployments in k8s do for kube-router) with the following:

```
sudo ctr run --privileged -t --mount type=bind,src=/lib/modules,dst=/lib/modules,options=rbind:ro --net-host docker.io/cloudnativelabs/kube-router-git:PR-1650@sha256:4109d544ae89cb08a694baf325bc54f15f449a3f36ae58bfe1be537a3b5f7e2d test /usr/local/bin/kube-router --cleanup-config
```

The ipset error are likely a symptom of the referencing iptables not being removed. Those errors are caused by this log line:

```
E0422 08:55:25.470290       7 network_services_controller.go:566] Failed to run iptables command: running [/sbin/iptables -t filter -X KUBE-ROUTER-SERVICES --wait]: exit status 4: iptables v1.8.10 (nf_tables):  CHAIN_DEL failed (Resource busy): chain KUBE-ROUTER-SERVICES
```

I can't reproduce that on my node, but that is probably due to not having the same contention over iptables as your cluster likely has. You mentioned in another post that you have configured some jobs to run with a frequency of 1 minute. If those types of nodes are the same ones where you are testing this, then you likely have a lot more going on with iptables than I do in my cluster.

My guess is that `--wait` isn't fully taking effect because the `xtables.lock` isn't being mounted into the container. I would try running with that mounted as well like the daemonset runs. So something like this:

```
sudo ctr run --privileged -t --mount type=bind,src=/lib/modules,dst=/lib/modules,options=rbind:ro --mount type=bind,src=/run/xtables.lock,dst=/run/xtables.lock,options=rbind:rw --net-host docker.io/cloudnativelabs/kube-router-git:PR-1650@sha256:4109d544ae89cb08a694baf325bc54f15f449a3f36ae58bfe1be537a3b5f7e2d test /usr/local/bin/kube-router --cleanup-config
```

See if that works for you?

aauren:
I have updated the user-guide to show a better version of the cleanup command which should resolve any errors for most users.

PR #1651, opened 2024-04-22
fix(DSR): setup DSR inside pod on local eps only

### LINKED ISSUE #1640: DSR mode: "rpc error: code = NotFound desc = could not find container "": container ID should not be empty"
**What happened?**
Unable to enable DSR mode with kube-router v2.1.0 (atleast with cri-o runtime):

```
W0404 06:42:26.785234  345351 linux_networking.go:102] got an IfaceHasNoAddr error while trying to delete address from netlink: cannot assign requested address (this is not normally bad enough to stop processing)
W0404 06:42:26.785901  345351 linux_networking.go:117] got a No such process error while trying to remove route: exit status 2 (this is not normally bad enough to stop processing)
I0404 06:42:26.788147  345351 linux_networking.go:758] Successfully assigned VIP: 172.16.69.69 in endpoint 192.168.3.68.
I0404 06:42:26.788255  345351 linux_networking.go:780] Successfully disabled rp_filter in endpoint 192.168.3.68.
I0404 06:42:26.788276  345351 linux_networking.go:790] Current network namespace after revert namespace to host network namespace: NS(15: 4, 4026531840)
E0404 06:42:26.789164  345351 service_endpoints_sync.go:60] Error setting up IPVS services for service external IP's and load balancer IP's: failed to setup DSR endpoint 172.16.69.69: unable to setup DSR receiver inside pod: failed to prepare endpoint 192.168.5.75 to do DSR due to: rpc error: code = NotFound desc = could not find container "": container ID should not be empty
```

DSR mode enabled without errors.

1. Install kubeadm k8s cluster without kube-proxy and with cri-o runtime.
2. Deploy kube-router v2.1.0 using kubeadm-kuberouter-all-features-dsr.yaml (https://github.com/cloudnativelabs/kube-router/blob/master/daemonset/kubeadm-kuberouter-all-features-dsr.yaml).
3. Make sure to make following changes in kubeadm-kuberouter-all-features-dsr.yaml:  
- set `--runtime-endpoint=unix:///run/crio/crio.sock` 
-  replace `/var/run/docker.sock` with `/run/crio/crio.sock` in volumeMounts and volumes configuration
- instead of:
```
      - name: kubeconfig
        configMap:
          name: kube-proxy
          items:
          - key: kubeconfig.conf
            path: kubeconfig
```
replace with:
```
      - name: kubeconfig
        hostPath:
          path: /var/lib/kube-router
```

 - Kube-Router Version (`kube-router --version`): v2.1.0
 - Kube-Router Parameters:         - --run-router=true - --run-firewall=true  - --run-service-proxy=true- --bgp-graceful-restart=true - --kubeconfig=/var/lib/kube-router/kubeconfig - --runtime-endpoint=unix:///run/crio/crio.sock
 - Kubernetes Version (`kubectl version`) : 1.29.2 (container runtime CRI-O v1.29.2)
 - Cloud Type: on premise
 - Kubernetes Deployment Type: Kubeadm
 - Cluster Size: 6 nodes

**Additional context**
It seems to be a problem with kube-router v2.x release. I also tested same configuration with kube-router v1.6.0 and DSR works like a charm.

aauren:
Only attempt to setup DSR inside containers for local endpoints. Setting up DSR inside the containers network namespace requires local pods / endpoints.

Fixes #1640 

@opipenbe - I tested this locally with my smaller development cluster and it appeared to work correctly, however, I would be very appreciative if you would test this in your cluster as well. A container should be built as part of the GitHub Actions CI on this PR, you can look in the CI log for a link to [kube-router-git](https://hub.docker.com/r/cloudnativelabs/kube-router-git/tags) for the newly built kube-router.

One thing that I still don't quite have an answer for though, is why this didn't also have this same error for you under kube-router v1.6.0 as the code path was essentially the same: https://github.com/cloudnativelabs/kube-router/blob/v1.6/pkg/controllers/proxy/service_endpoints_sync.go#L444-L447

Maybe something further up in the calling tree filtered it out?

Anyway, since to me, this seems like it is changing DSR logic that has existed for a long time, I'd appreciate the second test to ensure that it not only removes the error, but also still allows you to route to your service correctly.

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -289,6 +289,20 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
 	// https://github.com/kubernetes/kubernetes/pull/70530/files
 	setSysCtlAndCheckError(utils.IPv4ConfAllArpAnnounce, arpAnnounceUseBestLocalAddress)
 
+	// Ensure rp_filter=2 for DSR capability, see:
+	// * https://access.redhat.com/solutions/53031
+	// * https://github.com/cloudnativelabs/kube-router/pull/1651#issuecomment-2072851683
+	if nsc.isIPv4Capable {
+		sysctlErr := utils.SetSysctlSingleTemplate(utils.IPv4ConfRPFilterTemplate, "all", 2)
+		if sysctlErr != nil {
+			if sysctlErr.IsFatal() {
+				klog.Fatal(sysctlErr.Error())
+			} else {
+				klog.Error(sysctlErr.Error())
+			}
+		}
+	}
+
 	// https://github.com/cloudnativelabs/kube-router/issues/282
 	err = nsc.setupIpvsFirewall()
 	if err != nil {
--- a/pkg/controllers/proxy/service_endpoints_sync.go
+++ b/pkg/controllers/proxy/service_endpoints_sync.go
@@ -564,9 +564,13 @@ func (nsc *NetworkServicesController) setupExternalIPForDSRService(svc *serviceI
 				endpoint.ip, externalIP, err)
 		}
 
-		// add the external IP to a virtual interface inside the pod so that the pod can receive it
-		if err = nsc.addDSRIPInsidePodNetNamespace(externalIP.String(), endpoint.ip); err != nil {
-			return fmt.Errorf("unable to setup DSR receiver inside pod: %v", err)
+		// It's only for local endpoints that we can enter the container's namespace and add DSR receivers inside it.
+		// If we aren't local, then we should skip this step so that we don't accidentally throw an error.
+		if endpoint.isLocal {
+			// add the external IP to a virtual interface inside the pod so that the pod can receive it
+			if err = nsc.addDSRIPInsidePodNetNamespace(externalIP.String(), endpoint.ip); err != nil {
+				return fmt.Errorf("unable to setup DSR receiver inside pod: %v", err)
+			}
 		}
 
 		svcEndpointMap[externalIPServiceID] = append(svcEndpointMap[externalIPServiceID],
--- a/pkg/controllers/proxy/service_endpoints_sync.go
+++ b/pkg/controllers/proxy/service_endpoints_sync.go
@@ -476,7 +476,7 @@ func (nsc *NetworkServicesController) setupExternalIPForDSRService(svc *serviceI
 
 	dummyVipInterface, err := nsc.ln.getKubeDummyInterface()
 	if err != nil {
-		return errors.New("Failed creating dummy interface: " + err.Error())
+		return errors.New("Failed getting dummy interface: " + err.Error())
 	}
 
 	ipvsSvcs, err := nsc.ln.ipvsGetServices()


opipenbe:
Thank you @aauren for the great work! 

I just tested with PR-1651 image in my environment and I can confirm DSR feature now works correctly! I also did full environment restart in order to be sure this change does fix DSR.

aauren:
Hmm... In my cluster, the external IPs route within the cluster just fine DSR or no DSR.

Inside the cluster, for non-service-local VIPs every node should have an IPVS definition for the service, so I would imagine that it should "just work".

mrueg (approved):

opipenbe:
Thanks @aauren for clarification! I figured out why DSR packets are dropped in my environment - in the RHEL / Rocky Linux, multihoming is not enabled by default. DSR makes routing asymmetric, so rp_filter (default value 1) will drop DSR: https://access.redhat.com/solutions/53031 .

So I can also confirm that DSR is working successfully inside cluster after configuring kernel parameter `rp_filter = 2` for all nodes :)

aauren:
@opipenbe Thanks for taking the deep dive on this!

I have added a commit to this, that sets rp_filter to 2 during the startup of kube-router similar to other sysctl values that we do sanity checks on.

dsseng:
> @aauren , I tested DSR outside of the cluster since my service external IP is exposed from all worker nodes by BGP.
> 
> I also didnt find anything suspicious in the kube-router logs.
> 
> I dont know a good method to test DSR inside cluster. As I understand, it is by design that packets are dropped from cluster to external IPs with DSR feature? If It is somehow possible to test DSR inside the cluster, then let me know and I can test it also :)

Wouldn't setting it to 2 only if it is already 1 be better? kube-router sets this sysctl to 2 unconditionally (while typical kernel default is 0), which breaks our KubeSpan (https://github.com/siderolabs/talos/issues/9814), while 0 is more permissive than 2 and should allow kube-router to operate on our default settings.

So I believe checking existing values should be better so that kube-router doesn't tighten up the filtering policy potentially breaking other networking. I am also investigating potential improvements on the Talos side so that KubeSpan would work even with rp_filter=2 or 1

aauren:
@dsseng - This seems reasonable to me.

Do you have time to create a PR for this change?

dsseng:
Thank you for your response. I will perhaps contribute this change after completing the Talos part of work (since that should give us other benefits).

There's one culprit, see the different options:
1. all = 0, the desired interface is also 0 - no need for override
2. all = 1 - override anyway
3. all = 2 - no need for override
4. all = 0, but some of the specific interface values are 1 - we can't check all interfaces and due to this we miss setting the override.

What specific interfaces should have rp_filter in weak mode?

aauren:
That's some good foresight. Right now we most likely only need to worry about the `all` interface, as that's the way the functionality works now and we haven't gotten any bugs reported.

However, it wouldn't hurt to try to cover our bases.

I would recommend that we check the primary interface (as retrieved by: `NodeInterfaceAware.GetNodeInterfaceName()` https://github.com/cloudnativelabs/kube-router/blob/master/pkg/utils/node.go#L55) and `kube-bridge` at the very least and ensure that they are set the same way `all` is set.

dsseng:
Well, as docs state the value at the moment of filter run is the maximum of all and specific interface values. So if all is 0 and interface-specific is 1, then the behavior is 1, but if you set all to 2 it makes behavior for all the interfaces to be as 2.

So I'm not really sure about a viable and fully-featured solution from the kube-router side, yet thankfully I have isolated the problem from KubeSpan & Talos side, so CNI behavior cannot disrupt the connectivity.

dsseng:
Okay, marking packets incoming from the KubeSpan WireGuard interface and then setting the `conf/kubespan/src_valid_mark` sysctl to 1 helped resolve the issue on Talos. https://github.com/siderolabs/talos/pull/10074 fixes this on our side and seems to work even with strict filtering. Will do a fix on the Kube-Router side to ensure such disruptions don't happen on other systems

Issue #1630, opened 2024-04-01
kube-router can't find eth0 interface


mtb-xt:
**What happened?**
I can see this output in my kube-router logs
I'm trying to get DSR to work on my cluster.

> I0401 00:45:56.355730   20878 network_policy_controller.go:163] Starting network policy controller
W0401 00:45:56.421160   20878 linux_networking.go:637] Able to see the following interfaces: enp1s0 kube-bridge kube-dummy-if lo veth0353ae9f veth454d77b5 veth47bce93a veth9da5270f vethc27ffbcc vethc557e15c vethcdf11df4 
W0401 00:45:56.421190   20878 linux_networking.go:638] If one of the above is not eth0 it is likely, that the assumption that we've hardcoded in kube-router is wrong, please report this as a bug along with this output
E0401 00:45:56.421271   20878 hairpin_controller.go:46] unable to set hairpin mode for endpoint 172.21.100.21, its possible that hairpinning will not work as expected. Error was: failed to find the interface ID inside the container NS for endpoint IP: 172.21.100.21, due to: unable to read the ifaceID inside the container from /proc/4755/cwd/sys/class/net/eth0/iflink, output was: , error was: open /proc/4755/cwd/sys/class/net/eth0/iflink: no such file or directory

**What did you expect to happen?**
None of my interfaces are called 'eth0', unfortunately, so I'd expect kube-router to detect the interface name in some other way.

**How can we reproduce the behavior you experienced?**
1. Use on premises nodes with different interface names
2. Try to enable DSR

**Screenshots / Architecture Diagrams / Network Topologies**
Standard BGP full mesh in kube-router.

**System Information (please complete the following information):**
 - Kube-Router Version (`kube-router --version`): v2.1.0, built on 2024-03-02T21:50:00+0000, go1.21.7
 - Kube-Router Parameters: [e.g. --run-router --run-service-proxy --enable-overlay --overlay-type=full etc.]
        - --run-router=true
        - --run-firewall=true
        - --run-service-proxy=true
        - --bgp-graceful-restart=true
        - --hairpin-mode
        - --metrics-port=9010
        - --kubeconfig=/var/lib/kube-router/kubeconfig
        - --runtime-endpoint=unix:///run/containerd/containerd.sock
 - Kubernetes Version (`kubectl version`) : 1.28.2
 - Cloud Type: on premise
 - Kubernetes Deployment Type: k0s
 - Kube-Router Deployment Type: DaemonSet
[...]

PR #1657, opened 2024-04-27
Remove Hairpin Controller

aauren:
After taking another look at it, I realized that the hairpin controller that I had originally created to ensure that hairpin mode was enabled on the pod's network namespace link was entirely superfluous. Instead, user's should be using the little known `hairpinMode` of the bridge CNI plugin for kube-router.

This plugin is already in the pod's network namespace when it needs to run and is much more reliable than the controller that I created.

For now I'm leaving the code, but just removing references to it so that I can bring it back in the future if I decide that its helpful. However, it had quite a few problems with it, as mentioned in #1630.

I've also updated the dsr reference daemonset to include this mode, as I'm not sure why this was left out in the first place. The docs in the user-guide mention this CNI option quite a bit already, so it was already well documented there.

### DIFF
--- a/daemonset/kubeadm-kuberouter-all-features-dsr.yaml
+++ b/daemonset/kubeadm-kuberouter-all-features-dsr.yaml
@@ -17,6 +17,7 @@ data:
              "type":"bridge",
              "bridge":"kube-bridge",
              "isDefaultGateway":true,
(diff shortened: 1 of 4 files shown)


aauren:
Fixes #1630

PR #1672, opened 2024-05-16
fix(nsc): TCPMSS rules are created per tcp service

### LINKED ISSUE #1671: v2.1: DSR+TCPMSS with non-ready services not set-up correctly
**What happened?**
Have multiple services with the same externalIP, set up DSR for them. When one of the services are not ready, maxseg rules in nft for other services are missing. ref: https://github.com/cloudnativelabs/kube-router/issues/1664

Expected to work as in v1.6.

1. Create a deployment+service with dsr and externalIP, e.g.:
```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: ght
spec:
  selector:
    matchLabels:
      app: ght
  template:
    metadata:
      labels:
        app: ght
    spec:
      containers:
      - image: ghcr.io/rkojedzinszky/go-http-test
        name: go-http-test
---
apiVersion: v1
kind: Service
metadata:
  name: ght
  annotations:
    kube-router.io/service.dsr: tunnel
spec:
  externalIPs:
    - 192.168.9.88
  ports:
  - name: http
    port: 8080
    protocol: TCP
    targetPort: 8080
  selector:
    app: ght
```
2. Observe correct behavior, tcpdump on node with pod:
```
# nsenter -n -t 3898354 tcpdump -pvni any host 192.168.9.88
tcpdump: data link type LINUX_SLL2
tcpdump: listening on any, link-type LINUX_SLL2 (Linux cooked v2), snapshot length 262144 bytes
07:48:15.753384 kube-tunnel-if In  IP (tos 0x0, ttl 61, id 7374, offset 0, flags [DF], proto TCP (6), length 60)
    192.168.10.10.58322 > 192.168.9.88.8080: Flags [S], cksum 0x653b (correct), seq 3111575680, win 64240, options [mss 1440,sackOK,TS val 1197502074 ecr 0,nop,wscale 7], length 0
07:48:15.753490 eth0  Out IP (tos 0x0, ttl 64, id 0, offset 0, flags [DF], proto TCP (6), length 60)
    192.168.9.88.8080 > 192.168.10.10.58322: Flags [S.], cksum 0x94e1 (incorrect -> 0x531e), seq 1899687061, ack 3111575681, win 65160, options [mss 1460,sackOK,TS val 2093888449 ecr 1197502074,nop,wscale 7], length 0
07:48:15.755483 kube-tunnel-if In  IP (tos 0x0, ttl 61, id 7375, offset 0, flags [DF], proto TCP (6), length 52)
    192.168.10.10.58322 > 192.168.9.88.8080: Flags [.], cksum 0x7e7a (correct), ack 1, win 502, options [nop,nop,TS val 1197502077 ecr 2093888449], length 0
```

With default MTU of 1500, 40 (ip+tcp header) and an extra 20 (ipip) bytes are substracted, thus 1440 is seen in initial SYN packet.

3. Create a non-ready service, e.g.:
```
---
apiVersion: v1
kind: Service
metadata:
  name: ght-nr
  annotations:
    kube-router.io/service.dsr: tunnel
spec:
  externalIPs:
    - 192.168.9.88
  ports:
  - name: http-nr
    port: 9090
    protocol: TCP
    targetPort: 9090
  selector:
    app: ght-nr
```

4. Observe incorrect behavior, tcpdump on node with pod:
```
# nsenter -n -t 3898354 tcpdump -pvni any host 192.168.9.88
tcpdump: data link type LINUX_SLL2
tcpdump: listening on any, link-type LINUX_SLL2 (Linux cooked v2), snapshot length 262144 bytes
07:52:28.330275 kube-tunnel-if In  IP (tos 0x0, ttl 61, id 64827, offset 0, flags [DF], proto TCP (6), length 60)
    192.168.10.10.57548 > 192.168.9.88.8080: Flags [S], cksum 0xe5d7 (correct), seq 3011939874, win 64240, options [mss 1460,sackOK,TS val 1197754650 ecr 0,nop,wscale 7], length 0
07:52:28.330380 eth0  Out IP (tos 0x0, ttl 64, id 0, offset 0, flags [DF], proto TCP (6), length 60)
    192.168.9.88.8080 > 192.168.10.10.57548: Flags [S.], cksum 0x94e1 (incorrect -> 0x1685), seq 3856751251, ack 3011939875, win 65160, options [mss 1460,sackOK,TS val 2094141026 ecr 1197754650,nop,wscale 7], length 0
07:52:28.332584 kube-tunnel-if In  IP (tos 0x0, ttl 61, id 64828, offset 0, flags [DF], proto TCP (6), length 52)
    192.168.10.10.57548 > 192.168.9.88.8080: Flags [.], cksum 0x41e1 (correct), ack 1, win 502, options [nop,nop,TS val 1197754653 ecr 2094141026], length 0
```

You can see incorrect mss in initial SYN packet.

nft rules also reflect this, as there are no maxseg rules for `ip daddr 192.168.9.88 tcp dport 8080`:

```
table ip mangle {
        chain KUBE-IPTABLES-HINT {
        }

        chain KUBE-KUBELET-CANARY {
        }

        chain PREROUTING {
                type filter hook prerouting priority mangle; policy accept;
                ip daddr 192.168.9.1 tcp dport 1883 counter packets 5661 bytes 582074 meta mark set 0x2e69
                ip daddr 192.168.9.1 tcp flags syn / syn,rst counter packets 65 bytes 3356 tcp option maxseg size set 1440
                ip daddr 192.168.9.1 tcp dport 3100 counter packets 4996 bytes 2186703 meta mark set 0x3e6d
                ip daddr 192.168.9.0 tcp dport 53 counter packets 0 bytes 0 meta mark set 0x234
                ip daddr 192.168.9.0 tcp flags syn / syn,rst counter packets 0 bytes 0 tcp option maxseg size set 1440
                ip daddr 192.168.9.0 udp dport 53 counter packets 173 bytes 11214 meta mark set 0x1e8c
                ip daddr 192.168.9.2 tcp dport 1222 counter packets 0 bytes 0 meta mark set 0xa15
                ip daddr 192.168.9.2 tcp flags syn / syn,rst counter packets 0 bytes 0 tcp option maxseg size set 1440
                ip daddr 192.168.8.192 tcp dport 443 counter packets 0 bytes 0 meta mark set 0x2f16
                ip daddr 192.168.8.192 tcp dport 80 counter packets 0 bytes 0 meta mark set 0x17f1
                ip daddr 192.168.8.192 tcp flags syn / syn,rst counter packets 0 bytes 0 tcp option maxseg size set 1440
                ip daddr 192.168.9.88 tcp dport 8080 counter packets 17 bytes 862 meta mark set 0x18be
        }

        chain OUTPUT {
...
```

 - Kube-Router Version (`kube-router --version`): v2.1.1
 - Kube-Router Parameters:
        - --run-router=true
        - --run-firewall=true
        - --run-service-proxy=true
        - --advertise-pod-cidr=false
        - --advertise-external-ip
        - --iptables-sync-period=24h
        - --ipvs-sync-period=24h
        - --routes-sync-period=24h
        - --auto-mtu=false
        - --service-external-ip-range=192.168.8.192/27
        - --service-external-ip-range=192.168.9.0/24
        - --runtime-endpoint=unix:///var/run/crio/crio.sock

 - Kubernetes Version (`kubectl version`) : v1.29.4
 - Cloud Type: bare metal
 - Kube-Router Deployment Type: daemonset
 - Cluster Size: 10 nodes

rkojedzinszky:
This allows VIPs to be shared between services.

Fixes #1671

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -1629,17 +1629,16 @@ func (nsc *NetworkServicesController) setupMangleTableRule(ip string, protocol s
 	}
 
 	// setup iptables rule TCPMSS for DSR mode to fix mtu problem
-	mtuArgs := []string{"-d", ip, "-m", tcpProtocol, "-p", tcpProtocol, "--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS",
-		"--set-mss", strconv.Itoa(tcpMSS)}
-	err = iptablesCmdHandler.AppendUnique("mangle", "PREROUTING", mtuArgs...)
-	if err != nil {
-		return fmt.Errorf("failed to run iptables command to set up TCPMSS due to %v", err)
+	// only reply packets from PODs are altered here
+	if protocol == tcpProtocol {
+		mtuArgs := []string{"-s", ip, "-m", tcpProtocol, "-p", tcpProtocol, "--sport", port,
+			"--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS", "--set-mss", strconv.Itoa(tcpMSS)}
+		err = iptablesCmdHandler.AppendUnique("mangle", "PREROUTING", mtuArgs...)
+		if err != nil {
+			return fmt.Errorf("failed to run iptables command to set up TCPMSS due to %v", err)
+		}
 	}
-	mtuArgs[0] = "-s"
-	err = iptablesCmdHandler.AppendUnique("mangle", "POSTROUTING", mtuArgs...)
-	if err != nil {
-		return fmt.Errorf("failed to run iptables command to set up TCPMSS due to %v", err)
-	}
+
 	return nil
 }
 
@@ -1678,29 +1677,20 @@ func (nsc *NetworkServicesController) cleanupMangleTableRule(ip string, protocol
 	}
 
 	// cleanup iptables rule TCPMSS
-	mtuArgs := []string{"-d", ip, "-m", tcpProtocol, "-p", tcpProtocol, "--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS",
-		"--set-mss", strconv.Itoa(tcpMSS)}
-	exists, err = iptablesCmdHandler.Exists("mangle", "PREROUTING", mtuArgs...)
-	if err != nil {
-		return fmt.Errorf("failed to cleanup iptables command to set up TCPMSS due to %v", err)
-	}
-	if exists {
-		klog.V(2).Infof("removing mangle rule with: iptables -D PREROUTING -t mangle %s", args)
-		err = iptablesCmdHandler.Delete("mangle", "PREROUTING", mtuArgs...)
+	// only reply packets from PODs are altered here
+	if protocol == tcpProtocol {
+		mtuArgs := []string{"-s", ip, "-m", tcpProtocol, "-p", tcpProtocol, "--sport", port,
+			"--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS", "--set-mss", strconv.Itoa(tcpMSS)}
+		exists, err = iptablesCmdHandler.Exists("mangle", "PREROUTING", mtuArgs...)
 		if err != nil {
 			return fmt.Errorf("failed to cleanup iptables command to set up TCPMSS due to %v", err)
 		}
-	}
-	mtuArgs[0] = "-s"
-	exists, err = iptablesCmdHandler.Exists("mangle", "POSTROUTING", mtuArgs...)
-	if err != nil {
-		return fmt.Errorf("failed to cleanup iptables command to set up TCPMSS due to %v", err)
-	}
-	if exists {
-		klog.V(2).Infof("removing mangle rule with: iptables -D POSTROUTING -t mangle %s", args)
-		err = iptablesCmdHandler.Delete("mangle", "POSTROUTING", mtuArgs...)
-		if err != nil {
-			return fmt.Errorf("failed to cleanup iptables command to set up TCPMSS due to %v", err)
+		if exists {
+			klog.V(2).Infof("removing mangle rule with: iptables -D PREROUTING -t mangle %s", args)
+			err = iptablesCmdHandler.Delete("mangle", "PREROUTING", mtuArgs...)
+			if err != nil {
+				return fmt.Errorf("failed to cleanup iptables command to set up TCPMSS due to %v", err)
+			}
 		}
 	}
 
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -1639,6 +1639,26 @@ func (nsc *NetworkServicesController) setupMangleTableRule(ip string, protocol s
 		}
 	}
 
+	// Previous versions of MTU args were this way, we will clean then up for the next couple of versions to ensure
+	// that old mangle table rules don't stick around
+	// TODO: remove after v2.4.X or above
+	for firstArg, chain := range map[string]string{"-s": "POSTROUTING", "-d": "PREROUTING"} {
+		prevMTUArgs := []string{firstArg, ip, "-m", tcpProtocol, "-p", tcpProtocol, "--tcp-flags", "SYN,RST", "SYN",
+			"-j", "TCPMSS", "--set-mss", strconv.Itoa(tcpMSS)}
+		klog.V(2).Infof("looking for mangle rule with: %s -t mangle %s", chain, prevMTUArgs)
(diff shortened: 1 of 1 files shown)


pkg/controllers/proxy/network_services_controller.go
```diff
@@ -1629,17 +1629,20 @@ func (nsc *NetworkServicesController) setupMangleTableRule(ip string, protocol s
...
-	if err != nil {
-		return fmt.Errorf("failed to run iptables command to set up TCPMSS due to %v", err)
+	if protocol == tcpProtocol {
+		mtuArgs := []string{"-d", ip, "-m", tcpProtocol, "-p", tcpProtocol, "--dport", port,
+			"--tcp-flags", "SYN,RST", "SYN", "-j", "TCPMSS", "--set-mss", strconv.Itoa(tcpMSS)}
+		err = iptablesCmdHandler.AppendUnique("mangle", "PREROUTING", mtuArgs...)
+		if err != nil {
+			return fmt.Errorf("failed to run iptables command to set up TCPMSS due to %v", err)
+		}
+		mtuArgs[0] = "-s"
```
aauren, line 1639: Wouldn't you need to change `--dport` to `--sport` here as well as changing the `-d` to `-s`?

aauren:
Thanks for raising a PR. This indeed looks pretty minimal, and I don't think it hurts to make the TCPMSS configurations port specific. I just left one clarifying question as a review.

pkg/controllers/proxy/network_services_controller.go
rkojedzinszky: You are right, I've missed that. Howewer, according to [TCP](https://datatracker.ietf.org/doc/html/rfc9293#name-maximum-segment-size-option) it is enough to reduce MSS in the first SYN packet. On the other hand, I miss a similar rule in OUTPUT chain in table mangle, which I will add that soon after some testing.

rkojedzinszky: Unfortunately, reducing MSS in intiail SYN packets is not enough, at least linux implementation responds with a higher MSS value. According to kube-router's design, this way the sender side dont get any clue of the reduced MSS. In theory, it is an assymetric routing, where only the connecting path is affected, thus it would be enough to just reduce MSS on SYN+ACK reply packages. I need more research and experiment on this.

rkojedzinszky:
So it turns out that we dont need POSTROUTING rules with destination address specified here, only rules with source address specififaction, as that is the SYN+ACK packet, that will reach the connection initiator. That packet has to have a reduced MSS value. Also, as that packet comes from a PODs own ns, it is placed in PREROUTING instead of POSTROUTING. A more nice solution would be to have this setup in the PODs own network namespace rather than in each node's hostnetwork.

rkojedzinszky:
@aauren did you have time to check/verify the PR?

aauren:
@rkojedzinszky - I had enough time today to look into this.

I think that it looks good! I agree with your conclusion that it is only necessary to set this via source and PREROUTING instead of POSTROUTING. Additionally, I got a chance to test this on a cluster of my own, and it looks correct from what I can see in tcpdump.

I did make one small change to your branch where I pushed a commit that adds some cleanup code into the logic that will hopefully remove lingering MSS changes from previous versions of kube-router (e.g. before this change). It will unfortunately, only execute when the DSR service is updated or deleted / re-applied, but getting it to run as a one-off is too much logic I think for now. Re-creating or forcing an update to DSR services is at least better than having to reboot the node to remove old iptables rules IMO.

Can you test one more time in your cluster to make sure that it is working with the extra iptables rules I added? If so, let me know and we'll merge it.

rkojedzinszky:
@aauren shouldn't that code be moved out of the `if protocol == tcpProtocol` block?

rkojedzinszky:
@aauren I am running kube-router without your patch, but, I've rebooted all my nodes, so I have no rules left there from previous kube-router. Howewer, I like the idea to support the upgrade without needing a reboot. But, as in my previous comment, I suspect that cleanup shall be placed outside of the `if` block, where it originally was.

aauren:
> @aauren shouldn't that code be moved out of the if protocol == tcpProtocol block?

That's a good catch. I moved it out of the `protocol == tcpProtocol` block

aauren:
@rkojedzinszky please take one more look to make sure that I didn't miss something and if you're good we'll merge.

rkojedzinszky:
@aauren I think we also could do the cleanup of old rules during setting up DSR. Then there will be no lingering rules. What do you think?

aauren:
This seems fine to me.

Thanks for all of your work on this @rkojedzinszky!

rkojedzinszky:
@aauren sorry for that, next time I will do that way. Thanks for merging.

aauren:
No worries. I appreciate the help and review. :smile:

PR #1674, opened 2024-05-17
docs: improve scheduler docs

bhcleek:
Fix a broken link and edit for active voice.

Document how to configure a service to use the maglev scheduler.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -351,10 +351,11 @@ information see the section above.
 
 ## Load balancing Scheduling Algorithms
 
-Kube-router uses LVS for service proxy. LVS support rich set of
-[scheduling alogirthms](http://kb.linuxvirtualserver.org/wiki/IPVS#Job_Scheduling_Algorithms). You can annotate
-the service to choose one of the scheduling alogirthms. When a service is not annotated `round-robin` scheduler is
-selected by default
+Kube-router uses LVS for service proxy. LVS supports a rich set of [scheduling
+algorithms](https://en.wikipedia.org/wiki/Linux_Virtual_Server#Schedulers). The
+scheduling algorithm for a service is configured by means of annotations. The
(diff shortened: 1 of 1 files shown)


docs/user-guide.md
```diff
@@ -368,6 +369,17 @@ $ kubectl annotate service my-service "kube-router.io/service.scheduler=sh"
 
 #For destination hashing scheduling use:
 $ kubectl annotate service my-service "kube-router.io/service.scheduler=dh"
+
+#For maglev scheduling use:
+$ kubectl annotate service my-service "kube-router.io/service.scheduler=mh"
+
+# The maglev scheduler can be further tuned with additional options.
+#To use the maglev scheduler's fallback option use:
+$ kubectl annotate service my-service "kube-router.io/service.schedFlags=flag-1"
```
aauren, line 378: I'm not sure if you tested this and found it working or not in your cluster, but I noticed that in the code, we have `schedflags` instead of `schedFlags`. For what it's worth, I like your version better, but we'll probably keep it all lower case for consistency with existing annotations.

* https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/proxy/network_services_controller.go#L75
* https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/proxy/network_services_controller.go#L970

While I haven't specifically tested that annotations are case sensitive, I believe that they are. Either way, its probably best to have it all lower case to match the code.

aauren (changes requested):
@bhcleek Thanks for fixing up the documentation. I added one thing to change, but other than that it looks good and I'm sure that it will help future users out.

docs/user-guide.md
bhcleek: You're right. Thankfully, in practice I've used the correct  casing and only misspelled it in this PR.

Issue #1676, opened 2024-05-21
kube-router duplicates rules in the KUBE-ROUTER-INPUT chain


TPXP:
**What happened?**
Hello, we're happily using kube-router to handle network policies on our kubernetes cluster (--run-router=false --run-firewall=true --run-service-proxy=false). We recently upgraded to `v2.0.1` and `v2.1.2` and it seems that kube-router is slow to sync firewall rules on some of our nodes.

We believe this could cause some packet drops impacting performance of our applications. Our investigations led us to look at iptables on our server. These look mostly OK, except for the KUBE-ROUTER-INPUT chain in the ip filter table. Indeed, on some of our machines, it that some rules are repeated:
```
-A KUBE-ROUTER-INPUT -d 10.96.0.0/12 -m comment --comment "allow traffic to primary/secondary cluster IP range - BVHX2PIHDXXEO43X" -j RETURN
-A KUBE-ROUTER-INPUT -p tcp -m comment --comment "allow LOCAL TCP traffic to node ports - LR7XO7NXDBGQJD2M" -m addrtype --dst-type LOCAL -m multiport --d
ports 30000:32767 -j RETURN
-A KUBE-ROUTER-INPUT -p udp -m comment --comment "allow LOCAL UDP traffic to node ports - 76UCBPIZNGJNWNUZ" -m addrtype --dst-type LOCAL -m multiport --d
ports 30000:32767 -j RETURN
-A KUBE-ROUTER-INPUT -p tcp -m comment --comment "allow LOCAL TCP traffic to node ports - LR7XO7NXDBGQJD2M" -m addrtype --dst-type LOCAL -m multiport --d
ports 30000:32767 -j RETURN
-A KUBE-ROUTER-INPUT -p udp -m comment --comment "allow LOCAL UDP traffic to node ports - 76UCBPIZNGJNWNUZ" -m addrtype --dst-type LOCAL -m multiport --d
ports 30000:32767 -j RETURN
[... 2 last lines are repeated thousands of times ...]
```

Our servers run Ubuntu 22.04 LTS so we're using nftables under the hood. Asking `nftables` what it has gives curious results:
```
# nft list chain ip filter KUBE-ROUTER-INPUT
table ip filter {
        chain KUBE-ROUTER-INPUT {
                ip daddr 10.96.0.0/12  counter packets 0 bytes 0 return
[...]

PR #1678, opened 2024-05-27
feat(alpine): revert 3.19 -> 3.18

aauren:
@mrueg 

Revert because of issue with iptables v1.8.10 no longer working with -C check.

Still testing this in a local cluster to make sure that the problem doesn't re-occur. Long term, this needs to be chased down with upstream netfilter, but for now this should at least help us get to a stable version of kube-router on the v2.x release line.

Fixes: #1676

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,11 +15,10 @@ on:
     - prep-v[1-9].*
 
 env:
-  # Upgrading to Go 1.21.{0,1,2} seems to in cause some sort of race condition that causes tests to fail sporadically.
-  # Sometimes they fail in apparent race conditions, other times they end up with nil pointer and SIGSEGV errors. We'll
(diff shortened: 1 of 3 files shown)


mrueg (approved):

TPXP:
Thanks a lot for taking the time to investigate and finding the root cause 🙌 

Some nodes of our clusters run Linux 6.2, and we saw some issues with iptables failing on these kernels.
> E0522 07:41:48.172118       1 network_policy_controller.go:259] Aborting sync. Failed to sync network policy chains: %!v(MISSING)ipset v7.15: Kernel and userspace incompatible: settype hash:ip with revision 6 not supported by userspace.

I haven't thoroughly experimented with all versions so I can't say for sure which version of ipset is required - I think kube-router v2.x builds worked but, just to be sure, you may want to test your build against these kernels as well

PR #1684, opened 2024-06-07
fix: ensure that ipv6 is not disabled in kernel

ncopa:


aauren:
Thanks for the contribution! Sorry that it took me a bet to get around to reviewing it.

In looking at it, I suspect that there would be other problems that would be caused by IPv6 being disabled when kube-router starts. For instance, I noticed that when I tried setting `ipv6_disabled` on a working IPv6 VM on a VM of Ubuntu 24.04 it removed the IPv6 address assignment that I had made.

Now, that was static, maybe if I used dynamic IPv6 address assignment on my VM it would have re-assigned it during kube-router's after this enabled ipv6?

Either, way I don't see any harm in this setting, so we'll go ahead and merge it, but other than getting past the error that your experiencing, it might not be all that helpful to other users depending on their situation.

PR #1697, opened 2024-07-05
feat: convert execs to ip to netlink calls

aauren:
Not making direct exec calls to user binary interfaces has long been a principle of kube-router. When kube-router was first coded, the netlink library was missing significant features that forced us to exec out. However, now netlink seems to have most of the functionality that we need.

This converts all of the places where we can use netlink to use the netlink functionality.

The current state of this PR is untested and still needs to undergo significant testing:

* [x] Ensure IPv4 routes are getting populated correctly
* [x] Ensure IPv4 source routing is being added to custom table
* [x] Ensure IPv6 routes are getting populated correctly
* [x] Ensure IPv6 source routing is being added to custom table
* [x] Ensure IPv4 Service VIPs get added to the dummy interface
* [x] Ensure IPv6 Service VIPs get added to the dummy interface
* [x] Ensure DSR works
* [x] Ensure ipip encapsulation works
* [x] Ensure fou encapsulation works

### DIFF
--- a/pkg/controllers/proxy/linux_networking.go
+++ b/pkg/controllers/proxy/linux_networking.go
@@ -62,6 +62,7 @@ type netlinkCalls interface {
 }
 
 func (ln *linuxNetworking) ipAddrDel(iface netlink.Link, ip string, nodeIP string) error {
+	var nFamily int
 	parsedIP := net.ParseIP(ip)
 	parsedNodeIP := net.ParseIP(nodeIP)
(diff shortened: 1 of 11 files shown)


aauren:
There is a minor outstanding case where there is some problems with DSR traffic (see #1870) however, this was true whether netlink was used or iproute2 user-space tooling. So I think that we're fine leaving that issue alone for now.

Issue #676, opened 2019-02-21
Newly joined nodes don't receive all routes when using RR mode


Feder1co5oave:
On an already existing kubernetes cluster with kube-router successfully running in route reflector mode (`rr.server` annotation on 3 master nodes, `rr.client` annotation on all the workers), my workflow for joining new worker nodes is like this (the cluster was created with kubeadm):

for each new node:
- run kubeadm join on the node
- kubectl annotate node <new-node> kube-router.io/rr.client=23
- restart kube-router on the new node to put them in RR mode

Turns out the nodes previously joined to the cluster will receive routes for all the new nodes, but the new nodes will only receive routes for the "old" nodes. Restarting kube-router on the route reflector nodes solved this issue.

My troubleshooting suggested that whenever a new node joins, the rr server kube-router daemons will peer with that node right away, even before the `rr.client` annotation gets added to it. Thus, the new peer is assumed to be in full mesh mode even if in fact it will be in RR mode and will only peer with RR servers. BGP route reflector allows for BGP daemons in the same AS to peer in either full-mesh or RR mode. Export policies are such that RR servers will reflect advertisements:
- from RR peers to RR peers
- from full-mesh peers to RR peers
- from RR peers to full-mesh peers
- but NOT from full-mesh peers to other full-mesh peers (they're supposed to peer with each other in full-mesh, duh)

So the new nodes, assumed to be forming full-mesh, don't get advertisements about other new nodes.
Restarting RR servers forces them to reload the node list and annotations, and correctly peer with the new nodes in RR mode.

To fix this, kube-router should watch for annotation changes on nodes, and update its internal information about which nodes are forming full-mesh, and which are joined to a RR cluster.
While waiting for a fix to be implemented, I suggest the workaround to restart RR servers be documented for newcomers!

aauren:
Closed via #677

ofen:
Is it possible to make it automatically (without manual POD restart)?

murali-reddy:
@ofen reopening as still below make sense to be fixed.

> kube-router should watch for annotation changes on nodes, and update its internal information about which nodes are forming full-mesh, and which are joined to a RR cluster.
While waiting for a fix to be implemented, I suggest the workaround to restart RR servers be documented for newcomers!

zhaixigui:
Why can't it automatically recognize the newly added node? And to propagate routing。Why have to do such a dangerous operation to kill RR?

zhaixigui:
[...]

PR #1723, opened 2024-08-22
fix: update peers on node update

sebltm:
Fixes https://github.com/cloudnativelabs/kube-router/issues/676 by reloading the peer configuration when nodes are updated

### DIFF
--- a/pkg/controllers/routing/bgp_peers.go
+++ b/pkg/controllers/routing/bgp_peers.go
@@ -207,7 +207,8 @@ func (nrc *NetworkRoutingController) syncInternalPeers() {
 // connectToExternalBGPPeers adds all the configured eBGP peers (global or node specific) as neighbours
 func (nrc *NetworkRoutingController) connectToExternalBGPPeers(server *gobgp.BgpServer, peerNeighbors []*gobgpapi.Peer,
 	bgpGracefulRestart bool, bgpGracefulRestartDeferralTime time.Duration, bgpGracefulRestartTime time.Duration,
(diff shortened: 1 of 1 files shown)


aauren:
Thanks @sebltm!

I won’t be able to test this out in my setup for a week or so do to IRL complications, but I’ll give it a go as soon as I’m able to.

FWIW on the surface this seems super simple and straight forward, if it’s really this simple I’ll be face-palming that we didn’t do this earlier 🙂

aauren:
After reviewing your change and putting it through its paces a bit, I think that this is definitely a good start, but that there is more to the story here.

From what I can tell, putting in the `nrc.OnNodeUpdate(newObj)` will update remote nodes, but it would not work cohesively as that won't cause the node to reload its own annotations and settings. There are a bunch of settings for the node itself that are only considered either when the Routes controller is being created (https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/routing/network_routes_controller.go#L1424-L1682) or when it is being started (https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/routing/network_routes_controller.go#L1074-L1371)

Even in the functions that `nrc.OnNodeUpdate()` calls, there are a few bits of logic that specifically exclude the node that it is running on from being considered: https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/routing/bgp_peers.go#L56

I think that we're going to need a more holistic overhaul of the controller before #676 is resolved.

PR #1724, opened 2024-08-26
feat: add rhel9 spec file along with its own build script

damex:
here we add:
  - rpm spec file (rhel9 and derivatives)
  - build/lint script to actually build rpm package
  - docs

### DIFF
--- a/build/rhel9-rpm-package/BUILD/.gitignore
+++ b/build/rhel9-rpm-package/BUILD/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore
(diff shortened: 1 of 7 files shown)


mrueg:
If you want rpms, can we use goreleaser to create them? https://goreleaser.com/customization/nfpm/

damex:
> If you want rpms, can we use goreleaser to create them? https://goreleaser.com/customization/nfpm/

it is using fpm.
fpm is much worse solution for building rpms compared to building spec files.
spec file is a self contained thing to build software with that will be packaged to an rpm while fpm just takes a config or binary or whatever and packs them in to rpm. quite different isn't it?

mrueg:
It's using https://github.com/goreleaser/nfpm and not fpm :)

The advantage is that it'll be embedded in the release process and we don't need to have another process that might fail. At least @aauren and I won't be able to support rpm builds.

PR #1737, opened 2024-09-11
Enhance DSR docs aroung CRI socket mounting

### LINKED ISSUE #1719: documentation/DSR examples: mount directory containing the socket instead the socket directly
## Is your feature request related to a problem? Please describe

I've setup DSR following the guide, mounting crio.sock directly. Howewer, with this setup, if crio is restarted in the background, the socket will get stale, and kube-router won't be able to communicate with the runtime anymore, kube-router must also  be restarted.

## Describe the solution you'd like

Simply mount the directory holding the control socket.

jnummelin:
Change the mount examples to use whole directory as the socket file might change during runtime and thus mounting the single file might break the behaviour in case of container runtime restarts.

Also simplified all the examples to use containerd as CRI as it's probably the most used CRI.

Fixes #1719


aauren:
@jnummelin Thanks for fixing this up!

PR #1747, opened 2024-10-08
Remove some unused error fields from test cases

twz123:
Some small nits I stumbled upon while spelunking through the test cases.


aauren:
Thanks @twz123! Appreciate the contribution!

Issue #1738, opened 2024-09-12
kube-router Holding on to Routes


aauren:
## What happened?

Over time, it appears that kube-router hits conditions where it will keep BGP routes that have since been withdrawn. This likely happens because [route_sync.go](https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/routing/route_sync.go#L13) contains its own cache of routes and at some point it isn't able to receive a BGP update for one reason or another.

Because of this problem, kube-router continues to put back bad routes to nexthops that no longer contain the service which essentially blackholes the traffic bound for that service.

## What did you expect to happen?

kube-router to have an accurate route state at all times.

## How can we reproduce the behavior you experienced?

This behavior is not easily reproduced and the exact cause of the issue is not yet known. It is something that involves state over time.

## System Information (please complete the following information)

- Kube-Router Version (`kube-router --version`): 2.1.3
- Kube-Router Parameters: 

```
--advertise-external-ip=true --bgp-graceful-restart=true --bgp-graceful-restart-deferral-time=60s --enable-ibgp=false --enable-overlay=false --hairpin-mode=true --kubeconfig=/etc/kubernetes/kubectl-config.yaml --metrics-port=9081 --nodes-full-mesh=false --run-router=true --run-firewall=true --service-cluster-ip-range=172.28.0.0/16 --service-external-ip-range=192.168.1.0/24 --service-external-ip-range=192.168.2.0/24 --peer-router-ips=192.168.3.1,192.168.3.2,192.168.3.3 --peer-router-asns=4220000001,4220000001,4220000001 --peer-router-passwords-file=/etc/kube-router-bgp.conf --cluster-asn=4220000001
```

- Kubernetes Version (`kubectl version`) : 1.28.10
- Cloud Type: On Prem
- Kubernetes Deployment Type: Custom
- Kube-Router Deployment Type: System Service
- Cluster Size: ~100 nodes

## Logs, other output, metrics

No logs show up with this issue

## Additional context

kube-router probably needs to add a consistency check that happens periodically when the [routes_sync controller is running](https://github.com/cloudnativelabs/kube-router/blob/master/pkg/controllers/routing/route_sync.go#L55).

This would allow the controller to be primarily event driven, but also retrue it's state from time to time to ensure that it doesn't get into an inconsistent state with the desired state of the BGP subsystem.

aauren:
This was accidentally closed on the merge of #1739 and is not yet fully completed.

[...]

PR #1749, opened 2024-10-09
Fix: Add Route Reconciliation

aauren:
This PR is still very early and is not fully implemented, but I wanted to start to get some feedback going on the direction of this before I got too much further down the road.

The end goal of this PR is to enable route reconciliation within the route sync controller. However, in order to do that, I wanted to break out `injectRoute()` from the NetworkRoutesController.

Also, over the years, it has become obvious that the `injectRoute()` function is hard for new users to parse, and there is often a lot of confusion about how it works because the logic is complicated. So this also attempts to split out the logic in inject route into BGP type logic as well as Route type logic.

Final fix for: https://github.com/cloudnativelabs/kube-router/issues/1738


pkg/types.go
```diff
@@ -0,0 +1,45 @@
+package pkg
```
aauren, line 1: @twz123 I wanted to get your feedback about this convention within Go.

As I started to get into the refactoring of InjectRoute, I found that I kept getting into circular dependency type situations between the `bgp` and `routes` packages.

After watching a [video from Brian Ketelsen](https://www.youtube.com/watch?v=MzTcsI6tn-0) about structuring go programs, as well as several other posts, it seems like pulling up your domain models into the root of the project is a conventional way of working around circular dependencies.

Doing this with the below methods, did work out for me, but I noticed that this architecture is limited. For instance, I couldn't leave the `TunnelCreator` interface with `EncapType` and `EncapPort` because those are declared in the `tunnels` package. And for some reason, it didn't feel right to bring those up into this top level package either.

So I worked around it for now, but maybe this is showing that either I should have brought them up, or that maybe I'm embracing the wrong pattern here.

I would be curious for your thoughts.

pkg/routes/route_sync.go
```diff
@@ -1,82 +0,0 @@
-package routes
```
aauren, line 1: @twz123 I ended up bringing this back into the `routing` package because I realized that it was behaving a bit more like an extension of the NetworkRoutesController, in that, in order to have it process route reconciliation I needed it to work with the routes source of truth, which is BGP. At that point it is no longer strictly related to host routes (although now that I'm thinking about it, it means that maybe `host_route_sync.go` isn't the right name for the file any more either.

Anyway, curious about your thoughts here as well.

pkg/routes/linux_routes.go
```diff
@@ -30,3 +32,109 @@ func DeleteByDestination(destinationSubnet *net.IPNet) error {
 	}
 	return nil
 }
+
+func InjectRoute(subnet *net.IPNet, gw net.IP, ipa utils.NodeIPAware, tn pkg.Tunneler, rs pkg.RouteSyncer,
```
aauren, line 36: @twz123 - Now that this logic is disentangled with the BGP logic that used to be buried in here I like it a lot better. It feels a lot cleaner and more straight forward to me.

pkg/bgp/path.go
```diff
@@ -0,0 +1,45 @@
+package bgp
+
+import (
+	"github.com/cloudnativelabs/kube-router/v2/pkg"
+	gobgpapi "github.com/osrg/gobgp/v3/api"
+	"k8s.io/klog/v2"
+)
+
+func PathChanged(path *gobgpapi.Path, pl PeerLister, rs pkg.RouteSyncer, tc pkg.TunnelCleaner) error {
```
aauren, line 9: @twz123 - This feels like the right pattern, because it decouples the `routes` package from the `bgp` package by making it implementation agnostic. So I think that making all of the arguments to this function interfaces is the right thing to do, but wanted to see if you had any feedback.

pkg/routes/linux_routes.go
```diff
@@ -30,3 +32,115 @@ func DeleteByDestination(destinationSubnet *net.IPNet) error {
 	}
 	return nil
 }
+
+type LinuxRouter struct {
+	NodeIPA       pkg.NodeIPAware
+	Tunneler      pkg.Tunneler
+	RouteSyncer   pkg.RouteSyncer
```
aauren, line 39: I don't particularly like the circular dependency here. `LinuxRouter` requires `pkg.RouteSyncer` so that it can update the state host routes. `pkg.RouteSyncer` in turn requires `LinuxRouter` so that it can reuse the `InjectRoute()` logic when new routes are found during `checkCacheAgainstBGP()`.

It feels like there should be a better / cleaner way to do this.

aauren:
Most of this PR is out-dated. After several attempts to patch it up, I've found that my approach of adding BGP Cache Checking into the route_sync controller directly, just isn't the best way to do what I'm trying to do here. Over time I've seen that it is frought with problems and logic gaps.

I'm leaving this PR around for now, while I re-consider my approach here, but in the end, I think that most of the problems can be resolved by adopting commit `4c9fb28`.

aauren:
After further consideration I've decided not to take this PR further.

It feels like the reconcile loop kept running afoul of circular dependencies and it started stretching some of the abstractions to the breaking point.

Ultimately, the problem that it was trying to solve, was fixed via `4c9fb28`. It may still be worthwhile to have a reconcile loop on host routes at some point in the future, but for now kube-router is working well enough without this janky logic.

Issue #1754, opened 2024-10-20
DSR not working with Rocky / RHEL:  Unavailable desc = name resolver error: produced zero addresses


opipenbe:
## What happened?
DSR mode is not working with Rocky / RHEL 9 if using kube-router v2.2.0 and v2.2.1. 

In the kube-router logs:
`E1020 08:05:55.567390  372833 service_endpoints_sync.go:60] Error setting up IPVS services for service external IP's and load balancer IP's: failed to setup DSR endpoint 30.0.0.1: unable to setup DSR receiver inside pod: failed to prepare endpoint 192.168.6.78 to do DSR due to: rpc error: code = Unavailable desc = name resolver error: produced zero addresses
`

DSR is working successfully with kube-router v2.1.3 with Rocky Linux 9.4. Error above occurs with kube-router v2.2.0 and v2.2.1. I believe a change between kube-router v2.1.3 and v2.2.0 created this incompatibility for DSR. I also tested DSR with Ubuntu 24.04 & kube-router v2.2.x in the same cluster and it does not have such issue.

## What did you expect to happen?
DSR mode enabled without errors for RHEL and its clones.

## How can we reproduce the behavior you experienced?

1. Install kubeadm k8s cluster without kube-proxy and with cri-o runtime.
2. Deploy latest kube-router v2.2.1 using kubeadm-kuberouter-all-features-dsr.yaml (https://github.com/cloudnativelabs/kube-router/blob/master/daemonset/kubeadm-kuberouter-all-features-dsr.yaml).
3. Make sure to make following changes in kubeadm-kuberouter-all-features-dsr.yaml:  
- set `--runtime-endpoint=unix:///run/crio/crio.sock` 
-  replace `/var/run/docker.sock` with `/run/crio/crio.sock` in volumeMounts and volumes configuration
- instead of:
```
      - name: kubeconfig
        configMap:
          name: kube-proxy
          items:
          - key: kubeconfig.conf
            path: kubeconfig
```
replace with:
```
      - name: kubeconfig
        hostPath:
          path: /var/lib/kube-router
```

## System Information (please complete the following information)

- Kube-Router Version (`kube-router --version`): v2.2.0 and v2.2.1
- Kube-Router Parameters: - --run-router=true - --run-firewall=true - --run-service-proxy=true- --bgp-graceful-restart=true - --kubeconfig=/var/lib/kube-router/kubeconfig - --runtime-endpoint=unix:///run/crio/crio.sock
[...]

PR #1756, opened 2024-10-20
fix(dsr): change grpc resolver to passthrough

aauren:
@mrueg 

When we used DialContext() before it was deprecated, it manually set the default resolver to passthrough. NewClient() appears to not do this, and then several versions of grpc ago, they defaulted to the DNS resolver which broke our DSR implementation with CRI compatible container engines.

I'm not 100% sure that there isn't a better way to specify this, but I spent 20 - 30 minutes poking around the gRPC code base and I don't think that I can see an easy way to do this outside of specifying it this way.

Furthermore, the project itself seems to advocate for this approach in comments like those on: https://github.com/grpc/grpc-go/issues/1846

Fixes: #1754

### DIFF
--- a/pkg/cri/remote_runtime.go
+++ b/pkg/cri/remote_runtime.go
@@ -41,7 +41,11 @@ func NewRemoteRuntimeService(endpoint string, connectionTimeout time.Duration) (
 
 	klog.V(4).Infof("[RuntimeService] got endpoint %s (proto=%s, path=%s)", endpoint, proto, addr)
 
-	if proto != "unix" {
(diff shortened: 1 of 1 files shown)


mrueg (approved):

aauren:
Yes, sorry, I should have prepped you for that. If you used the build from this pipeline instead of just applying the patch, then you would have hit the regression that I fixed in another PR: https://github.com/cloudnativelabs/kube-router/pull/1755

Since both have been reviewed, I plan on merging both today, and then cutting a new release.

PR #1771, opened 2024-11-26
fix: avoid to use http.DefaultServeMux which includes defaults routes inited by go like /debug/pprof

morlay:
https://cs.opensource.google/go/go/+/refs/tags/go1.23.3:src/net/http/pprof/pprof.go;l=100

metrics may just metrics

if enable pprof in metrics server, the flags `--enable-pprof` may be confused.

### DIFF
--- a/pkg/metrics/metrics_controller.go
+++ b/pkg/metrics/metrics_controller.go
@@ -260,13 +260,15 @@ func (mc *Controller) Run(healthChan chan<- *healthcheck.ControllerHeartbeat, st
 	DefaultRegisterer.MustRegister(BuildInfo)
 	DefaultRegisterer.MustRegister(ControllerIpvsMetricsExportTime)
 
+	mux := http.NewServeMux()
+
 	srv := &http.Server{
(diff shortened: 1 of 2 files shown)


pkg/metrics/metrics_controller.go
```diff
@@ -260,13 +260,15 @@ func (mc *Controller) Run(healthChan chan<- *healthcheck.ControllerHeartbeat, st
 	DefaultRegisterer.MustRegister(BuildInfo)
 	DefaultRegisterer.MustRegister(ControllerIpvsMetricsExportTime)
 
+	mux := &http.ServeMux{}
```
aauren, line 263: Small nit here, can you change this to `http.NewServeMux()` instead? It amounts to the same, but seems better to use the default function provided.

aauren (changes requested):
Can you also do the same for the health controller as well? https://github.com/cloudnativelabs/kube-router/blob/master/pkg/healthcheck/health_controller.go#L223-L226

morlay:
updated.

aauren (approved):
LGTM

Thanks for the contribution! :slightly_smiling_face:

PR #1791, opened 2024-12-30
fix(NSC): only set rp_filter to 2 if it is 1

dsseng:
Setting rp_filter to 2 when it is 0 will override its status to be always enabled (in the loose mode).
This behavior could break some networking solutions as it made packet admission rules more strict.

Reference: https://github.com/cloudnativelabs/kube-router/pull/1651#issuecomment-2561151350 and subsequent

### DIFF
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -225,6 +225,15 @@ type endpointSliceInfo struct {
 // map of all endpoints, with unique service id(namespace name, service name, port) as key
 type endpointSliceInfoMap map[string][]endpointSliceInfo
 
+func checkRPFilter1(ifname string) bool {
+	rpFilterValue, err := utils.GetSysctlSingleTemplate(utils.IPv4ConfRPFilterTemplate, ifname)
+	if err != nil {
+		klog.Errorf("failed to get rp_filter value for %s: %s", ifname, err.Error())
+		return false
+	}
+	return strings.TrimSpace(rpFilterValue) == "1"
+}
+
 // Run periodically sync ipvs configuration to reflect desired state of services and endpoints
 func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.ControllerHeartbeat,
 	stopCh <-chan struct{}, wg *sync.WaitGroup) {
@@ -286,16 +295,22 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
 	// https://github.com/kubernetes/kubernetes/pull/70530/files
 	setSysCtlAndCheckError(utils.IPv4ConfAllArpAnnounce, arpAnnounceUseBestLocalAddress)
 
-	// Ensure rp_filter=2 for DSR capability, see:
+	// Ensure rp_filter=2 (or leave 0 untouched) for DSR capability, see:
 	// * https://access.redhat.com/solutions/53031
 	// * https://github.com/cloudnativelabs/kube-router/pull/1651#issuecomment-2072851683
+	// Only override rp_filter if it is set to 1, as enabling it from 0 to 2 can cause issues
+	// with some network configurations which use reverse routing
 	if nsc.krNode.IsIPv4Capable() {
-		sysctlErr := utils.SetSysctlSingleTemplate(utils.IPv4ConfRPFilterTemplate, "all", 2)
-		if sysctlErr != nil {
-			if sysctlErr.IsFatal() {
-				klog.Fatal(sysctlErr.Error())
-			} else {
-				klog.Error(sysctlErr.Error())
+		for _, ifname := range []string{"all", "kube-bridge", "kube-dummy-if", nsc.krNode.GetNodeInterfaceName()} {
+			if checkRPFilter1(ifname) {
+				sysctlErr := utils.SetSysctlSingleTemplate(utils.IPv4ConfRPFilterTemplate, ifname, 2)
+				if sysctlErr != nil {
+					if sysctlErr.IsFatal() {
+						klog.Fatal(sysctlErr.Error())
+					} else {
+						klog.Error(sysctlErr.Error())
+					}
+				}
 			}
 		}
 	}
--- a/pkg/utils/sysctl.go
+++ b/pkg/utils/sysctl.go
@@ -29,13 +29,18 @@ type SysctlError struct {
 	additionalInfo string
 	err            error
 	option         string
+	hasValue       bool
 	value          int
 	fatal          bool
 }
 
 // Error return the error as string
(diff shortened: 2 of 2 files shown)


dsseng:
Also this doesn't work in all cases as I see, since this won't disable rp_filter on these in case all is set to 1 or 2:
https://github.com/cloudnativelabs/kube-router/blob/93498fb1acddbecc06fd404e5222071530e69bc3/pkg/controllers/proxy/linux_networking.go#L769-L787

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -286,10 +286,24 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
 	// https://github.com/kubernetes/kubernetes/pull/70530/files
 	setSysCtlAndCheckError(utils.IPv4ConfAllArpAnnounce, arpAnnounceUseBestLocalAddress)
 
-	// Ensure rp_filter=2 for DSR capability, see:
+	// Only override rp_filter if it is set to 1, as enabling it from 0 to 2 can cause issues with some network configurations
+	rpFilter := false
+	for _, ifname := range []string{"all", "kube-bridge", nsc.krNode.GetNodeInterfaceName()} {
```
dsseng, line 291: Is this list sufficient and specific to what we should check? I am not competent in kube-router's internal operation, so I do not know on which interfaces rp_filter could disrupt kube-router's operation.

aauren (changes requested):
There's one linting error that needs correcting, and I left a request to add one more interface that I missed in our previous conversation, and I had one question.

I also deployed it to my test harness and while I haven't tested all of the boundary conditions for rp_filter values, I was able to confirm that it runs similar to how it ran before.

But other than that, this change looks really good! Thanks for putting this together for us!

aauren:
> Also this doesn't work in all cases as I see, since this won't disable rp_filter on these in case all is set to 1 or 2:
> 
> https://github.com/cloudnativelabs/kube-router/blob/93498fb1acddbecc06fd404e5222071530e69bc3/pkg/controllers/proxy/linux_networking.go#L769-L787

I'm not totally sure if I understand this comment correctly, doesn't the last case set rp_filter to `0` on the `all` interface and `0` is more permissive than `1` or `2`?

Also, just to be clear, this logic is setting that value for the interfaces found inside the container's network namespace in case that helps clarify anything for you.

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -286,10 +286,24 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
...
+			rpFilter = true
+			break
+		}
+	}
+
+	// Ensure rp_filter=2 (or leave 0 untouched) for DSR capability, see:
 	// * https://access.redhat.com/solutions/53031
 	// * https://github.com/cloudnativelabs/kube-router/pull/1651#issuecomment-2072851683
-	if nsc.krNode.IsIPv4Capable() {
+	if nsc.krNode.IsIPv4Capable() && rpFilter {
```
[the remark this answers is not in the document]
dsseng: I agree we should perhaps override only the necessary interfaces (`"kube-bridge", "kube-dummy-if", nsc.krNode.GetNodeInterfaceName()`)

dsseng:
> > Also this doesn't work in all cases as I see, since this won't disable rp_filter on these in case all is set to 1 or 2:
> > https://github.com/cloudnativelabs/kube-router/blob/93498fb1acddbecc06fd404e5222071530e69bc3/pkg/controllers/proxy/linux_networking.go#L769-L787
> 
> I'm not totally sure if I understand this comment correctly, doesn't the last case set rp_filter to `0` on the `all` interface and `0` is more permissive than `1` or `2`?
> 
> Also, just to be clear, this logic is setting that value for the interfaces found inside the container's network namespace in case that helps clarify anything for you.

Okay, if it's inside the container that's not to be considered. Thanks for clarifying

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -225,6 +225,15 @@ type endpointSliceInfo struct {
 // map of all endpoints, with unique service id(namespace name, service name, port) as key
 type endpointSliceInfoMap map[string][]endpointSliceInfo
 
+func checkRpFilter1(ifname string) bool {
```
aauren, line 228: nitpic, but this would probably be better named `checkRPFilter1` since `RP` is an acronym for reverse path.

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -286,16 +295,23 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
...
+	rpFilterAll := checkRpFilter1("all")
 	if nsc.krNode.IsIPv4Capable() {
-		sysctlErr := utils.SetSysctlSingleTemplate(utils.IPv4ConfRPFilterTemplate, "all", 2)
-		if sysctlErr != nil {
-			if sysctlErr.IsFatal() {
-				klog.Fatal(sysctlErr.Error())
-			} else {
-				klog.Error(sysctlErr.Error())
+		for _, ifname := range []string{"kube-bridge", "kube-dummy-if", nsc.krNode.GetNodeInterfaceName()} {
+			if rpFilterAll || checkRpFilter1(ifname) {
```
aauren, line 306: Just setting the `rp_filter` setting on the main interfaces is most likely enough, but I've had enough experience with this project where making smaller changes breaks some sort of edge case for someone else. So since I'm likely to miss some edge case for someone else in my testing, I'd feel better if we could just add the `all` interface here to this list instead of the check above.

My thought is that without setting `rp_filter` on the `all` interface, we'll miss the `veth` devices that are created by the CNI dynamically. We'll also miss other people's interfaces tunnelling interfaces for things like wireguard if they have those or some such.

Thoughts?

aauren (commented):
Thanks for the update! 2 more comments, but I think we're just about at the end here.

pkg/controllers/proxy/network_services_controller.go
dsseng: veth might be the case, yes. So perhaps if all is 1 we set it to 2, if 0 we keep it

aauren: Yeah, I agree.

aauren: Just wanted to check in, did you want to make this change? Or would you prefer for me to push to your branch with the change?

dsseng: Ah, will do in a moment. Got caught into other tasks briefly. Sorry for a delay

aauren (approved):
LGTM!

Thanks so much for your contribution @dsseng!

PR #1811, opened 2025-02-07
V2.0.1 pp

pavel-pulsepoint:


aauren:
Closing this, as I don't think that you meant to open this PR against the upstream repo.

aauren:
However, if you ever feel like any of the changes you carry locally would be applicable to other kube-router users and wanted to contribute them upstream, feel free to open appropriate PRs and we could look into them!

pavel-pulsepoint:
I inherited this code and have been trying to update it from upstream unknown 6-year-ago version to 2.0.7. There are two main features that are required in our case:
1. node weights using IPVS; and 
2. custom EgressIPs

aauren:
Thanks for the explanation of the important changes that are carried.

Are you able to describe a bit more of the purpose of either of them?

The custom egress IPs is probably the most easily explained. I'm guessing that you want to control exactly which IP k8s traffic egresses from to maybe color the traffic a certain way on your network? That seems like it could probably be useful to others.

I've played with adding something similar to [Cilium's egress gateways](https://docs.cilium.io/en/stable/network/egress-gateway/egress-gateway/) in the past. Its been something that's requested a few times. This would be a coarser version of that which changes it on a per-node basis instead of a per-pod basis I'm thinking which is a bit more straight forward, but less granular.

The node weights I'm not sure that I really understand at all. Maybe you have a very heterogeneous set of nodes in your cluster and you want to represent that at the service routing layer? If so, in which ways are your nodes different? CPU architecture and speed?

Issue #1816, opened 2025-02-27
support change the default port of 50051


Asutorufa:
## Is your feature request related to a problem? Please describe

I already have another service run at 50051 port, and I can't change it.

## Describe the solution you'd like

Support to change default port of 50051, eg: ENV/ARGS. 

## Describe alternatives you've considered

I want to change the port, but here seems no cli ARGS or ENV to configure it.

## Additional context

```
I0227 13:39:48.557244 4015617 network_routes_controller.go:310] Starting network route controller 
time="2025-02-27T13:39:48Z" level=warning msg="listen failed" Error="listen tcp 192.168.2.245:50051: bind: address already in use" Key="192.168.2.245:50051" Topic=grpc 
time="2025-02-27T13:39:48Z" level=fatal msg="failed to listen grpc port" Err="listen tcp 192.168.2.245:50051: bind: address already in use"
              
```

PR #1820, opened 2025-03-02
feat(options): add flag for GoBGP grpc port

aauren:
Introduces the option `--gobgp-admin-port` which allows for setting the port that will be used for exposing GoBGP's grpc port.

Fixes: #1816

### DIFF
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -167,7 +167,7 @@ func (s *KubeRouterConfig) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&s.GlobalHairpinMode, "hairpin-mode", false,
 		"Add iptables rules for every Service Endpoint to support hairpin traffic.")
 	fs.Uint16Var(&s.GoBGPAdminPort, "gobgp-admin-port", defaultGoBGPAdminPort,
-		"Port to connect to GoBGP for administrative purposes.")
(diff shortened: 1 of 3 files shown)


aauren:
@rbrtbnfgl / @mrueg / @twz123 - This one is hopefully a simple one to review

pkg/options/options.go
```diff
@@ -164,6 +166,8 @@ func (s *KubeRouterConfig) AddFlags(fs *pflag.FlagSet) {
 		"Excluded CIDRs are used to exclude IPVS rules from deletion.")
 	fs.BoolVar(&s.GlobalHairpinMode, "hairpin-mode", false,
 		"Add iptables rules for every Service Endpoint to support hairpin traffic.")
+	fs.Uint16Var(&s.GoBGPAdminPort, "gobgp-admin-port", defaultGoBGPAdminPort,
+		"Port to connect to GoBGP for administrative purposes.")
```
mrueg, line 170: Worth mentioning that 0 = disabling it?

mrueg (approved):

pkg/options/options.go
aauren: That didn't really work before, but I made it work now. There have been several requests in the past to not have the GoBGP gRPC port exposed, so hopefully this makes that more accessible for those use-cases.

Thanks for bringing this up!

PR #1824, opened 2025-03-12
Update aws-sdk to v2

mrueg:
sdk v1 will be EOL in July 2025 https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

This is not tested since I don't have an AWS account that supports an environment with kube-router right now, but a translation based on the migration steps in the v2 guide.


aauren (approved):
Sorry for the late review.

Thanks for putting this together! I got a chance to test it out and it all worked exactly as it should.

aauren:
There was something off about either this branch or my rebase and force push to it. Anyway, it was merged here: https://github.com/cloudnativelabs/kube-router/commit/ff8d42fd83b71be070cc101a25635ad3dc117538

Thanks again @mrueg !

PR #1859, opened 2025-06-03
Support --master without --kubeconfig

### LINKED ISSUE #1614: Fix CLI Options - --master doesn't work without --kubeconfig
**What happened?**

A user attempting to only override the kube-apiserver address with `--master` would expect that it would pick up the credentials mounted into the pod via `/run/secrets/`.

The APIs that kube-router uses from client-go forces kube-router to get the config from an actual kubeconfig file and if the user only supplies `--master` it looks like kube-router will always fail to contact the kube-apiserver.

In this case, the only way that the user can override the kube-apiserver details, is to overload the ENV vars that kubernetes supplies and set `KUBERNETES_SERVICE_HOST` & `KUBERNETES_SERVICE_PORT`

1. Launch kube-router with `--master` but no `--kubeconfig` file

qbnit0:
If --master is specified, but --kubeconfig is not, use in-cluster config and override the API server host.

This would fix #1614.

### DIFF
--- a/pkg/cmd/kube-router.go
+++ b/pkg/cmd/kube-router.go
@@ -41,12 +41,26 @@ func NewKubeRouterDefault(config *options.KubeRouterConfig) (*KubeRouter, error)
 	var err error
 	version.PrintVersion(true)
 	version.PrintVersionMessages(true)
-	// Use out of cluster config if the URL or kubeconfig have been specified. Otherwise use incluster config.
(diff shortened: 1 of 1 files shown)


qbnit0:
Just did a quick test, works as expected.

qbnit0:
All done!

aauren:
Thanks for your contribution! I appreciate you taking the time to fix an issue that's been in the backlog for a while! 🙂

PR #1874, opened 2025-07-02
Provide real-time service metrics

rkojedzinszky:
Service metrics are collected during scraping now, rather than prepopulating them. This allows more frequent collection, and not bound to service synchronization period anymore. For example, on a PI cluster, CPU cycles can be saved with higher synchronization intervals, howewer, metrics still can be collected real-time.


rkojedzinszky:
I was wondering if linux namespace switches will collide with statsitics collection in different threads. Howewer, I found that existing code already takes care of it [here](https://github.com/cloudnativelabs/kube-router/blob/bcaced787ba6de8cf106784a6ea4947617ae9bab/pkg/controllers/proxy/utils.go#L282) and [here](https://github.com/cloudnativelabs/kube-router/blob/bcaced787ba6de8cf106784a6ea4947617ae9bab/pkg/controllers/proxy/hairpin_controller.go#L70). Am I right?

aauren:
Thanks for submitting this @rkojedzinszky. I'm always excited to see people from the community that want to make kube-router better!

As a small note for next time, typically with larger change sets like this, it is best to open an issue ahead of time with the project just to ensure that we don't waste your development time on a feature that might not be accepted by the project.

The issue also helps because the template usually guides users through the process of describing the business value of the change a little better. In this case, what is the real-world problem that you're trying to solve?

Is it fair to say that you're seeing an adverse impact on minimal systems from IPVS metric collection when it is attached to the service update timer? Are you able to quantify this? What type of intervals do you use after this update is made? How has it impacted your cluster?

rkojedzinszky:
Closes https://github.com/cloudnativelabs/kube-router/issues/1875

rkojedzinszky:
@aauren I really appreciate your attention, I've created an issue for this. Sorry for me being hasty.

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -735,75 +744,154 @@ func (nsc *NetworkServicesController) publishMetrics(serviceInfoMap serviceInfoM
 		}
 	}()
 
-	ipvsSvcs, err := nsc.ln.ipvsGetServices()
+	ipvsHandle, err := ipvs.New("")
```
aauren, line 747: I'm not sure that I understand why you replaced `nsc.ln.ipvsGetServices()` with a manual call to `ipvsHandle.GetServices()`. Can you explain this a bit more?

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -725,7 +720,21 @@ func (nsc *NetworkServicesController) syncIpvsFirewall() error {
 	return nil
 }
 
-func (nsc *NetworkServicesController) publishMetrics(serviceInfoMap serviceInfoMap) error {
+func (*NetworkServicesController) Describe(ch chan<- *prometheus.Desc) {
```
aauren, line 723: Over time I've been trying to decrease the size of the controller files. Lets go ahead and move this out of this file and into a dedicated `metrics.go` file inside the `controllers/proxy` package to separate the concerns a bit and reduce the complexity of this file.

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -735,75 +744,154 @@ func (nsc *NetworkServicesController) publishMetrics(serviceInfoMap serviceInfoM
...
+
+	type svcMapKey struct {
+		ip       string
+		uPort    uint16
+		protocol uint16
+	}
 
-		protocol = convertSvcProtoToSysCallProto(svc.protocol)
-		for _, ipvsSvc := range ipvsSvcs {
+	serviceMap := map[svcMapKey]*serviceInfo{}
```
aauren, line 766: I like what you did here with the map instead of having the embeded for loops!

However, this `Collect()` function is still a bit large. Let's go ahead and separate this `serviceMap` building into its own function. We could potentially separate out the second `for` loop into its own function as well and then just call both from `Collect()`.

pkg/controllers/proxy/service_endpoints_sync.go
```diff
@@ -861,24 +858,3 @@ func (nsc *NetworkServicesController) cleanupDSRService(fwMark uint32) error {
 	delete(nsc.fwMarkMap, fwMark)
 	return nil
 }
-
-func (nsc *NetworkServicesController) cleanupStaleMetrics(activeServiceEndpointMap map[string][]string) {
```
aauren, line 865: Are you sure that we don't need any sort of cleanup function any longer with the changes you've made? I haven't gotten a chance to test it yet, but have you made sure that when a service is removed it no longer produces stale metrics?

I remember before this function was added there were quite a few issues filed complaining about how it was ruining people's prometheus cardinality by continuing to track stale metrics for services and label sets that were no longer in the cluster.

aauren (changes requested):
Thanks again for this PR @rkojedzinszky!

I really appreciate the thought and time that you put into this PR and I agree with the use-case. Just because people have long (or short) service sync times, doesn't mean that it should change the granularity of the metrics produced. I really like this approach to making the metrics produce just in time.

A lot of the review (other than requesting that metric logic be pulled out of the `network_services_controller.go` file, is mostly questions and genuine query for your thoughts. Let me know what you think!

pkg/controllers/proxy/service_endpoints_sync.go
rkojedzinszky: As metrics are collected real-time and created manually, they are not stored anywhere. With this refactor there will be no stale metrics.

pkg/controllers/proxy/network_services_controller.go
aauren: Thanks for the clarification.

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -115,7 +118,7 @@ type NetworkServicesController struct {
 	krNode              utils.NodeAware
 	syncPeriod          time.Duration
 	mu                  sync.Mutex
-	serviceMap          serviceInfoMap
+	serviceMap          unsafe.Pointer
```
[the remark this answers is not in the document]
aauren: I personally feel that using channels is just more idiomatic and understandable for future use-cases.

But its fine for now. If you don't mind, just add a comment above the unsafe pointer declarations with this info in it for now. We can see how it goes and refactor later if needed.

aauren:
> > @rkojedzinszky I think that we're almost there with this one. We just need to separate the logic out into a metrics.go file, break up the collect function a bit, and then add a comment above the unsafe pointer declaration.
> > Let me know if you run into any time constraints and I can make the changes as well. I think this will be a good one to get in for the rest of the community.
> 
> I am bit busy now, I'll be able to work on that in the next two weeks.

Sounds good!

rkojedzinszky:
> @rkojedzinszky I think that we're almost there with this one. We just need to separate the logic out into a metrics.go file, break up the collect function a bit, and then add a comment above the unsafe pointer declaration.

I've replaced the unsafe.Pointer to atomic.Pointer, that makes its meaning more obvious.

> 
> Let me know if you run into any time constraints and I can make the changes as well. I think this will be a good one to get in for the rest of the community.

aauren:
@rkojedzinszky - My apologies, I didn't get a notification about your rebase or your request for review after my last comment.

Did not intend to sit on this so long. I will take a look and give a final review this weekend.

rkojedzinszky:
> @rkojedzinszky - My apologies, I didn't get a notification about your rebase or your request for review after my last comment.
> 
> Did not intend to sit on this so long. I will take a look and give a final review this weekend.

Sorry, I've missed to post a notification. Thanks for your efforts.

aauren (approved):
Upon testing this out, I actually realized that until your patch we were never exposing externalIPs to the metric set. So thanks for adding that!

I also realized that we were not collecting LoadBalancer IPs either, so I pushed that change into your branch. I hope you don't mind, I just wanted to save us another back and forth since I've let this PR sit for a pretty long time already.

I tested as thoroughly as I could with a few services and then hundreds of services. With hundreds of services it actually goes quite a bit slower than I thought it would (~2 seconds or so). But this change actually results in slightly better response times than the previous (non-dynamic) change set. Not totally sure why, but I'll take it.

It looks like for large clusters we could probably come back to the metrics as a whole and see if there is a way to make them more performant across the board. But for now, this change set seems to be all around better than what we had before.

Thanks @rkojedzinszky !

PR #1894, opened 2025-08-07
Update k8s Deps

aauren:
@jnummelin @rbrtbnfgl @mrueg @twz123 

This is a PR to update Kubernetes API versions from 0.31.X to 0.33.X

A big part of this update was the official deprecation of the Endpoints API in favor of EndpointSlices. We've started to use EndpointSlices in several places in the project, but Endpoints was still pretty highly used. As such this was a somewhat complicated refactor. If you have a second, please give it a look over and give feedback if you see something that should be cleaned up.


aauren:
FYI @catherinetcai as well if you have a chance to test this change

aauren:
Haven't received much feedback on this one, but I ran it through multiple rounds of the Kubernetes network conformance tests as well as running it locally and haven't seen any regressions. So I'm going to proceed with it before it gets more merge conflicts.

Issue #1393, opened 2022-11-08
Restructure node peer annotations


rkojedzinszky:
**Is your feature request related to a problem? Please describe.**
For many external peers for a node, it might be difficult to change per-node settings (e.g. password, asn prepends, etc.) because they are stored in different arrays.

**Describe the solution you'd like**
Restructure configuration in a serialized json or yaml format, with elements of peer configurations.

Instead of:
```
apiVersion: v1
kind: Node
metadata:
  annotations:
...
    kube-router.io/peer.asns: "64640"
    kube-router.io/peer.ips: 192.168.8.254
    kube-router.io/peer.passwords: cGFzc3dvcmQ=
...
```

We could have:
```
apiVersion: v1
kind: Node
metadata:
  annotations:
...
    kube-router.io/peers: |
      - remoteip: 192.168.8.254
        remoteasn: 64640
        password: cGFzc3dvcmQ=
...
``

For compatibility, both formats can be supported at the same time.

**Additional context**
If having multiple peers, and for example we want some fields have defaults or not specified (e.g. a password), the list can read confusing. Also, right now there are many settings applicable for a peer, and restructuring them into a real struct would enhance readability, and also would make processing it in code much simpler.

aauren:
This seems reasonable. We should also probably replace the options `--peer-router-passwords-file` (https://github.com/cloudnativelabs/kube-router/blob/master/pkg/options/options.go#L195) with an option to instead pass a YAML config file that has the same structure as the annotation.

That would allow the user multiple ways to specify the same data either at runtime via node annotation or via configuration file that is passed to kube-router at startup. The same Go object should be able to decode both.

PR #1914, opened 2025-09-09
Consolidate Node Peer Annotations

catherinetcai:
Adds support for peer configuration that's in YAML format according to: https://github.com/cloudnativelabs/kube-router/issues/1393

I consolidated the peer configurations from both the consolidated annotation and single annotations into populating a single config struct.

Tested this by spinning up a cluster using https://github.com/aauren/kube-router-automation and adding the following annotations onto the aws-controller and aws-worker nodes and then running `kubectl exec` into the pods to validate that the BGP configurations were picked up. 

```yaml
apiVersion: v1
kind: Node
metadata:
  name: aws-controller
  annotations:
    kube-router.io/peer.ips: "10.95.0.254"
    kube-router.io/peer.asns: "4200000001"
    kube-router.io/peer.localips: "192.168.1.0"
```

```yaml
apiVersion: v1
kind: Node
metadata:
  name: aws-worker
  annotations:
    kube-router.io/peers: |
      - remoteip: 10.0.0.1
        remoteasn: 64640
        password: cGFzc3dvcmQ=
        localip: 192.168.0.1
```

I'm going to keep this MR in the draft state until I'm able to do a more thorough testing cycle with kubetest2. 

@aauren


pkg/controllers/routing/network_routes_controller.go
```diff
@@ -43,12 +44,14 @@ const (
 	nodeCustomImportRejectAnnotation = "kube-router.io/node.bgp.customimportreject"
 	pathPrependASNAnnotation         = "kube-router.io/path-prepend.as"
 	pathPrependRepeatNAnnotation     = "kube-router.io/path-prepend.repeat-n"
-	peerASNAnnotation                = "kube-router.io/peer.asns"
-	peerIPAnnotation                 = "kube-router.io/peer.ips"
-	peerLocalIPAnnotation            = "kube-router.io/peer.localips"
+
+	peerASNAnnotation     = "kube-router.io/peer.asns"
```
aauren, line 48: We should probably put a comment above these saying that they're deprecated or some such so that we don't have to carry them indefinitely.

We should also update the markdown docs with the same message

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -1084,106 +1088,22 @@ func (nrc *NetworkRoutingController) startBgpServer(grpcServer bool) error {
...
-					klog.Errorf("Failed to stop bgpServer: %s", err2)
-				}
-
-				return fmt.Errorf("failed to parse node's Peer Local Addresses Annotation: %s", err)
-			}
-		}
 
 		// Create and set Global Peer Router complete configs
-		nrc.globalPeerRouters, err = newGlobalPeers(peerIPs, peerPorts, peerASNs, peerPasswords, peerLocalIPs,
+		nrc.globalPeerRouters, err = newGlobalPeers(peerCfgs.RemoteIPs(), peerCfgs.Ports(), peerCfgs.RemoteASNs(), peerCfgs.Passwords(), peerCfgs.LocalIPs(),
```
aauren, line 1106: We could probably consolidate this to make it just accept the peerCfg instead of piece-mealing it out like this.

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -1496,3 +1416,182 @@ func NewNetworkRoutingController(clientset kubernetes.Interface,
 
 	return &nrc, nil
 }
+
+type bgpPeerConfig struct {
```
aauren, line 1420: I've been slowly trying to pull functionality out of the monolith controllers. I would do that here. As a matter of fact, there is a relatively recent `bgp` package at the top `pkg` level now.

Probably the best thing to do would be to externalize this, struct and its related functions (although I think most of them already are) and then put them in that package to continue grouping BGP functionality together.

aauren (commented):
@catherinetcai Thanks for starting work on this!

Most of my comments are nits, there are no major problems that I can see with this work. I like the fact that you pulled out all of the BGP parsing stuff into functions on a struct. I think that's already going a long ways towards making the code more readable.

The other main thing I was looking for was keeping backwards compatibility, but it looks like you've done that pretty well.

The configuration structure itself looks pretty good, although I'm still mulling that over a bit to see if there would be some way to make some of the items less duplicated. For instance, people are likely to have similar `Password`, `RemoteASN`, and `RemoteIP` for many of the peers. Maybe it would be possible to define a global default that people could specify once, and then allow individual entries to override that default when necessary?

Or maybe we could introduce a concept of peer groups? https://github.com/osrg/gobgp/blob/master/docs/sources/configuration.md?plain=1#L175-L187 which would follow more of the gobgp standard?

FRR has this standard as well: https://github.com/aauren/kube-router-automation/blob/main/ansible/playbooks/roles/bgp_router/templates/frr.conf.j2#L36-L42

pkg/controllers/routing/network_routes_controller.go
catherinetcai: Updated the markdown docs with the new annotation + examples.

pkg/controllers/routing/network_routes_controller.go
catherinetcai: Hows this?

pkg/controllers/routing/network_routes_controller.go
catherinetcai: Moved this into the `bgp` package. I also noticed that there's a lot of duplication with the validation logic for the BGP peer configs, so I decided to lift that out from the routing package as well and put it into the `bgp` package.

pkg/bgp/peer_config.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	if len(remoteIPs) != len(b64EncodedPasswords) && len(b64EncodedPasswords) != 0 {
+		return errors.New("invalid peer router config. The number of passwords should either be zero, or " +
+			"one per peer router. Use blank items if a router doesn't expect a password. Example: \"pass,,pass\" " +
+			"OR [\"pass\",\"\",\"pass\"]")
+	}
+	if len(remoteIPs) != len(ports) && len(ports) != 0 {
+		return fmt.Errorf("invalid peer router config. The number of ports should either be zero, or "+
+			"one per peer router. If blank items are used, it will default to standard BGP port, %s. "+
+			"Example: \"port,,port\" OR [\"port\",\"\",\"port\"]", strconv.Itoa(options.DefaultBgpPort))
```
aauren: If we return an error here, then we wouldn't actually use the default option right? I believe that returning an error here would mean that the BGP server would fail to start in `startBGPServer()` right?

Its possible that I misunderstood the code path here, but if I'm right, then we may want to change the helpers above to have an `else` statement so that we always have a default. Then remove the default blurb from this message as at this point we'll already have the default?

Alternatively we could turn these into a warn or an info log rather than an error.

docs/bgp.md
```diff
@@ -75,11 +75,45 @@ For example:
...
+- `port` (optional): BGP port (defaults to 179 if not specified)
+
+```shell
+kubectl annotate node <kube-node> \
+kube-router.io/peers="$(cat <<'EOF'
+- remoteip: 192.168.1.99
+  remoteasn: 65000
+  password: U2VjdXJlUGFzc3dvcmQK,
+- remoteip: 192.168.1.100
+  remoteasn: 65000'
```
aauren: We have a trailing apostrophe here, and trailing commas on the base64 passwords I think.

pkg/controllers/routing/network_routes_controller.go
aauren: Much better!

pkg/controllers/routing/network_routes_controller_test.go
```diff
@@ -2618,7 +2621,157 @@ func Test_routeReflectorConfiguration(t *testing.T) {
...
+				peerLocalIPAnnotation:  "192.168.0.1,192.168.0.2",
+			},
+			nil,
+			false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			bgpPeerCfgs, err := bgpPeerConfigsFromAnnotations(tc.nodeAnnotations)
```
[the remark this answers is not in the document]
catherinetcai: 🤦🏼‍♀️ Super embarrassing, should have caught that. Sorry. This was sloppy.

docs/bgp.md
catherinetcai: Ah, fail, thank you.

pkg/bgp/peer_config.go
catherinetcai: > I looked back at the original implementation and while I think this validation causes there to be an earlier exit, the end result seems to ultimately be the same.
> 
> Going to just write out my interpretation of the original logic since it's very likely I'm just reading this incorrectly:
> 
>     * We grab the list of peer ports [here](https://github.com/cloudnativelabs/kube-router/blob/00b46196001975939be6c1672f42a1f8e7990cc9/pkg/controllers/routing/network_routes_controller.go#L1134-L1148)
> 
>     * We then pass that list of peer ports (along with everything else) to [newGlobalPeers](https://github.com/cloudnativelabs/kube-router/blob/00b46196001975939be6c1672f42a1f8e7990cc9/pkg/controllers/routing/bgp_peers.go#L303-L307). This is where I lifted the validation logic from. This does the same length check and returns an error if the length of the ports don't match the length of IPs.
> 
>     * If an error is returned from newGlobalPeers, we end up [calling StopBgp](https://github.com/cloudnativelabs/kube-router/blob/00b46196001975939be6c1672f42a1f8e7990cc9/pkg/controllers/routing/network_routes_controller.go#L1199-L1206)
> 
> 
> I wonder based off of [this comment](https://github.com/cloudnativelabs/kube-router/blob/00b46196001975939be6c1672f42a1f8e7990cc9/pkg/controllers/routing/network_routes_controller.go#L1136-L1137) if the original implementation had a bug and was supposed to actually look like this:
> 
> ```go
> // Default to default BGP port if port annotation is not found
> var peerPorts = make([]uint32, len(peerIPs))
> 
> for i := range peerPorts {
>   peerPorts[i] = options.DefaultBgpPort
> }
> ```
> 
> Or - the validation was supposed to warn (like you pointed out) and not actually return an error, since later in newGlobalPeers, we do construct the Peer structs with a [default port](https://github.com/cloudnativelabs/kube-router/blob/00b46196001975939be6c1672f42a1f8e7990cc9/pkg/controllers/routing/bgp_peers.go#L339-L345).
> 
> Anyway - I think this might be a moot point since I don't think this was desirable behavior to begin with, so I'll update this MR to follow what you were suggesting 😄

Ah never mind. You can ignore me. I'm just apparently arguing with myself now. The thing I was missing again was that the original implementation meant you had peer ports in the form of `179,,179` to make it pass the length check. Sorry - I'm finally picking up what you're putting down now.

go.mod
```diff
@@ -1,89 +1,81 @@
 module github.com/cloudnativelabs/kube-router/v2
```
aauren: It looks like the rebase here, favored your older library versions instead of the ones that were updated by dependabot. We should fix this up to favor the upgraded library versions (as well as Go 1.25).

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -1098,107 +1135,30 @@ func (nrc *NetworkRoutingController) startBgpServer(grpcServer bool) error {
...
-		if !ok {
-			klog.Infof("Could not find BGP peer info for the node in the node annotations so " +
-				"skipping configuring peer.")
-			return nil
-		}
-
-		asnStrings := stringToSlice(nodeBgpPeerAsnsAnnotation, ",")
-		peerASNs, err := stringSliceToUInt32(asnStrings)
+		klog.V(2).
+			Infof("Attempting to construct peer configs from annotation: %+v", node.Annotations)
```
aauren: 2 things here... In general, I try never to split lines on chained functions in the kube-router code base unless there is no other option. In this case, I think it would be better to either wrap the parameter `node.Annotations` or if that doesn't get it below the char limit, then I would split the string across multiple lines.

I'm a bit iffy on this. Technically tier 2 logging should probably only be used for debugging, but technically this can also leak BGP passwords in logs. I could see going both ways on this, but usually its better to lean more secure even if that gives people less debug info.

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -1098,107 +1135,30 @@ func (nrc *NetworkRoutingController) startBgpServer(grpcServer bool) error {
...
-			return fmt.Errorf("failed to parse node's Peer ASN Numbers Annotation: %s", err)
+			return err
 		}
-
-		// Get Global Peer Router IP Address configs
-		nodeBgpPeersAnnotation, ok := node.Annotations[peerIPAnnotation]
-		if !ok {
-			klog.Infof("Could not find BGP peer info for the node in the node annotations " +
-				"so skipping configuring peer.")
+		// Early exist because no BGP peer info was set in annotations for the node
```
aauren: nitpic, but it should be `Early exit`

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -470,7 +501,8 @@ func (nrc *NetworkRoutingController) watchBgpUpdates() {
 					if path.NeighborIp == "<nil>" {
 						return
 					}
-					klog.V(2).Infof("Processing bgp route advertisement from peer: %s", path.NeighborIp)
+					klog.V(2).
+						Infof("Processing bgp route advertisement from peer: %s", path.NeighborIp)
```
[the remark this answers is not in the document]
aauren: It looks like this can be configured in golines via: https://github.com/segmentio/golines?tab=readme-ov-file#chained-method-splitting

pkg/controllers/routing/network_routes_controller.go
aauren: Same here for breaking the line during function chaining.

The more that I think about it, I don't think that logging the full annotation when debugging is increased to level 2 is a big deal. People should expect some more "secret" type information to get displayed when they get into deep debugging levels.

aauren (approved):
@catherinetcai - There is still 3 open nits, but none of them are important enough to block merging. If you're able to resolve them, great, but even if not we'll get this in.

Can you do one final rebase where you resolve the conflicts with go.mod and go.sum, and pair down the number of commits into logical blocks?

I'll be on the lookout to merge right away after this is done so that it 🤞 doesn't get any more conflicts before the final merge.

catherinetcai:
Closing in favor of https://github.com/cloudnativelabs/kube-router/pull/1972, which is the squashed version of this

PR #1919, opened 2025-09-27
fix(ipset): ignore non-kube-router ipsets

### LINKED ISSUE #1918: Race condition resulting in other ipsets being modified
## What happened?

I originally encountered this issue on k3d/k3s, specifically [v1.32.6+k3s1](https://github.com/k3s-io/k3s/releases/tag/v1.32.6%2Bk3s1) which included kube-router 2.5.0. When deploying [istio](https://istio.io/latest/) with the istio-cni in ambient mode, Istio manages two ipsets used for healthprobe traffic. I experience intermittent behavior where these ipsets would unexpectedly be missing certain pod IPs, despite istio's logs indicating them being added.

After some testing I identified that the behavior disappeared after disabling network policy within k3s (the component that uses kube-router). With the proper log level I was able to see this in kube-router logs:
```console
I0924 15:40:28.817837      27 policy.go:83] Attempting to attain ipset mutex lock
I0924 15:40:28.817840      27 policy.go:85] Attained ipset mutex lock, continuing...
I0924 15:40:28.818554      27 ipset.go:595] ipset (ipv6? false) restore looks like:
create TMP-HHYMGYLRMJXZ3ET7 hash:ip family inet hashsize 1024 maxelem 65536 comment
flush TMP-HHYMGYLRMJXZ3ET7
create istio-inpod-probes-v4 hash:ip family inet hashsize 1024 maxelem 65536 comment
swap TMP-HHYMGYLRMJXZ3ET7 istio-inpod-probes-v4
flush TMP-HHYMGYLRMJXZ3ET7
create TMP-PTNDKBUZZRYGF25Q hash:ip family inet6 hashsize 1024 maxelem 65536 comment
flush TMP-PTNDKBUZZRYGF25Q
create istio-inpod-probes-v6 hash:ip family inet6 hashsize 1024 maxelem 65536 comment
swap TMP-PTNDKBUZZRYGF25Q istio-inpod-probes-v6
flush TMP-PTNDKBUZZRYGF25Q
destroy TMP-HHYMGYLRMJXZ3ET7
destroy TMP-PTNDKBUZZRYGF25Q
I0924 15:40:28.819189      27 policy.go:183] Restoring IPv4 ipset took 657.136µs
```

As best I can tell there seems to be a race condition where kube-router's save/restore is racing with istio-cni adding an IP to its sets, resulting in inconsistent behavior where IPs go missing.

## What did you expect to happen?

kube-router to not affect other ipsets present on my node.

An exact reproduction of my issue is in this github gist: https://gist.github.com/mjnagel/d790315dc93553a2853830e22675b71a

This involves:
- make a k3d cluster with the latest 1.33.x image
- istioctl install with proper ambient/cni values for k3s/k3d
- apply a strict peerauth so that probes get blocked if they are proxied through ztunnel
- apply a namespace in ambient mode + a deployment with a bunch of replicas to reproduce the issue faster

- Kube-Router Version (`kube-router --version`): 2.5.0, consumed through k3s v1.32.6+k3s1
- Kube-Router Parameters: unsure - this is consumed through k3s
- Kubernetes Version (`kubectl version`) : v1.32.6
- Cloud Type: local machine
- Kubernetes Deployment Type: k3s
- Kube-Router Deployment Type: part of k3s
- Cluster Size: 1 node

## Logs, other output, metrics

aauren:
Attempt to filter out sets that we are not authoritative for to avoid race conditions with other operators (like Istio) that might be attempting to modify ipsets at the same time.

This still needs some refinement and probably a refactor or two, but wanted to get it out early so that you could test it and see if it actually resolves what you were seeing in #1918. I tried it in my cluster, and after proving that I could see kube-router modifying other ipsets, then applied this change, and saw that it no longer contained references in its output.

FYI @mjnagel @ilrudie

Attempts to fix #1918

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller_test.go
+++ b/pkg/controllers/netpol/network_policy_controller_test.go
@@ -741,6 +741,10 @@ func (ips *fakeIPSet) Restore() error {
 	return nil
 }
 
+func (ips *fakeIPSet) RestoreSets(_ []string) error {
+	return nil
+}
+
 func (ips *fakeIPSet) Flush() error {
 	return nil
 }
--- a/pkg/controllers/netpol/policy.go
+++ b/pkg/controllers/netpol/policy.go
@@ -152,9 +152,11 @@ func (npc *NetworkPolicyController) syncNetworkPolicyChains(networkPoliciesInfo 
 		}
 	}
 
+	knownIPSets := utils.ConvertMapKeysToSlice[string](activePolicyIPSets)
+
 	for ipFamily, ipset := range npc.ipSetHandlers {
 		restoreStart := time.Now()
-		err := ipset.Restore()
+		err := ipset.RestoreSets(knownIPSets)
 		restoreEndTime := time.Since(restoreStart)
 
 		if npc.MetricsEnabled {
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -710,7 +710,7 @@ func (nsc *NetworkServicesController) syncIpvsFirewall() error {
 
 		setHandler.RefreshSet(serviceIPPortsSetName, serviceIPPortsIPSets[family], utils.TypeHashIPPort)
 
-		err := setHandler.Restore()
+		err := setHandler.RestoreSets([]string{serviceIPsIPSetName, serviceIPPortsSetName})
 		if err != nil {
 			return fmt.Errorf("could not save ipset for service firewall: %v", err)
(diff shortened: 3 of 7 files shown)


mjnagel:
This seems to behave as expected, not modifying Istio ipsets in testing. I went through testing with the steps in [this gist](https://gist.github.com/mjnagel/fdd72fd0cbe7150a24255a51965f460d). When using the latest image it failed due to the linked issue, but with an image built from this branch everything worked.

PR #1941, opened 2025-11-03
build(deps): bump github.com/ccoveille/go-safecast from 1.6.1 to 1.8.1 - ccoVeille version

ccoVeille:
Closes #1937 

Related to https://github.com/ccoVeille/go-safecast/issues/126 and https://github.com/ccoVeille/go-safecast/pull/122

PR #1943, opened 2025-11-04
feat(test): use safecast.RequireConvert in tests

ccoVeille:
I'm the owner and maintainer of [ccoVeille/go-safecast](https://github.com/ccoVeille/go-safecast)

This is somehow a follow-up of #1941

I have noticed you were checking the err returned in a safecast conversion in a test.

my lib supports RequireConvert that try to convert and fails the provided test.

This PR can be merged separately of #1942, either before or after.

[v1](https://pkg.go.dev/github.com/ccoveille/go-safecast#RequireConvert) and [v2](https://pkg.go.dev/github.com/ccoveille/go-safecast/v2#RequireConvert) supports RequireConvert

So it's safe.

Here my PR is just a suggestion.


aauren:
Sorry that it took so long for us to get to this!

Thanks for even more contributions and looking out for users of your library! Appreciate you!

PR #1950, opened 2025-11-14
fix: do not advertise Pod IPv4 CIDR in a Cluster composed of IPv6-only Nodes

abbshr:
in case of pods subnet configured with both IPv4 and IPv6 CIDR, but nodes only have IPv6 addresses.

kube-router will report error and won't exit:
```
2025-11-13T12:27:06.433971867+08:00 stderr F E1113 04:27:06.433945  789979 network_routes_controller.go:393] Error advertising route: invalid nexthop address: <nil>
```

even though providing the key command line options as follow:
```
              "--run-router=true",
              "--run-firewall=false",
              "--run-service-proxy=true",
              "--enable-ipv4=true",
              "--enable-ipv6=true",
              "--router-id=generate",
              "--enable-overlay=true",
```

as a result, the BGP rib will be empty, and the results are quite confusing.

so i modify the code, when nodes in a IPv6-only environment, kube-router won't advertise the IPv4 Pod CIDR.

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -498,43 +498,46 @@ func (nrc *NetworkRoutingController) advertisePodRoute() error {
 	}
 
 	// Advertise IPv4 CIDRs
-	nodePrimaryIPv4IP := nrc.krNode.FindBestIPv4NodeAddress()
-	if nrc.krNode.IsIPv4Capable() && nodePrimaryIPv4IP == nil {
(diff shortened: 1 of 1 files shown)


aauren:
Thanks for this submission @abbshr!

I'm sorry to say that this was a slightly obvious oversight. I guess even with IPv6 support added I was still treating IPv4 addresses as first-class citizens.

aauren:
Fix contained in v2.6.3

PR #1963, opened 2025-12-01
Fix EndpointSlice -> Service Resolution

### LINKED ISSUE #1957: kube-router is too strict in EndpointSlice verification
## What happened?

I'm running `rook` that manages `ceph` cluster for me.

`kube-router` reports once a second:

```
E1128 02:07:36.021982  331328 network_services_controller.go:996] unable to lookup service from EndpointSlice, skipping: the ownerReferences field on EndpointSlice (rook-ceph) doesn't agree with with the kubernetes.io/service-name label (rook-ceph-active-mons) for rook-ceph/rook-ceph-active-mons-ipv4 EndpointSlice
```

https://github.com/cloudnativelabs/kube-router/blob/1ce59e7fab33259b1e2244740201dc126005f5d1/pkg/utils/service.go#L52-L70

I have made a research and found that in the initial KEP (https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/0752-endpointslices) `EnpointSlice` indeed was supposed to have the `kubernetes.io/service-name` value and the owner name to match.

**BUT**

If we look into the current implementation: https://github.com/kubernetes/kubernetes/blob/4bed36e03e7bd699b089d33da6f7d7c9ef9eb661/pkg/apis/discovery/types.go#L24

we will see that those constraint has never landed the actual kubernetes code (or documentation).

So `kube-router` is being too pedantic in verification.

## What did you expect to happen?

Given nothing (?) in kubernetes code or documentation requires those to be identical - `kube-router` should not treat it as an error. And possibly just skip with no (?) logging at all, as it's not an error in any way.

Create an `EndpointSlice` with different service name and owner name.

1. Step 1
2. Step 2
3. Step 3
4. Step 4

## Screenshots / Architecture Diagrams / Network Topologies

- Kube-Router Version (`kube-router --version`): 2.6.2
- Kubernetes Version (`kubectl version`) : 1.32.6
- Kubernetes Deployment Type: [e.g. EKS, GKE, Kops, Kubeadm, etc.]
- Kube-Router Deployment Type: [e.g. DaemonSet, System Service]
- Cluster Size: [e.g. 200 Nodes]

## Logs, other output, metrics

aauren:
Do not return an error when the `kubernetes.io/service-name` label and the `metadata.ownerReference` don't agree with each other, just prefer the Kubernetes service name label which is similar to all other major service providers.

Fixes #1957 

FYI @zerkms - If you can, try out this change and just double-check that this fixes the error you were experiencing.

### DIFF
--- a/pkg/utils/service.go
+++ b/pkg/utils/service.go
@@ -73,7 +73,7 @@ func ServiceNameforEndpointSlice(es *discoveryv1.EndpointSlice) (string, error) 
 }
 
 // ServiceForEndpoints given EndpointSlice object return Service API object if it exists
-func ServiceForEndpointSlice(ci *cache.Indexer, es *discoveryv1.EndpointSlice) (interface{}, bool, error) {
+func ServiceForEndpointSlice(ci *cache.Indexer, es *discoveryv1.EndpointSlice) (any, bool, error) {
 	svcName, err := ServiceNameforEndpointSlice(es)
 	if err != nil {
 		return nil, false, err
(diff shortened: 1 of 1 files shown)


zerkms:
Grepped through the code - wouldn't it after that trigger another warning:

```
klog.Errorf("failed to convert endpoint slice resource to service: %s", err)
```

Indeed, it looks like in this case it would happen much more rarely - only once the `EndpointSlices` has updated, but still, if it's not an error - would it also need a downgrade severity level too?

zerkms:
And yep - I can confirm `cloudnativelabs/kube-router-git:PR-1963` works as intended and does not produce anything unexpected on the default logging level anymore.

PR #1972, opened 2025-12-15
Consolidate Node Annotations - Squashed

catherinetcai:
Just this MR https://github.com/cloudnativelabs/kube-router/pull/1914 - but squashed. Done so out of paranoia since I don't want to squash then force push and lose things.


aauren:
Thanks Cat!

Issue #1965, opened 2025-12-03
Sporadic `failed to get list of addrs: results may be incomplete or inconsistent`


zerkms:
## What happened?

Now that https://github.com/cloudnativelabs/kube-router/issues/1962 returns the complete error message - I can finally see that it's

```
W1203 03:00:02.919551    3323 network_routes_controller.go:608] unable to get subnet for node IP: 10.35.9.6, err: failed to get list of addrs: results may be incomplete or inconsistent... skipping
W1203 03:00:02.951787    3323 linux_tunnels.go:139] failed to get tunnel link by name tun-aa2ffae2d77: Link not found, will attempt to create it
I1203 03:00:02.952835    3323 linux_tunnels.go:222] failed to check if fou is enabled on the link tun-aa2ffae2d77: failed to get link by name: Link not found, going to try to clean up and recreate the tunnel anyway
W1203 03:00:02.986407    3323 linux_tunnels.go:478] failed to list fou ports: no such file or directory
I1203 03:00:02.992879    3323 linux_tunnels.go:278] Creating tunnel tun-aa2ffae2d77 with encap ipip for destination 10.35.9.9
```

My speculation: apparently, during node boot up - various OS/kubernetes components mess with the network configuration. Hence if `kube-router` starts concurrently with everything else, there is a chance that sometimes it will clash and fail.

## What did you expect to happen?

A suggestion: what if `LocalLinkQuerier::AddrList` implementation was a wrapper over `vishvananda/netlink` and performed several retries in case if `nl.ErrDumpInterrupted` is returned? Eg: retry 10 times with 1 second interval.

As a temporary (or permanent? :-D ) mitigation I'm having an `initContainer` that `sleep 10`, which apparently is enough to avoid the clashes, but it's feels "meh" :-) 

## How can we reproduce the behavior you experienced?

1. Step 1
2. Step 2
3. Step 3
4. Step 4

## Screenshots / Architecture Diagrams / Network Topologies

If applicable, add those here to help explain your problem.

## System Information (please complete the following information)

- Kube-Router Version (`kube-router --version`): [e.g. 1.0.1] 2.6.3
- Kube-Router Parameters: [e.g. --run-router --run-service-proxy --enable-overlay --overlay-type=full etc.]
- Kubernetes Version (`kubectl version`) : [e.g. 1.18.3] 1.32.x
- Cloud Type: [e.g. AWS, GCP, Azure, on premise] on premise
[...]

PR #1974, opened 2025-12-25
fix(nrc): Retry netlink.Handle functions that raise netlink.ErrDumpInterrupted Errors

catherinetcai:
NOTE: Will pull this MR out of draft state once I've been able to test it in a K8s cluster

Addresses the error surfaced in https://github.com/cloudnativelabs/kube-router/issues/1965, but only partially addresses the overall issue.

I think this MR is a good starting point for retrying the ErrDumpInterrupted errors. There are a few other places that call netlink functions that raise this error. 

Taking suggestions for the retry/backoff approach. I don't have a specific rationale for picking the retry interval/max retry attempt numbers, though I did some research into other libraries to see what they did since the [netlink docs](https://docs.kernel.org/userspace-api/netlink/intro.html#dump-consistency) don't provide any hints on quickly/often to retry:

* [Cilium](https://github.com/cilium/cilium/blob/87fe8c00e8fddecea52a9d1386a0d08b2b9342b6/pkg/datapath/linux/safenetlink/netlink_linux.go#L25) retries 30 times with constant 1 millisecond intervals
* [AWS VPC CNI](https://github.com/dcoppa/amazon-vpc-cni-k8s/blob/cdfc4c573bdcdfc8994b67311bccc09416a38f91/pkg/netlinkwrapper/netlink.go#L78-L100) retries 5 times with 100 millisecond intervals
* [containernetworking CNI plugins](https://github.com/containernetworking/plugins/blob/8c4a6b3bebbd5b11707665aefe621fc236a30ea7/pkg/netlinksafe/netlink.go#L56-L63) retries 5 times without waiting

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,12 +105,12 @@ ifeq "$(BUILD_IN_DOCKER)" "true"
 		-v $(GO_MOD_CACHE):/go/pkg/mod \
 		-w /go/src/github.com/cloudnativelabs/kube-router $(DOCKER_BUILD_IMAGE) \
 		sh -c \
-		'go install gotest.tools/gotestsum@latest && CGO_ENABLED=0 gotestsum --format gotestdox -- -timeout 30s github.com/cloudnativelabs/kube-router/v2/cmd/kube-router/ github.com/cloudnativelabs/kube-router/v2/pkg/...'
+		'go install gotest.tools/gotestsum@latest && CGO_ENABLED=0 gotestsum --format gotestdox -- -timeout 30s github.com/cloudnativelabs/kube-router/v2/cmd/kube-router/ github.com/cloudnativelabs/kube-router/v2/...'
 else
 ifeq ($(shell command -v gotestsum 2>/dev/null),)
 	go install gotest.tools/gotestsum@latest
 endif
-	gotestsum --format gotestdox -- -timeout 30s github.com/cloudnativelabs/kube-router/v2/cmd/kube-router/ github.com/cloudnativelabs/kube-router/v2/pkg/...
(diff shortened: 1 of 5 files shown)


Makefile
aauren: Can you update this statement and the one below it the same way you did for the `test` target above?

(at some point I'm going to need to refactor this Makefile a bit 🙁 )

internal/nlretry/handle.go
```diff
@@ -0,0 +1,35 @@
...
+// raise netlink.ErrDumpInterrupted.
+// Creates an empty netlink.Handle if the handle passed in is nil.
+func NewHandle(h *netlink.Handle) *handle {
+	if h == nil {
+		h = &netlink.Handle{}
+	}
+	return &handle{Handle: h}
+}
+
+func (r *handle) AddrList(link netlink.Link, family int) ([]netlink.Addr, error) {
```
aauren: Not really a big deal for this one considering the call chain, but since we are just starting this paradigm, we should probably accept the context from the caller here and pass that down so that in the future, other functions where it matters more will follow suit and be able to cancel the retries call if needed.

catherinetcai: I was split on this as well.

I don't mind updating it to accept the context from the caller instead. Both the wrapped Handler and the LinkQuerier interface are both for internal usage anyway, so there's no reason for us to have to expose the exact same interface as the underlying netlink library.

aauren: I guess if you don't mind, we might as well do it as I think it will be handy in future extensions to be able to cancel netlink calls if they run too long.

aauren (approved):

internal/nlretry/handle.go
aauren: Perfecto!

PR #1975, opened 2025-12-29
fix: Add retry wrapper to netlink function calls that raise ErrDumpInterrupted

catherinetcai:
Follow up to https://github.com/cloudnativelabs/kube-router/pull/1974

I wrapped as many of the netlink function that raise the ErrDumpInterrupted errors as I could find, though it's possible I've missed a few.

Will pull this out of draft once I've had a chance to test it E2E.

### DIFF
--- a/internal/nlretry/addr.go
+++ b/internal/nlretry/addr.go
@@ -6,6 +6,9 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
+// AddrList is a wrapper around netlink.AddrList that retries with exponential backoff on
+// netlink.ErrDumpInterrupted. The retry can be interrupted and the result/error will be returned if the
+// context is cancelled.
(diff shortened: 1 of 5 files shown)


internal/nlretry/addr.go
```diff
@@ -0,0 +1,14 @@
+package nlretry
+
+import (
+	"context"
+
+	"github.com/vishvananda/netlink"
+)
+
+func AddrList(ctx context.Context, link netlink.Link, family int) ([]netlink.Addr, error) {
```
aauren: nit(non-blocking): It would be nice to add godoc for each of these new public functions just as a matter of best practice and to make IDE integration nicer and be explicit about what the intent is (even if the code is pretty obvious as it is)

catherinetcai: How's this? Sorry if the comments are really repetitive, but I think it should make the intent relatively clear 😬

aauren:
Thanks @catherinetcai !

PR #1985, opened 2026-01-14
feat(retry.go): add debug logging to retries

aauren:
Brings visibility to #1965 

FYI @catherinetcai


catherinetcai:
> Brings visibility to #1965
> 
> FYI @catherinetcai

Thank you!

PR #1999, opened 2026-01-25
Improve Network Services Tests - Part 1

aauren:
This PR seeks to improve our overall test coverage in the Network Service Controller. Especially after I recently found via #1997 that we were basically not testing the NetworkServiceController (NSC) at all because Ginkgo tests were not configured correctly. :disappointed: 

This is part of what I hope is a larger set of PRs seeking to improve the test-ability of the NSC. I started off by prioritizing TrafficPolicy and DSR related tests which have historically been our biggest source of regressions and bugs.

This initial PR also does a lot of the leg-work of mocking out the various parts of the NSC that have so far not been mocked and frequently require host changes or change the host itself.

This PR stacks with #1998


catherinetcai (approved):
👏🏼

Issue #848, opened 2020-02-29
Monitoring metrics do not report external BGP peers


salanki:
The Prometheus metric `kube_router_controller_bgp_peers` does not include eBGP peers. Some setups, ie. when peering with a ToR switch this is the only type of peer that is used. Monitoring proper function is impossible in these setups.

As the node can be in `Ready` state even with 0 BGP peers traffic from any pod scheduled on a node without the peer established will be blackholed, leading to service interruptions.

I'd suggest either including eBGP peers in the `kube_router_controller_bgp_peers` count, or adding a separate metric for it.

It would also make sense to add an option signal the CNI as NonReady until a peer has been established.

murali-reddy:
@salanki thanks for suggestion. I am marking the bug for backlog of improvements.

PR #2003, opened 2026-02-14
feat(gobgp): add kube_router_bgp_peer_info metric

damex:
Replace the misleading kube_router_controller_bgp_peers gauge
  which only counts 'cluster nodes'. with a new per peer metric
  kube_router_bgp_peer_info with 'GaugeVec' that exposes actual
  BGP session state from gobgp. labels include peer address, asn,
  and session state. metric value is 1 if established and 0 otherwise.

Closes: https://github.com/cloudnativelabs/kube-router/issues/848

### DIFF
--- a/pkg/controllers/routing/network_routes_controller.go
+++ b/pkg/controllers/routing/network_routes_controller.go
@@ -436,6 +436,10 @@ func (nrc *NetworkRoutingController) Run(
 			nrc.syncInternalPeers()
 		}
 
+		if nrc.MetricsEnabled {
+			nrc.updateBGPPeerMetrics()
+		}
+
 		if err == nil {
 			healthcheck.SendHeartBeat(healthChan, healthcheck.NetworkRoutesController)
 		} else {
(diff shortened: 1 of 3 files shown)


damex:
i did see that workflows got an approval and received message about this:

https://github.com/cloudnativelabs/kube-router/actions/runs/22016039123/job/63619708028 pipeline here complained about not handling SESSION_STATE_UNSPECIFIED.

i originally map all others states and had fallback to 'unknown' (unspecified is supposed to be 'unknown'). so i have added mapping for this 'unspecified' case too and it satisfies linter.

i did run lint locally and it passed
```
damex@foosha:~/Documents/kube-router$ git status
On branch feature/implement-bgp-peers-info-metric
Your branch is up to date with 'origin/feature/implement-bgp-peers-info-metric'.

nothing to commit, working tree clean
damex@foosha:~/Documents/kube-router$ golangci-lint run
0 issues.
```

please run pipeline again

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -729,6 +733,61 @@ func (nrc *NetworkRoutingController) injectRoute(path *apiutil.Path) error {
...
+	metrics.ControllerBGPPeerInfo.Reset()
+	err := nrc.bgpServer.ListPeer(context.Background(), &gobgpapi.ListPeerRequest{}, func(peer *gobgpapi.Peer) {
+		if peer.Conf == nil || peer.State == nil {
+			return
+		}
+		peerAddress := peer.Conf.NeighborAddress
+		peerType := bgpPeerTypeString(peer.State.GetType())
+		peerASN := strconv.FormatUint(uint64(peer.Conf.PeerAsn), 10)
+		peerState := bgpSessionStateString(peer.State.SessionState)
+		var value float64
```
catherinetcai: Very likely that I'm missing something so apologies ahead of time here if I'm misinterpreting the intent here! 

I think the convention I've seen for gauge metrics like this, where all of the useful information is packed into the labels, is to uniformly set all of the values to 1. 

Mixing the 0 and 1 values is a bit confusing for two reasons: 

1) It implies there an additional dimension of information being conveyed through the value, when there isn't: it's actually redundant with the state label.
2) It makes querying the metrics inconsistent. If you have all of them uniformly set to 1, you can see which peers are in which state by doing: `sum by (state) (bgp_peer_info)`. I suppose you can also use `count by (state) (bgp_peer_info)` to achieve the same effect and ignore the values, but `sum by (x)` is a common pattern with Prometheus queries.

aauren:
Hey @damex apologies for getting to this late.

Thanks for the contribution! Everything looks good here, and I agree with all of @catherinetcai's suggestions. I tried deploying it and it looks good to me.

PR #2051, opened 2026-04-03
feat(healthcheck): make healtcheck bind address configurable

dodgex:
#### What type of PR is this?

feature

This PR makes the bind address used for the healthcheck endpoint configurable.
Similar to and pretty much based on: https://github.com/cloudnativelabs/kube-router/pull/1570 that makes metrics bind address configurable.

Fixes https://github.com/cloudnativelabs/kube-router/issues/2043

No

None

```release-note
NONE
```

This PR is pretty much a copy/paste and adapting of changes from https://github.com/cloudnativelabs/kube-router/pull/1570.

The development and contribution guides ask/require to execute certain commands. Unfortunately, some of them fail for me.

- `make kube-router` and `make gofmt` are successful.
- `make gofmt-fix` fails as it misses `goimports`. I tried to run `go install golang.org/x/tools/cmd/goimports@latest` but the fixing still failed with the same error. As I did not change imports, I assume this can be "ignored".
- `make test-pretty` fails with `DONE 592 tests, 1 failure in 0.062s`, where the failing test is 
   > kube-router_test.go:53: docs/user-guide.md 'command line options' section does not match `kube-router --help`.
- plain `make` fails with
  > docker run --rm -v /home/dodge/projects/kube-router:/work -w /work alpine:3.23 sh -c \
        'wget -qO- https://github.com/crate-ci/typos/releases/download/v1.33.1/typos-v1.33.1-x86_64-unknown-linux-musl.tar.gz | tar xz -C /usr/local/bin && typos'
28FBA63FD4730000:error:0A000126:SSL routines::unexpected eof while reading:ssl/record/rec_layer_s3.c:698:
ssl_client: SSL_connect
wget: error getting response: Connection reset by peer
tar: invalid magic
tar: short read
make: *** [Makefile:139: spellcheck] Error 1

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -20,6 +20,9 @@
   - [Docker](#docker)
   - [containerd](#containerd)
 - [trying kube-router as alternative to kube-proxy](#trying-kube-router-as-alternative-to-kube-proxy)
+- [Binding health check endpoint to a specific host IP](#binding-health-check-endpoint-to-a-specific-host-ip)
+  - [Using the primary host address via Downward API](#using-the-primary-host-address-via-downward-api)
+  - [Using a specific IP address](#using-a-specific-ip-address)
 - [Advertising IPs](#advertising-ips)
 - [External IP and LoadBalancer IP Validation](#external-ip-and-loadbalancer-ip-validation)
   - [How It Works](#how-it-works)
@@ -130,6 +133,7 @@ Usage of kube-router:
       --gobgp-admin-address string                    Address for GoBGP server. Used in combination with gobgp-admin-port to expose GoBGP for administrative purposes. Setting this to empty string will default the address to 127.0.0.1. (default "127.0.0.1")
       --gobgp-admin-port uint16                       Port to connect to GoBGP for administrative purposes. Setting this to 0 will disable the GoBGP gRPC server. (default 50051)
       --hairpin-mode                                  Add iptables rules for every Service Endpoint to support hairpin traffic.
+      --health-addr string                            Health check address to listen on (Default: all interfaces; ensure to configure the 'livenessProbe' to use the same IP address)
       --health-port uint16                            Health check port, 0 = Disabled (default 20244)
   -h, --help                                          Print usage information.
       --hostname-override string                      Overrides the NodeName of the node. Set this if kube-router is unable to determine your NodeName automatically.
@@ -289,6 +293,62 @@ and if you want to move back to kube-proxy then clean up config done by kube-rou
 ```
 
 and run kube-proxy with the configuration you have.
+
+## Binding health check endpoint to a specific host IP
+
+Using `--health-addr` you can bind health check to specific host IP.
+When using this option, you have to ensure, that the `livenessProbe` is configured to use the same IP address.
+
+### Using the primary host address via Downward API
+
+You can use the Kubernetes [Downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/) to determine
+the primary IP address of the node to which the Pod is assigned and use it for the health check.
+
+```yaml
+env:
+  - name: HOST_IP
+    valueFrom:
+      fieldRef:
+        fieldPath: status.hostIP
+args:
+  - --health-addr=$(HOST_IP)
+```
+
+Ensure to configure the `livenessProbe` to use the `HOST_IP` as the `host` value.
+
+```yaml
+livenessProbe:
+  httpGet:
+    path: /healthz
+    port: 20244
+    host: $(HOST_IP)
+    scheme: HTTP
+  initialDelaySeconds: 10
+  periodSeconds: 10
+```
+
+### Using a specific IP address
+
+You can also use a specific IP address for the health check. For example 127.0.0.1 to bind locally, or any other
+available IP address.
+
+```yaml
+args:
+  - --health-addr=127.0.0.1
+```
+
+You still need to configure the `livenessProbe` to use the chosen IP address for the `host` value.
+
+```yaml
+livenessProbe:
+  httpGet:
+    path: /healthz
+    port: 20244
+    host: 127.0.0.1
+    scheme: HTTP
+  initialDelaySeconds: 10
+  periodSeconds: 10
+```
 
 ## Advertising IPs
 
--- a/pkg/cmd/kube-router.go
+++ b/pkg/cmd/kube-router.go
@@ -107,6 +107,14 @@ func (kr *KubeRouter) Run() error {
 	defer close(healthChan)
 	stopCh := make(chan struct{})
 
+	// Verify the health address/port combo provided is listenable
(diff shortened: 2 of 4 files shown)


catherinetcai:
Hi @dodgex, sorry for the delay in getting to this MR!

The health check implementation is missing a check to ensure that the address/port combination is actually bindable. The metrics server does something similar here: https://github.com/cloudnativelabs/kube-router/blob/d2359f280ece15a34bd91e54250d1f5c997c4af8/pkg/cmd/kube-router.go#L148

Can you also provide a little more context around your home lab setup. I'd like to get more insight onto why having the health check bind to all interfaces is causing problems for you.

dodgex:
Hi @catherinetcai, no worries. 

I added a check if the healt address/port is bindable, based on the check for the metrics.

> Can you also provide a little more context around your home lab setup. I'd like to get more insight onto why having the health check bind to all interfaces is causing problems for you.

Well, as I mentioned in https://github.com/cloudnativelabs/kube-router/issues/2043, I am currently more or less in an evalating phase, testing k0s for personal stuff. So currently there is no actual kubernetes setup yet, beside my playground / PoC VM. My current experience with Kubernetes so far is only the "using" side, providing YAML files (svc, deployments etc) for our applications and managing the stuff installed on our cluster with ArgoCD. This project is me liking tech stuff way to much and wanting to learn more about it. :)

I would not say, that having health check bind on all interfaces is actually causing real problems. This is honestly mostly me, like stuff "clean". And as I managed to configure all other ports im my playground (one thanks to a [similar PR in k0s](https://github.com/k0sproject/k0s/pull/7384) itself, that got merged today), I thought, it might be helpful, to contribute what is needed to make it possible.

And even when I, in the next phase of this project decide to roll k0s on my online servers, this wouldn't be that big of a deal, as the firewall would block any access anyway. 

tldr: It is pretty much all about limiting the scope to what is needed. And a healtcheck, as harmless as it might be, has no reasons (i can think of) to be available on more interfaces than necessary. :)

catherinetcai (approved):

docs/user-guide.md
```diff
@@ -129,6 +129,7 @@ Usage of kube-router:
       --excluded-cidrs strings                        Excluded CIDRs are used to exclude IPVS rules from deletion.
       --gobgp-admin-port uint16                       Port to connect to GoBGP for administrative purposes. Setting this to 0 will disable the GoBGP gRPC server. (default 50051)
       --hairpin-mode                                  Add iptables rules for every Service Endpoint to support hairpin traffic.
+      --health-addr string                            Health check address to listen on, (Default: all interfaces)
```
aauren: Could you add some help text further down in this guide that talks about how to set this in a dynamic way (via the k8s downward API)?

The default is safe for everyone and represents the existing functionality, however, my worry is that people that try to set this without understanding it will shoot themselves in the foot a bit. If they specify a health address that is not the one that Kubernetes is looking for in the health probe, then kube-router will fail the health check and just repeatedly kill the pod.

It might be nice to give a full example of how this setting could be configured to:

1. Find the correct IP of the host dynamically via the downward API, something like:
```yaml
env:
  - name: HOST_IP
    valueFrom:
      fieldRef:
        fieldPath: status.hostIP
args:
  - --health-addr=$(HOST_IP)
```
2. Use that same address in the health check configuration for kube-router:
```yaml
livenessProbe:
  httpGet:
    path: /healthz
    port: 20244
    host: $(HOST_IP)
    scheme: HTTP
  initialDelaySeconds: 10
  periodSeconds: 10
```

An alternative config could be to just bind locally, which doesn't involve downward API, but still needs a liveness probe change:
1. Bind to 127.0.0.1
```yaml
args:
  - --health-addr=127.0.0.1
```
2. Update that in the health check configuration for kube-router:
```yaml
livenessProbe:
  httpGet:
    path: /healthz
    port: 20244
    host: 127.0.0.1
    scheme: HTTP
  initialDelaySeconds: 10
  periodSeconds: 10
```

aauren (changes requested):
Thanks for the contribution @dodgex !

From my perspective this looks good, I am just requesting some documentation that gives users some direction on how to use this option safely and not shoot themselves in the foot.

With these change I think this change should be good to go.

docs/user-guide.md
dodgex: Thanks for your feedback. Adding an extra help text is indeed a good idea. In addition to the help text in the user guide, I added a little note about `livenessProble` to the flag in options.go.

aauren:
Thanks @dodgex !

I just approved the CI pipeline (should have done that during my last pass :disappointed: ).

It looks like linting is failing on some trailing spaces in the user guide.

Unit testing is failing because the `kube-router --help` text has to match the code block that's in the user-guide.

docs/user-guide.md
aauren: Thanks @dodgex !  Looks great!

aauren (approved):

dodgex:
The single trailing space was an accident after adding the new line in the file to keep the line length short. Fixed that.

It seems, that in the test the `--help` command results in giving `livenessProbe` instead of `string` as the type... I have no real clue why this is. But I suspect it was caused by the backtick I used to quote the word in the description. this assumption is further validated by the lack of thos ticks in the output. I replaced them with single quotes `'`

test expected ouput
> --health-addr livenessProbe                     Health check address to listen on, (Default: all interfaces; ensure to configure the livenessProbe to use the same IP address)

vs guide
> --health-addr string                            Health check address to listen on, (Default: all interfaces; ensure to configure the `livenessProbe` to use the same IP address)


Also I removed the stray coma `, ` from the description.

catherinetcai:
Sorry @dodgex - looks like linting is still failing.

aauren:
Dang you got hit by the long line linter. If you have `make` on your development host you should just be able to run that to ensure that you don't have any surprises. There are one or two minor checks the Makefile isn't able to do, but in general we try to keep it 1:1 what the CI pipeline uses it to prevent frustration like this.

dodgex:
Sorry for the inconvenience. I fixed the long line and ran `make lint`. I also tried to run just `make`, but that fails with the error below, as I am running in WSL and it seems my PC/Hostname is full caps in WSL.

The full `make` ran the `doctoc` target, so the ToC of the user-guide.md is now correctly updated too.

> ERROR: failed to build: invalid tag "cloudnativelabs/MY-PCNAME-git:amd64-make-health-addr-configurable": repository name must be lowercase

aauren:
Interesting! Thanks for the feedback, this may be the first time that we've tried running on WSL. I have a Windows VM so I'll try to fix up any other problems in the makefile and submit a subsequent PR for that to make the windows experience better.

aauren:
Thanks for your contribution @dodgex!

dodgex:
> Thanks for your contribution @dodgex!

Thank you both for your patience in getting this ready.

PR #2064, opened 2026-04-29
feat: implement nftables-only mode for NetworkPolicyController

thomasferrandiz:
#### What type of PR is this?

feature

This PR implements an nftables-only mode using the knftables library (created for kube-proxy and also used in flannel).
A new flag "--use-nftables-for-netpol" is added to switch from iptables mode to nftables mode.

The implementation requires a refactor of NetworkPolicyController: all the common parts are moved to NetworkPolicyControllerBase and the backend-dependent parts are implemented in
NetworkPolicyControllerIptables and NetworkPolicyControllerNftables respectively.

New unit, integration and e2e tests are added to ensure that both implementations have similar behaviour.

* What tool was used: Copilot and Claude
* To what extent was the tool used? The main refactoring was designed manually. AI was used to generate nftables-code equivalent to the existing iptables code. AI was also used to add most test cases and to handle minor changes like linting errors.
* Help us understand if a human was in the loop or not for this PR? Yes. 
The refactoring and new feature was added  step by step and each one was designed and tested by a human.
I also tested manually the integration with the k3s project where kube-router's NetworkPolicyController is used extensively.

Integration tests and e2e tests were added (with the help of AI) to check that the behaviour of the iptables and nftables-based netpol controllers are similar.

```release-note
A new flag "--use-nftables-for-netpol" is added to switch from iptables mode to nftables mode.
The default behaviour is unchanged.

```

If integrated in kube-router, the code will be used in k3s so I can help to maintain it.


thomasferrandiz:
This PR is not directly related but I see that there is already an issue about nftables in kube-router here: https://github.com/cloudnativelabs/kube-router/issues/2034

aauren:
Hey @thomasferrandiz! Good to see you again!

Sweeping in with another large contribution I see! :wink: 

I'm just going to be honest, it may take me a few to get to review this one. I had some thoughts on how I wanted to approach a secondary IP filtering implementation. Mainly with a bit of an eye to see if there was any (easy-ish) way to wrap it into an interface that would allow us to substitute implementations.

My lofty end goal, would be to be able to provide 3 different IP filtering implementations. One that is `iptables` based, one that is `nftables` based, and one that is `ebpf` based (I've been keeping a bit of an eye on things like https://github.com/facebook/bpfilter).

I've had several stabs at this, but I haven't found an abstraction level that I like just yet. But I appreciate this contribution all the same! Anyway, I'll probably try to take one or two more pokes, if I find something that I like, then I would look to wrap this work up into that. If I can't find anything that seems worth it (which has already happened several times) then I'll probably just accept your work here directly and then extend it to the other controllers (routing / service).

FYI @catherinetcai

thomasferrandiz:
> Hey @thomasferrandiz! Good to see you again!
> 
> Sweeping in with another large contribution I see! 😉
> 
> I'm just going to be honest, it may take me a few to get to review this one. I had some thoughts on how I wanted to approach a secondary IP filtering implementation. Mainly with a bit of an eye to see if there was any (easy-ish) way to wrap it into an interface that would allow us to substitute implementations.
> 
> My lofty end goal, would be to be able to provide 3 different IP filtering implementations. One that is `iptables` based, one that is `nftables` based, and one that is `ebpf` based (I've been keeping a bit of an eye on things like https://github.com/facebook/bpfilter).
> 
> I've had several stabs at this, but I haven't found an abstraction level that I like just yet. But I appreciate this contribution all the same! Anyway, I'll probably try to take one or two more pokes, if I find something that I like, then I would look to wrap this work up into that. If I can't find anything that seems worth it (which has already happened several times) then I'll probably just accept your work here directly and then extend it to the other controllers (routing / service).
> 
> FYI @catherinetcai

Glad to be back :-)

I understand that it will take some time to review such a large PR.
I wasn't planning for it to be so large but it was a larger undertaking than I thought.

test/e2e/netpol/framework_test.go
```diff
@@ -0,0 +1,374 @@
...
+		Stdout: &stdout,
+		Stderr: &stderr,
+	})
+	exitCode := 0
+	if err != nil {
+		// remotecommand wraps non-zero exit codes as errors; treat them as
+		// connection failures rather than fatal test errors.
+		exitCode = 1
+	}
+	return stdout.String(), stderr.String(), exitCode, nil
```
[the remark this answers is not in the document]
thomasferrandiz: done

test/e2e/netpol/netpol_test.go
```diff
@@ -0,0 +1,1019 @@
...
+					exitCode, err, blocked)
+				return blocked
+			}, consistentlyDuration, pollInterval).Should(BeTrue(),
+				"API server access should be blocked by egress policy")
+		})
+	})
+})
+
+// Ensure the intstr import is used (named port helper).
+var _ = intstr.FromString
```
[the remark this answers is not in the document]
thomasferrandiz: done

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -935,53 +216,43 @@ func NewNetworkPolicyController(clientset kubernetes.Interface,
...
+		// in case of a restart with a different configuration
+		npc := NetworkPolicyControllerIptables{NetworkPolicyControllerBase: &NetworkPolicyControllerBase{}}
+		npc.Cleanup()
+		return NewNetworkPolicyControllerNftables(&npcBase, clientset, config,
+			podInformer, npInformer, nsInformer, linkQ, knftInterfaces)
+	} else {
+		// Cleanup any existing nftables rules before starting iptables controller to avoid conflicts
+		// in case of a restart with a different configuration
+		npc := NetworkPolicyControllerNftables{NetworkPolicyControllerBase: &NetworkPolicyControllerBase{}}
+		npc.Cleanup()
```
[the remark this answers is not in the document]
thomasferrandiz: done

.github/workflows/ci-e2e-netpol.yml
```diff
@@ -0,0 +1,185 @@
...
+
+    - name: Load kube-router image into kind
+      run: kind load docker-image "${{ env.KUBE_ROUTER_IMAGE }}" --name e2e
+
+    - name: Deploy kube-router (${{ matrix.netpol-backend }} NetworkPolicy backend)
+      run: |
+        MANIFEST=$(cat daemonset/generic-kuberouter.yaml \
+          | sed 's|[^ ]*cloudnativelabs/kube-router[^ ]*|kube-router:e2e-test|g' \
+          | sed 's|imagePullPolicy: Always|imagePullPolicy: Never|g')
+        if [[ "${{ matrix.netpol-backend }}" == "nftables" ]]; then
```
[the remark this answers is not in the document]
thomasferrandiz: done

.github/workflows/ci-e2e-netpol.yml
```diff
@@ -0,0 +1,185 @@
...
+          MANIFEST=$(echo "$MANIFEST" \
+            | yq '(select(.kind == "DaemonSet") | .spec.template.spec.containers[0].args) += ["--use-nftables-for-netpol=true"]')
+        fi
```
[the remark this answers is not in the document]
thomasferrandiz: done, yq is not used anymore

.github/workflows/ci-e2e-netpol.yml
```diff
@@ -0,0 +1,185 @@
...
+  KIND_NODE_IMAGE: "kindest/node:v1.32.2"
+
+jobs:
+  # Executes Integration Tests (require nft binary + kernel nftables support) 
+  ci-integration-tests:
+    name: ci-integration-tests
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v6
```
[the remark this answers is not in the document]
thomasferrandiz: done

thomasferrandiz:
@aauren Would you have some time to check the PR?
I think all the comments from the AI code review have been answered and I rebased it to include the `default-deny` flag that you added.

aauren:
> @aauren Would you have some time to check the PR?
> I think all the comments from the AI code review have been answered and I rebased it to include the `default-deny` flag that you added.

Thanks @thomasferrandiz - Sorry I've been a bit busy lately, but looking to get this looked at as the next thing now that I got #2106 finished. Appreciate you taking the first pass with the bots.

pkg/cmd/kube-router.go
```diff
@@ -263,26 +273,37 @@ func (kr *KubeRouter) Run() error {
...
+		nftCancel()
 		if err != nil {
-			return fmt.Errorf("failed to create iptables handlers: %w", err)
+			return fmt.Errorf("failed to create nftables interfaces: %v", err)
 		}
 		npc, err := netpol.NewNetworkPolicyController(kr.Client,
-			kr.Config, podInformer, npInformer, nsInformer, &ipsetMutex, nil, iptablesCmdHandlers, ipSetHandlers,
-			ipValidator)
+			kr.Config, podInformer, npInformer, nsInformer, &ipsetMutex, nil,
+			iptablesCmdHandlers, ipSetHandlers, ipValidator, knftablesInterfaces, kr.Config.UseNftablesForNetpol)
```
aauren: issue(blocking): This is already a fairly large instantiation, we probably need to begin bundling up some of these into a struct that can be passed around or some such. Maybe one that is informer based and one that is filter based?

Also, we already pass in `kr.Config` as the 2nd parameter, we probably can reference that instead of passing `kr.Config.UseNftablesForNetpol` as a separate boolean right?

docs/user-guide.md
```diff
@@ -180,6 +180,7 @@ Usage of kube-router:
       --service-tcpfin-timeout duration               Specify TCP FIN timeout for IPVS services in standard duration syntax (e.g. '5s', '1m'), default 0s preserves default system value (default: 0s)
       --service-udp-timeout duration                  Specify UDP timeout for IPVS services in standard duration syntax (e.g. '5s', '1m'), default 0s preserves default system value (default: 0s)
       --strict-external-ip-validation                 When enabled, the proxy module validates externalIPs and loadBalancerIPs against configured CIDR ranges (--service-external-ip-range and --loadbalancer-ip-range). When strict mode is enabled and no range is configured, all externalIPs / loadBalancerIPs are rejected (default-deny). Disable this flag to restore previous behavior of accepting all IPs without validation. (default true)
+      --use-nftables-for-netpol                       Use nftables instead of iptables for the network policy controller. Requires nft >= 1.0.1 and kernel nftables support.
```
aauren: issue(blocking); We probably want to add some specific docs for this. The user-guide itself is already getting a bit long, so maybe we start a dedicated doc around filtering or policy?

Specifically, we should add a bit more information about the difference between nftables and iptables backed policy. The fact that by default it uses iptables for all controllers, and spelling out just a bit more about how this flag only applies to the network policy controller (even though it should be self-evident by the name, doesn't hurt to be even more concrete as often users view kube-router as one lump sum of capabilities rather than distinct controllers).

Maybe also adding something about it being experimental? You've done a ton of testing so we're about as sure as we can be, but until its been out in the wild for a few releases, it may still have conflicts or bugs on some hosts. Especially when it comes to the interplay of the NRC and NPC controllers that still potentially use the iptables backend on some legacy systems.

test/e2e/netpol/suite_test.go
```diff
@@ -0,0 +1,75 @@
+// Package netpol_e2e contains end-to-end tests for the kube-router
```
aauren, line 1: praise: We've been looking at adding e2e tests, but it just never came around, thanks for taking this opportunity to add them and verify the rules!

.github/workflows/ci-e2e-netpol.yml
```diff
@@ -0,0 +1,190 @@
+name: ci-e2e-netpol
```
aauren, line 1: suggestion(non-blocking): One of the things that we really try to do is get local test flows to match up with CI whenever possible. This helps developers have confidence that things that pass make will also end up being valid PRs rather than pushing and getting surprised. This also helps prep for new releases of kube-router without making a bunch of post PR adjustments.

Would you be willing to abstract this logic out into something like a `hack/e2e-netpol.sh` shell script that could be both referenced from this CI file as well as from the Makefile?

I would guess that we could have a `test-integration` target that looks something like:

```make
test-integration:
	$(DOCKER) run --privileged -v $(PWD):/go/src/... $(DOCKER_BUILD_IMAGE) \
		sh -c 'apk add --no-cache nftables && go test -tags integration \
		-run TestIntegration -timeout 60s .../pkg/controllers/netpol/...'
```

This may even fit well within `test-privileged` which does something almost identical.

And we could also have a e2e target that would execute the same script that e2e-netpol.sh that CI does and create a kind cluster and run the e2e tests.

I know that's a bit of work, but it would have a lot of value, especially to me, what are your thoughts on implementing something like that?

pkg/controllers/netpol/npc_nftables.go
```diff
@@ -0,0 +1,1505 @@
+package netpol
```
aauren, line 1: question(blocking): What types of performance testing did you do with this?

The reason why I ask is that my read of the code shows that we're calling knftables.Run() on a per policy / per pod approach rather than as a single large transaction that saves in one chunk. Additionally, knftables has taken the approach of exec'ing out to the [nft binary](https://github.com/kubernetes-sigs/knftables/blob/master/README.md?plain=1#L17-L31) rather than using the API or c library wrapper.

My guess is that enabling this in a large cluster would run us into some of the same performance problems that caused us to have to initially refactor the iptables integration. On some large clusters, it could sometimes take upwards of 5 minutes to fully sync the full policy. This one is at least grouped by policy and pod, so we're probably not up at that level yet, but my read of the code is that on a dual-stack node that had say 50 pods and 50 policies, and assuming approximately 10ms of overhead for process startup and the like, we'd be looking at something like 50 seconds of processing time right?

Is this assumption correct?

aauren (changes requested):

.github/workflows/ci-e2e-netpol.yml
thomasferrandiz: I agree, it can be useful to have as much tests as possible locally.

pkg/controllers/netpol/npc_nftables.go
thomasferrandiz: I haven't had time to do any perf testing yet unfortunately.
There is certainly room for improvements in the way nftables is called at the moment.

I think your guess is correct so we could look for ways to reduce the number of calls to minimize that overhead.

test/e2e/netpol/suite_test.go
thomasferrandiz: Thanks!

pkg/cmd/kube-router.go
thomasferrandiz: Good catch, passing the boolean is completely unnecessary.

I agree that instantiating the NPC is getting unwieldy.
I can try to refactor things a bit but I'm not sure this PR is the right place to do that.

pkg/controllers/netpol/npc_nftables.go
aauren: Ok, I think that as long as the documentation outlines that this is a first pass at adding nftables integration and may not be the most performant choice for medium to large clusters, we can iterate in the future and make this more performant. Mostly just double-checking.

pkg/cmd/kube-router.go
aauren: Fair enough. We can come back to it in the future PR, especially as the informers would involve touching all of the various controller initializations.

aauren:
@thomasferrandiz - It looks like you're working on implementing the local test-ability. Thanks for that!

Once you have that and the documentation updated, let me know and I'll do another quick pass and we can likely merge it.

docs/user-guide.md
thomasferrandiz: I added a short doc about the new flag with its known limitations.

thomasferrandiz:
@aauren I think all the changes are done. The e2e tests can run locally and I added a short doc about using nftables in the NPC.

aauren:
Thanks for following up on the feedback. Let me test these changes during the week before you fix the conflicts.

twz123:
I can give this a spin in the k0s integration tests, as well, if that'd be helpful.

aauren:
> I can give this a spin in the k0s integration tests, as well, if that'd be helpful.

I would appreciate it @twz123 !

aauren:
Thanks again for working through this with us @thomasferrandiz and also for bearing with us. I was hoping to get to this earlier in the week, but it feels like this week has just come and gone in a flash.

I ran the nftables netpol path head-to-head against the iptables implementation on the same dual-stack cluster over two sessions. In general, the nftables implementation held up really well. I tested it with over 30 different network policy regressions that have come up in the past, and in general, it passed most of them. What follows are the four breaks I hit, all of which reproduce only on nftables mode and none of which reproduce under iptables.

## BREAK-1: ipBlock `except` return skips later rules (enforcement divergence)

**Symptom:** A peer excluded by one rule's `ipBlock.except` stays blocked even when a later rule in the same policy should allow it. Kubernetes ORs all rules together, so this wrongly rejects traffic the iptables path allows (it uses ipset `nomatch`, which only fails that one match).

**Reproduction:** Apply a two-rule ingress policy where rule 1 is `ipBlock 10.242.0.0/16 except <client-b>/32` and rule 2 is `namespaceSelector team=b`, then connect from client-b to server:8080. It refuses on nftables, allows on iptables; the egress variant fails identically.

**Code pointers:** The `ip saddr @<except-set> counter return` in `processIngressRulesNft` and `processEgressRulesNft` (`pkg/controllers/netpol/npc_nftables.go:767` and `:871`) exits the whole per-policy chain instead of just failing that single ipBlock match.

## BREAK-2: stale chain/set GC never succeeds (resource leak)

**Symptom:** Policy chains embed the sync version, so every full sync orphans the previous set, and none of them ever get collected. I watched chains grow monotonically (20 to 78 for one active policy over ~15 min in round 1, and 123 chains for 5 policies by the end of round 2) with `Resource busy` log spam every sync.

**Reproduction:** Apply and delete a few policies over a few minutes on any node, then count the `KUBE-NWPLCY-*` chains in `kube-router-filter-ipv4` and watch it climb. Identical churn on iptables stays bounded at 4 chains.

**Code pointers:** The GC blocks in `syncNetworkPolicyChains` (`pkg/controllers/netpol/npc_nftables.go:967`) run before `syncPodFirewallChains` (`:1104`) rewrites the pod-fw jumps, so the old chains are still referenced and deletion fails; because every stale deletion shares one transaction, the one guaranteed failure aborts the whole batch.

## BREAK-3: allow-all window during every full sync (critical bypass)

**Symptom:** Every full sync flushes the base chains (which have `policy accept`) in one transaction, then restores the per-pod jumps and the TAIL jump in later, separate transactions, so for a slice of every sync no policy is enforced on the node. Because KUBE-NWPLCY-COMMON accepts ESTABLISHED, any connection that sneaks through stays usable indefinitely.

**Reproduction:** Apply only `default-deny-ingress`, confirm client-b to server:8080 refuses at rest, then hammer a few hundred connects from client-b while forcing full syncs with roughly one no-op policy event per second. I saw ~37% of attempts leak through on nftables versus zero on iptables.

**Code pointers:** `fullPolicySync` (`pkg/controllers/netpol/npc_nftables.go:163`) calls `ensureTopLevelChains` (`:234`), which flushes the base chains in its own transaction, while the jumps come back only later via `syncPodFirewallChains` (`:1104`) and `ensureTailChainPosition` (`:1460`).

## BREAK-4: switching back to iptables leaves nftables enforcing (mode-switch trap)

**Symptom:** After I removed `--use-nftables-for-netpol`, the `kube-router-filter-*` tables stayed hooked with policy state frozen at switch time, so the node kept rejecting traffic based on stale rules while policy changes only reached iptables. Recovery needed a manual `nft delete table` on every node.

**Reproduction:** Run nftables mode with `default-deny-ingress`, drop the flag, wait for rollout, delete all NetworkPolicies, then test client-b to server:8080. It still refuses with zero policies in the cluster, and `nft list tables` still shows both kube-router-filter tables.

**Code pointers:** The iptables branch guards its nftables cleanup with `if len(knftInterfaces) > 0` at `pkg/controllers/netpol/network_policy_controller.go:278`, but `NewKnftablesInterfaces` (`npc_nftables.go:49`) returns nil when `UseNftablesForNetpol` is false, so that cleanup is dead code in iptables mode. Since `NetworkPolicyControllerNftables.Cleanup()` (`:1478`) already builds its own interfaces when the slice is empty, the guard can likely just be dropped, with care for hosts that lack nftables entirely.

docs/nftables.md
```diff
@@ -0,0 +1,62 @@
...
+
+- `nft` >= 1.0.1
+- Kernel with nftables support enabled (Linux >= 3.13; most modern distributions satisfy this)
+
+## Usage
+
+Pass the flag when starting kube-router:
+
+```bash
+kube-router --run-network-policy-controller --use-nftables-for-netpol
```
twz123: You probably mean `--run-firewall --use-nftables-for-netpol` here? NB: `--run-firewall` defaults to true anyways.

twz123:
> > I can give this a spin in the k0s integration tests, as well, if that'd be helpful.
> 
> I would appreciate it @twz123 !

Looks good from  the k0s integration tests and the CNCF conformance tests point of view:

* https://github.com/k0sproject/k0s/actions/runs/29326824624
* https://github.com/k0sproject/k0s/actions/runs/29326944284

(Note that the Oracle 9.3 tests are broken due to the AMIs no longer being available, and the single test failure for Alpine 3.24 is a flake.)

thomasferrandiz:
Thanks for testing @aauren.
I tried to fix each break in a separate commit to make the changes easier to review.

When you're ok with the code, I can squash the commits and rebase the branch.

aauren:
Hey @thomasferrandiz - I still found a few regressions after your fixes. I went ahead and fixed them myself and pushed to your branch.

I'm not putting them through another round of integration testing using the k8s conformance tests. Let me see how these pan out over the next few days. Assuming there are no gotchas in there, we should finally be ready to land this plane.

aauren (approved):
Alright, @thomasferrandiz - I think that this PR is finally ready to go. I found a few other bugs by going through the k8s conformance suite, most resulted in nft policy errors that caused further policies to not apply. But as far as I can tell all of these are patched up now.

Can you do a final rebase to fix the conflict on this PR, and then squash some of your other changes?

thomasferrandiz:
@aauren Thanks. The conflict is fixed. Do you want me to also squash all your commits into one so the PR is just two commits?

aauren:
> @aauren Thanks. The conflict is fixed. Do you want me to also squash all your commits into one so the PR is just two commits?

No, that's ok. Thanks Thomas!

PR #2077, opened 2026-05-09
Update gobgp dep

aauren:
#### What type of PR is this?

dependency

Updates GoBGP and Grype and enables secure version pinning to ensure against supply chain attacks.

None

* What tool was used: Claude
* To what extent was the tool used? Just to update the dependency-updater code

```release-note
NONE
```

### DIFF
--- a/build/dependency-updater/config/config_test.go
+++ b/build/dependency-updater/config/config_test.go
@@ -72,7 +72,6 @@ func TestLookupToolRepo(t *testing.T) {
 		{"GORELEASER_VERSION", "goreleaser/goreleaser", true},
 		{"CNI_VERSION", "containernetworking/plugins", true},
 		{"TYPOS_VERSION", "crate-ci/typos", true},
-		{"GRYPE_VERSION", "anchore/grype", true},
 		{"UNKNOWN_VERSION", "", false},
 	}
 	for _, tt := range tests {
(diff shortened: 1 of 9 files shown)


build/dependency-updater/updater/classify_test.go
```diff
@@ -26,6 +26,8 @@ func TestClassifyMakeVar(t *testing.T) {
 		{"GORELEASER_VERSION", "v2.13.3", updater.KindToolVersion},
 		{"DOCTOC_VERSION", "2.3.0", updater.KindToolVersion},
 		{"CNI_VERSION", "v1.9.0", updater.KindToolVersion},
+		// Tool version — commit SHA from prior SHA-pinning run
+		{"GOBGP_VERSION", "abc123def456789012345678901234567890abcd", updater.KindToolVersion},
 		// Derived (Makefile expansion)
 		{"GRYPE_IMAGE", "anchore/grype:$(GRYPE_VERSION)", updater.KindDerived},
```
catherinetcai: Might be missing something, but I think this needs to be updated since GRYPE_VERSION got removed from the Makefile

aauren: Good catch, I don't think we need to test that path any longer. I've removed that stuff.

build/dependency-updater/config/config_test.go
```diff
@@ -88,3 +87,25 @@ func TestLookupToolRepo(t *testing.T) {
 		})
 	}
 }
+
+func TestShouldPinBySHA(t *testing.T) {
```
catherinetcai, line 91: Is this test meant to be a guard against accidentally adding anything else here? 

I just worry it's not a test that's a ton of value to add otherwise...

catherinetcai (approved):

build/dependency-updater/config/config_test.go
aauren: I think this one is still valuable for testing against regressions of handling to GoBGP version i.e. it pinning by version instead of by SHA.

Issue #873, opened 2020-04-08
The establishment of iptables rules lags behind Pod traffic occurrence


jimmy-zh:
**What happened:**

In some cases, NPC's establishment of iptables rules lags behind Pod traffic occurrence, which results in network isolation failing to take effect. Maybe this is a knotty problem in consideration of K8s event synchronization mechanism.

**How to reproduce it:**
1. Deploy the following `Network Policy` in the K8s cluster:
```
apiVersion: extensions/v1beta1
kind: NetworkPolicy
metadata:
  name: test-np
  namespace: diagnosis
spec:
  egress:
  - to:
    - ipBlock:
        cidr: 14.215.0.0/16
  podSelector: {}
  policyTypes:
  - Egress
```
2. Deploy the following `Deployment` in the K8s cluster:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: diagnosis
  namespace: diagnosis
spec:
  replicas: 2
  selector:
    matchLabels:
      app: diagnosis
  template:
    metadata:
      labels:
        app: diagnosis
    spec:
      nodeName: 10.0.1.30
      containers:
      - name: diagnosis
        command: ["/bin/sh"]
        args: ["-c","ping 220.181.38.150"]
        image: jimmyzh/diagnosis:v1
```
3. Observe the logs of certain Pods:
```
kubectl logs diagnosis-66f959fd76-2wpln -n diagnosis | head

PING 220.181.38.150 (220.181.38.150) 56(84) bytes of data.
64 bytes from 220.181.38.150: icmp_seq=1 ttl=53 time=3.66 ms
64 bytes from 220.181.38.150: icmp_seq=2 ttl=53 time=3.57 ms
64 bytes from 220.181.38.150: icmp_seq=3 ttl=53 time=3.58 ms
```
4. Observe the iptables rules:
```
iptables-save -t filter

-A KUBE-POD-FW-3F2W5JGDV6M5LYIN -m comment --comment "rule for stateful firewall for pod" -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
-A KUBE-POD-FW-3F2W5JGDV6M5LYIN -m comment --comment "run through nw policy test-np" -j KUBE-NWPLCY-ONWOLJXYLLQ2RYI2
[...]

PR #2079, opened 2026-05-18
Add deny by default to netpol

aauren:
#### What type of PR is this?

feature

Closes the race window where a new pod is routable on the node before its `KUBE-POD-FW-*` chain has been programmed, by introducing `--netpol-default-deny` command line option. The initial commit on this branch added CIDR-scoped REJECTs at the tail of `KUBE-NWPLCY-TAIL`, but those failed for pod-to-pod traffic on the same node because every per-pod chain marks accepted packets with `0x20000`, and the `KUBE-NWPLCY-TAIL` ACCEPT-on-mark rule fired before the CIDR REJECTs could.

This PR finishes the feature by introducing a per-IP-family `kube-router-local-pods` ipset of pod IPs whose firewall chain has been programmed this sync. Two new ipset-gated REJECT rules are prepended to `KUBE-NWPLCY-TAIL` ahead of the ACCEPT-on-mark; they fire for any local pod IP not yet in the set. The existing CIDR-scoped REJECTs stay at the tail as defense-in-depth.

When `--netpol-default-deny` is disabled the ipset is never created and the rule layout is unchanged from today. The feature self-disables (with an error log) when this node's pod CIDRs cannot be detected via `node.spec.podCIDRs` or the `kube-router.io/pod-cidrs` annotation.

The user-guide gets a brief section on the feature; the troubleshooting guide gets a symptom-first entry ("NetworkPolicy not enforced on freshly-launched pods") with the operational details.

Further protects against findings discussed in #873 

* What tool was used: Claude (opencode)
* To what extent was the tool used? Drafted the implementation, tests, and docs on the second commit (`feat(npc): enable default deny for pod<->pod`) and the third commit  (`doc: how --netpol-default-deny works`) from a detailed plan. The first commit on the branch was hand-written.
* If drafted, how detailed of a plan did you create for the AI? Very detailed
* Help us understand if a human was in the loop or not for this PR? Yes — every change was  reviewed and the AI was redirected several times (troubleshooting framing, helper  consolidation, function renames, docstring brevity).

Unit tests cover the new rule layout, ipset refresh, drift recovery, IPv6 prefix handling, and cleanup. Manual cluster validation (deploying the `netpol-race-tester` DaemonSet from `kube-router-automation`, verifying `iptables -L KUBE-NWPLCY-TAIL` and `ipset list kube-router-local-pods`, toggling the flag off and confirming the ipset and extra rules are gone) is the remaining checklist item before merge.

I also tested against a local Kubernetes cluster with a custom pod workload that attempted to make early connections both to cluster internal and cluster external endpoints. I toggled default-deny on and off and watched the workloads.

```release-note
NONE
```

* The feature is gated behind `--netpol-default-deny` (off by default), so this is purely  additive for existing deployments.
* This features **ONLY** works when kube-router is aware of the pod CIDRs on a per-node basis. This happens by default when kube-router is operating the pod routing, and it can also happen with other cluster networking plugins that use `--allocate-node-cidrs=true` from kube-controller-manager. It can also be used by other more custom configurations as long as the user annotates the node with the kube-router `kube-router.io/pod-cidrs` annotation. If neither or these are configured, kube-router cannot safely gate traffic and will warn and continue.
* The per-sync ipset refresh runs **before** `ensureTailChainPosition` installs the TAIL jump in the iptables-restore buffer, so the new rules cannot fire against a stale ipset during a fresh kube-router start.

### DIFF
--- a/docs/troubleshoot.md
+++ b/docs/troubleshoot.md
@@ -38,6 +38,7 @@ solutions.
 - [Network Policy Issues](#network-policy-issues)
   - [Egress Policy Blocking Service Access](#egress-policy-blocking-service-access)
   - [Pods Have Network Access Before Policy is Applied](#pods-have-network-access-before-policy-is-applied)
+  - [NetworkPolicy not enforced on freshly-launched pods](#networkpolicy-not-enforced-on-freshly-launched-pods)
   - [ipBlock Not Matching Real Client IP](#ipblock-not-matching-real-client-ip)
 - [Overlay and Tunnel Issues](#overlay-and-tunnel-issues)
   - [Cross-Subnet Connectivity Failures](#cross-subnet-connectivity-failures)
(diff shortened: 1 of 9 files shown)


docs/troubleshoot.md
```diff
@@ -620,7 +621,103 @@ this window persist because `ESTABLISHED,RELATED` flows are always allowed
...
+```sh
+# What kube-controller-manager allocated (when --allocate-node-cidrs=true is in effect):
+kubectl get node <node-name> -o jsonpath='{.spec.podCIDRs}{"\n"}'
+
+# What the annotation override is set to (empty when no annotation is present):
+kubectl get node <node-name> -o jsonpath='{.metadata.annotations.kube-router\.io/pod-cidrs}{"\n"}'
+```
+
+If neither source is populated, kube-router logs the message below at startup, silently disables
+`--netpol-default-deny`, and continues without protection. Either enable `--allocate-node-cidrs`
```
catherinetcai, line 687: I wonder if it makes more sense if this flag is set for kube-router to fail loudly rather than proceeding silently. I'm guessing users who opt into this are coming from a security conscious posture, so they'd rather have this fail rather than risking having this proceed silently with a warning in logs that you can miss.

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -564,14 +579,245 @@ func (npc *NetworkPolicyController) ensureTopLevelChains() {
...
+}
+
+// ensureLocalPodsIPSetExists makes sure the kube-router-local-pods ipset exists for every IP family. No-op when
+// --netpol-default-deny is disabled. The Save+Sets lookup exists only so we can log a one-shot "was missing"
+// warning on drift; Create itself is idempotent via `create -exist`.
+func (npc *NetworkPolicyController) ensureLocalPodsIPSetExists() {
+	if !npc.defaultDeny {
+		return
+	}
+	if nil != npc.ipsetMutex {
```
catherinetcai: I know it's nil != for consistency with the rest of what's in this file but err - should we start reversing this? Not an error, just a weird stylistic thing.

docs/troubleshoot.md
aauren: That's a fair point, I'll fix it!

pkg/controllers/netpol/network_policy_controller.go
aauren: I removed all of the yoda conditions. Those probably snuck in because of my old Java coding practices. But you're right, it doesn't belong in Go code.

PR #2084, opened 2026-05-18
build(deps): replace docker/docker with moby/moby/client

mrueg:
#### What type of PR is this?

dependency

Migrate from the legacy github.com/docker/docker dependency to the new split module github.com/moby/moby/client. This reduces the dependency footprint by dropping several transitive deps no longer needed (containerd/log, moby/sys/atomicwriter, moby/term, morikuni/aec, pkg/errors, grpc-gateway, otlptrace/otlptracehttp, etc.).

Also add client.WithAPIVersionNegotiation() to the Docker client initialisation in getContainerPidWithDocker so the client negotiates the API version with the daemon automatically.

See also: https://github.com/moby/moby#go-modules for the message about github.com/docker/docker.

* What tool was used: 
* To what extent was the tool used? Create commit message
* If drafted, how detailed of a plan did you create for the AI? No

```release-note

```


aauren (approved):
LGTM

PR #2092, opened 2026-05-22
fix(bgp): also reject defaultRouteSetV6 from peers

### LINKED ISSUE #2091: Kube-router accepts v6 defaultroute from external peers
## What happened?

Actually, I wanted to import IPv4 defaultroute to my k8s cluster from external bgp peers, and found that this does not work. Also found that doing the same with IPv6 did work. Howewer, digging the code, I've realized that the intent is to not accept defaults at all.

1. Set up external bgp peer with kube-router, using address-family ipv6
2. Configure the external peer to originate default
3. Observe that default enters RIB

- Kube-Router Version (`kube-router --version`): `Running kube-router version v2.9.0, built on 2026-04-28T07:01:55+0000, go1.25.9`
- Kube-Router Parameters:
```
        - --enable-ipv4=true
        - --enable-ipv6=true
        - --bgp-graceful-restart
        - --enable-overlay=false
        - --run-router=true
        - --run-firewall=true
        - --run-service-proxy=true
        - --advertise-pod-cidr=true
        - --advertise-external-ip
        - --iptables-sync-period=24h
        - --ipvs-sync-period=24h
        - --routes-sync-period=24h
        - --auto-mtu=false
        - --service-external-ip-range=192.168.8.192/27
        - --service-external-ip-range=192.168.9.0/24
        - --service-external-ip-range=2001:470:20ab:fff:485:387::/96
        - --runtime-endpoint=unix:///var/run/crio/crio.sock
        - --run-loadbalancer=true
        - --loadbalancer-ip-range=192.168.9.0/24
        - --advertise-loadbalancer-ip=true
```
- Kubernetes Version (`kubectl version`) : 1.35.4
- Cloud Type: on premise
- Cluster Size: 10+ nodes

rkojedzinszky:
#### What type of PR is this?

bug

Rejects IPv6 defaultroute advertisements from external peers.

Fixes #2091

No

None

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/routing/bgp_policies.go
+++ b/pkg/controllers/routing/bgp_policies.go
@@ -870,12 +870,12 @@ func (nrc *NetworkRoutingController) addImportPolicies() error {
 			statementNames = append(statementNames, statement.Name)
 		}
 
-		for _, defaultset := range []string{defaultRouteSet, defaultRouteSetV6} {
+		for _, defaultSet := range []string{defaultRouteSet, defaultRouteSetV6} {
 			statement := gobgpapi.Statement{
 				Conditions: &gobgpapi.Conditions{
 					PrefixSet: &gobgpapi.MatchSet{
 						Type: gobgpapi.MatchSet_TYPE_ANY,
-						Name: defaultset,
+						Name: defaultSet,
 					},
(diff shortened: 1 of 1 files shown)


aauren:
@greptileai

PR #2093, opened 2026-05-24
Add better bgp policy unit testing

aauren:
#### What type of PR is this?

bug

Extends the existing IPv6 parity work in `addImportPolicies` (PR #2092) and closes related test-infrastructure gaps:

- **Adds IPv6 support to `customImportReject`** — the `kube-router.io/node.bgp.customimportreject` annotation can now
  contain a mix of V4 and V6 CIDRs. Previously GoBGP rejected the mixed defined set with "multiple families" and the
  import policy then failed referencing the never-created set. CIDRs are now split by family into
  `customImportRejectSet` and `customImportRejectSetV6` with the correct `MaskLengthMax` per family (32 vs 128), and
  matched in the import policy via the same V4/V6 loop pattern used by the default-route reject.
- **Fixes a silent-pass bug in the `checkPolicies` test helper** — the helper was calling GoBGP's `ListPolicy` with an
  exact name filter that never matched the actual policy name (missing the `"1"` version suffix appended by
  `incrementAndCreatePolicy`), so the per-statement validation loop was effectively skipped for every existing test
  case. Fixes the filter and introduces a `statementsEquivalent` helper using `proto.Equal` that tolerates two
  implementation-detail mismatches (statement `Name` and `RpkiResult`) that had drifted across many expected statements
  without anyone noticing.

To meaningfully exercise the now-validating runner, the PR also:

- Adds **dual-stack variants** for the existing external-peers, iBGP-disabled, AS-prepend, communities, and
  customImportReject scenarios in `Test_AddPolicies`.
- Adds a new **`Test_AddDefinedSetContents`** function that verifies the actual prefix/neighbor list contents of every
  relevant GoBGP defined set (so e.g. `defaultRouteSetV6` is now asserted to actually contain `::/0` rather than being
  trusted via a name reference alone).
- **Backfills five V4-only test cases** with the cross-family `defaultRouteSetV6 × allPeerSet` reject statement that
  the V6 default-route fix generates unconditionally for V4 peer sets.

* What tool was used: Claude Code (Sonnet)
* To what extent was the tool used? Drafted most of the code and tests; human directed scope, reviewed each step, and
  made architectural decisions (separation of `Test_AddPolicies` vs `Test_AddDefinedSetContents`, commit splitting,
  whether to fix the runner bug now or later, etc.).
* If drafted, how detailed of a plan did you create for the AI? Iterative — short conversational prompts driving
  discovery, with an explicit TDD cycle requested for the customImportReject feature.

```release-note
NONE
```

- Commits are ordered for narrative readability, not perfect bisectability: the runner-fix commit
  (`fix(NRC): correct exact-match policy name in checkPolicies test runner`) leaves V4-only tests failing until the
  next commit (`test(NRC): backfill expected V6 default-route cross-family statements`) lands. The first commit's
  message explicitly calls this out. Squash the two if strict bisectability is required.
- The only runtime behavior change is the V6 customImportReject support in `bgp_policies.go`. All other commits are
  test-internal.

### DIFF
--- a/pkg/controllers/routing/bgp_policies.go
+++ b/pkg/controllers/routing/bgp_policies.go
@@ -32,9 +32,11 @@ const (
 	iBGPPeerSet       = "iBGPpeerset"
 	iBGPPeerSetV6     = "iBGPpeersetv6"
 
-	customImportRejectSet = "customimportrejectdefinedset"
-	defaultRouteSet       = "defaultroutedefinedset"
-	defaultRouteSetV6     = "defaultroutedefinedsetv6"
+	customImportRejectSet   = "customimportrejectdefinedset"
+	customImportRejectSetV6 = "customimportrejectdefinedsetv6"
+
+	defaultRouteSet   = "defaultroutedefinedset"
+	defaultRouteSetV6 = "defaultroutedefinedsetv6"
 
 	kubeRouterExportPolicy = "kube_router_export"
 	kubeRouterImportPolicy = "kube_router_import"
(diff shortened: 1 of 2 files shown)


aauren:
@greptileai take a second pass

PR #2107, opened 2026-07-01
Remove dead error checks

twz123:
#### What type of PR is this?

cleanup

The `newGlobalPeers` function is now infallible, but the err checks weren't removed. Enable the govet linter as a safeguard for this.

No.

None.

```release-note
NONE
```


aauren:
Good catch! Thanks!

aauren (approved):

PR #2118, opened 2026-07-05
Implement BFD support

### LINKED ISSUE #1117: BFD support in BGP
Kube-router BGP support is great, but with support of BFD BGP feaature make KubeProxy more Great!

the support of BFD while optimize BGP time convergence, without it the router need to wait hold time timer (generaly around 120s) to black list the route to dead target ip... with BFD can be below seconds....

catherinetcai:
#### What type of PR is this?

feature

Implement basic BFD support into kube-router for external BGP peers

Fixes https://github.com/cloudnativelabs/kube-router/issues/1117

* What tool was used:  Claude Code (with GLM 5.2 & Deepseek V4 Flash)
* To what extent was the tool used?  Used it to author a quick first draft so I could get to testing. Don't have a ton of exposure to BFD either, so I used it as a jumping off point for research into understanding how it's implemented and the various config options since the various docs I was reading from hardware providers/networking guides weren't clear about how it behaves. (Was mostly trying to figure out how it behaves if graceful restart is also enabled.) I also used it during testing for `vtysh` commands for FRR (since I'm not familiar with it) and to construct the `iptables` rule to drop BFD traffic. 
* If drafted, how detailed of a plan did you create for the AI?  Pretty detailed about what I wanted the configs to look like and the actual entrypoints in the config to write.

I did some testing prior to cleanliness refactors and I'll do additional rounds of integration tests after feedback. 

Tested it a few times with kube-router-automation, which spins up a controller, a worker, and an external BGP peer (with FRR). I updated the FRR configs with BFD configs. I also added routes on the FRR node that were exported to kube-router.

To simulate failure, I tried a few combos: 

* `systemctl kill frr` to kill FRR after peering
* `iptables -A INPUT -p udp --dport 3784 -s <FRR_IP> -j DROP` - In the kube-router pods to drop external BFD traffic to see what happened

To validate behavior, I would exec into the kube-router pod and run `watch -n 1 ip route show` and `watch -n 1 gobgp global rib` to see how quickly routes got yanked & how quickly GoBGP would mark a peer state from Established -> Idle -> Active. 

I would then restore the peer session by either doing `systemctl start frr` or dropping the iptables rule I added 

I then tested behavior with a few scenarios:

**BFD + GR both enabled in external peer + kube-router nodes** 
As expected, BFD seems to supersede GR. 

**Node Annotation Implemention/Asymmetric Setup Per Node**

The other test cases were done with the cluster wide `--enable-bfd` flag.

I tested the per-node annotations as well. I set up the controller with the legacy single BGP config annotations (which should not get any BFD configs because the legacy annotations do not support BFD as they're deprecated) and the worker with the consolidated node BGP annotation (with the BFD config). 

I verified on the FRR side that the controller was a BGP peer, but BFD was not configured and that the worker was a BGP peer with BFD configured. 

Sorry for the novel - I also took more detailed notes with test configs that I'm happy to share. 

```release-note
NONE
```

There a few implementation choices I made that I'd like feedback on:

* kube-router exposes the timer intervals as microseconds, because that's what GoBGP uses. However, FRR uses milliseconds - and after doing the integration testing, I think the milliseconds are a lot easier to deal with. Wanted to get feedback on this
* The cluster-wide BFD flags all only take 1 set of configs for the global external peers for ease of implementation. However, I can see how it might be desirable to provide the flexibility to configure separate values per peer. I'm totally open to changing it. 
* The single BGP annotations are deprecated, so I didn't feel like we should support new features with them to continue encouraging their use. I'm okay with changing the implementation to support them though, it's not very difficult to.

### DIFF
--- a/docs/bgp.md
+++ b/docs/bgp.md
@@ -24,6 +24,9 @@ pod IPs, service IPs, etc.).
   - [Restoring pre-2.9.0 behavior](#restoring-pre-290-behavior)
 - [Overriding the next hop](#overriding-the-next-hop)
 - [Overriding the next hop and enable IPIP/tunnel](#overriding-the-next-hop-and-enable-ipiptunnel)
+- [BFD (Bidirectional Forwarding Detection)](#bfd-bidirectional-forwarding-detection)
+  - [Configuring BFD](#configuring-bfd)
+  - [BFD & Graceful Restart](#bfd--graceful-restart)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -358,3 +361,57 @@ Specifically, people need to take care when combining `--override-nexthop` and `
 they understand their network, the flows they desire, how the kube-router logic works, and the possible side effects
 that are created from their configuration. Please refer to [this PR](https://github.com/cloudnativelabs/kube-router/pull/1025)
 for the risk and impact discussion.
+
+## BFD (Bidirectional Forwarding Detection)
+
+kube-router integrates GoBGP's support for BFD for external peers. Please review GoBGP's
+[docs on BFD](https://github.com/osrg/gobgp/blob/master/docs/sources/bfd.md)
+for details on features and limitations, as GoBGP's implementation for BFD is new,
+and does not support all features.
+
+### Configuring BFD
+
+You can configure BFD settings either through CLI flags for global external BGP peers
+or configure them on a per-node basis. Ensure that detection multiplier,
+required min RX interval, and desired min TX interval to either match or
+are slower than the settings configured on the peer, as GoBGP does not currently
+handle timer negotiation. Note that required min RX interval and desired min TX
+interval are both in _microseconds_, as some implementations of BFD (such as FRR)
+handle timer values in milliseconds.
+
+**Global External BGP Peers:**
+
+You can enable BFD for global external BGP peers by specifying: `--enable-bfd=true`.
+BFD timer settings can be configured via `--bfd-detection-multiplier`,
+`--bfd-required-min-rx-interval`, and `--bfd-desired-min-tx-interval`. Each of these
+flags only accept one value to configure BFD settings for all peers. If different
+BFD configurations are required for external peers, please use the per-node BGP
+annotations.
+
+**Per-Node BFD Configuration:**
+
+Per-node BFD settings can be configured via the `kube-router.io/peers` annotation:
+
+```yaml
+- remoteip: 192.168.1.99
+  remoteasn: 65000
+  password: "<b64 encoded password here>"
+  bfd:
+    enabled: true
+    port: 3785
+    detection_multiplier: 2
+    required_min_rx_interval: 1000000
+    desired_min_tx_interval: 1000000
+```
+
+Per-node BFD settings are _not_ supported via the individual BGP node annotations,
+which have been deprecated since v2.7.0.
+
+### BFD & Graceful Restart
+
+kube-router will log warnings when BFD and graceful restart are both configured.
+Configuring both BFD and graceful restart is not recommended as they have contradictory
+effects and can cause inconsistent routing behavior depending on vendor implementation.
+In GoBGP, if BFD and graceful restart are both configured and a peer is
+detected as down, graceful restart seems to be bypassed and peer routes are
+immediately flushed.
--- a/docs/bgp.md
+++ b/docs/bgp.md
@@ -376,8 +376,7 @@ or configure them on a per-node basis. Ensure that detection multiplier,
 required min RX interval, and desired min TX interval to either match or
 are slower than the settings configured on the peer, as GoBGP does not currently
 handle timer negotiation. Note that required min RX interval and desired min TX
-interval are both in _microseconds_, as some implementations of BFD (such as FRR)
-handle timer values in milliseconds.
+interval are both in _milliseconds_, instead of microseconds, which GoBGP uses.
 
 **Global External BGP Peers:**
 
@@ -400,8 +399,8 @@ Per-node BFD settings can be configured via the `kube-router.io/peers` annotatio
     enabled: true
     port: 3785
     detection_multiplier: 2
-    required_min_rx_interval: 1000000
-    desired_min_tx_interval: 1000000
+    required_min_rx_interval: 1000
+    desired_min_tx_interval: 1000
 ```
 
 Per-node BFD settings are _not_ supported via the individual BGP node annotations,
--- a/pkg/bgp/bfd.go
+++ b/pkg/bgp/bfd.go
@@ -1,0 +1,52 @@
+package bgp
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/cloudnativelabs/kube-router/v2/pkg/options"
+	gobgpapi "github.com/osrg/gobgp/v4/api"
+	"k8s.io/utils/ptr"
+)
+
+type BFDConfig struct {
+	Enabled bool    `yaml:"enabled"`
+	Port    *uint32 `yaml:"port"`
+
+	DesiredMinTxInterval  *uint32 `yaml:"desired_min_tx_interval"`
+	DetectionMultiplier   *uint32 `yaml:"detection_multiplier"`
+	RequiredMinRxInterval *uint32 `yaml:"required_min_rx_interval"`
+}
+
+func (b BFDConfig) String() string {
+	fields := []string{fmt.Sprintf("Enabled: %t", b.Enabled)}
+	if b.Port != nil {
+		fields = append(fields, fmt.Sprintf("Port: %d", *b.Port))
+	}
+	if b.DesiredMinTxInterval != nil {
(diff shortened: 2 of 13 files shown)


pkg/controllers/routing/network_routes_controller.go
```diff
@@ -1150,6 +1155,14 @@ func (nrc *NetworkRoutingController) startBgpServer(grpcServer bool) error {
 			nrc.bgpHoldtime,
 			nrc.krNode.GetPrimaryNodeIP().String(),
 		)
+		if err != nil {
+			err2 := nrc.bgpServer.StopBgp(context.Background(), &gobgpapi.StopBgpRequest{})
+			if err2 != nil {
+				klog.Errorf("Failed to stop bgpServer: %s", err2)
```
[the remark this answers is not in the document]
catherinetcai: Looks like some of this came in with a bad rebase. Removing.

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -1150,6 +1155,14 @@ func (nrc *NetworkRoutingController) startBgpServer(grpcServer bool) error {
...
+			}
+
+			return fmt.Errorf("failed to process Global Peer Router configs: %w", err)
+		}
```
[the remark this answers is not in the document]
catherinetcai: This got pulled in as a result of a bad rebase, cleaning up.

pkg/bgp/bfd_test.go
```diff
@@ -0,0 +1,129 @@
...
+			name: "Enabled",
+			config: BFDConfig{
+				Enabled: true,
+			},
+			expected: "BFDConfig{Enabled: true}",
+		},
+		{
+			name: "Port",
+			config: BFDConfig{
+				Port: new(uint32(3784)),
```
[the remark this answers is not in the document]
catherinetcai: This is outdated guidance. As of Go 1.26.2, new now takes a type or expression and returns a pointer. https://pkg.go.dev/builtin@go1.26.2#new

pkg/controllers/routing/network_routes_controller_test.go
```diff
@@ -2599,8 +2599,43 @@ func Test_bgpPeerConfigsFromAnnotations(t *testing.T) {
...
+				peersAnnotation: `- remoteip: 10.0.0.1
+  remoteasn: 64640
+  password: cGFzc3dvcmQ=
+  localip: 192.168.0.1
+  bfd:
+    port: 37850
+    detection-multiplier: 2
+    desired-min-tx-interval: 2
+    required-min-rx-interval: 2
+- remoteip: 10.0.0.2
```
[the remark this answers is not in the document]
catherinetcai: This was a valid find. The test has been refactored. They were originally written when PeerConfigs exported its fields. PeerConfigs was refactored to no longer export any of its fields and the tests were no longer testing expectations due to the `cmp` setting to ignore unexported fields. Oops.

docs/bgp.md
```diff
@@ -358,3 +361,57 @@ Specifically, people need to take care when combining `--override-nexthop` and `
...
+annotations.
+
+**Per-Node BFD Configuration:**
+
+Per-node BFD settings can be configured via the `kube-router.io/peers` annotation:
+
+```yaml
+- remoteip: 192.168.1.99
+  remoteasn: 65000
+  password: U2VjdXJlUGFzc3dvcmQK
```
mrueg: I know this is an example only, can we replace the value here so scanners don't accidentally pick it up?

catherinetcai: Done!

docs/bgp.md
```diff
@@ -358,3 +361,57 @@ Specifically, people need to take care when combining `--override-nexthop` and `
...
+
+### Configuring BFD
+
+You can configure BFD settings either through CLI flags for global external BGP peers
+or configure them on a per-node basis. Ensure that detection multiplier,
+required min RX interval, and desired min TX interval to either match or
+are slower than the settings configured on the peer, as GoBGP does not currently
+handle timer negotiation. Note that required min RX interval and desired min TX
+interval are both in _microseconds_, as some implementations of BFD (such as FRR)
+handle timer values in milliseconds.
```
aauren: suggestion(blocking): You were asking for feedback on this in your initial PR description, and I think that your gut on this is 100% correct. I think that milliseconds is a lot easier for people to reason about than microseconds (even if that's what the RFC is in).

To give some additional validation here:

* Cisco IOS - [milliseconds](https://www.cisco.com/c/en/us/td/docs/ios/12_0s/feature/guide/fs_bfd.html)
* Juniper JunOS - [milliseconds](https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/statement/minimum-interval-edit-protocols-bgp-liveness-detection.html)
* BIRD - [time flexible, but milliseconds recommended](https://bird.network.cz/doc/bird-6.html#ss6.2)

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -1369,6 +1389,18 @@ func NewNetworkRoutingController(clientset kubernetes.Interface,
 		klog.Warning("--gobgp-admin-port is 0 but --gobgp-admin-address is non-empty, not binding GoBGP socket to an address")
 	}
 
+	nrc.enableBFD = kubeRouterConfig.EnableBFD
+	nrc.bfdPort = kubeRouterConfig.BFDPort
+	nrc.bfdDetectionMultiplier = kubeRouterConfig.BFDDetectionMultiplier
+	nrc.bfdMinRxInt = kubeRouterConfig.BFDRequiredMinRxInterval
+	nrc.bfdMinTxInt = kubeRouterConfig.BFDDesiredMinTxInterval
```
aauren: question: any reason not just to move this into a `BFDConfig` struct right off the bat and just store that rather than storing individual fields on the controller?

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -1369,6 +1389,18 @@ func NewNetworkRoutingController(clientset kubernetes.Interface,
...
+
+	if nrc.enableBFD {
+		if nrc.bfdDetectionMultiplier == 0 || nrc.bfdDetectionMultiplier > options.BFDDetectionMultiplierMax {
```
aauren: suggestion(non-blocking): I would move all of this into a `Validate()` method on the `BFDConfig` struct so that:

* Validation could be tested in `bfd_test.go` holistically
* So that the NRC doesn't have to know as much about BFD options
* So that Peer annotations could be validated in the same way as the command options here

If you went that route, I would also move the maxes into bfd.go so that all of the info about BFD was contained in `bfd.go`.

aauren: issue(blocking): I think that we should also validate that BFD Port is within a valid range as well (i.e. non-zero and less than 65,535

aauren (changes requested):
@catherinetcai - Great work!!!

I tested this as well in my own environment and everything worked flawlessly! All of my comments here are about code-structure / defaults. Thanks so much for your work on this!

Re your question about updating the legacy annotations or not, I think you made the right call, we should not be bringing new functionality into the legacy annotations.

One other thing, that might be worth documenting, open to feedback here, is that in my testing I found that BGP can become re-established after BFD fails before BFD is re-established. If BGP fails during that window (about 30 seconds) the routes will not be revoked.

From what I can tell, this is compliant with the RFC for BFD and is similar to what other default BFD implementations do, but it might be worth putting in the documents? Maybe not, as its a bit of an edge case? I'll leave it up to you.

docs/bgp.md
catherinetcai: Updated to use milliseconds instead of microseconds!

pkg/controllers/routing/network_routes_controller.go
catherinetcai: Nope, probably better to do it this way. I think I originally structured it like this for max flexibility, but that just causes sprawl.

pkg/controllers/routing/network_routes_controller.go
catherinetcai: Good callout re: validation. Updated all of it into `bfd.go` and added tests there!

pkg/bgp/bfd.go
```diff
@@ -0,0 +1,52 @@
...
+		return nil
+	}
+
+	return &gobgpapi.BfdPeerConfig{
+		Enabled:                  true,
+		Port:                     ptr.Deref(b.Port, options.DefaultBFDPort),
+		DetectionMultiplier:      ptr.Deref(b.DetectionMultiplier, options.DefaultBFDDetectionMultiplier),
+		DesiredMinimumTxInterval: ptr.Deref(b.DesiredMinTxInterval, options.DefaultBFDDesiredMinTxInterval) * 1000,
+		RequiredMinimumReceive:   ptr.Deref(b.RequiredMinRxInterval, options.DefaultBFDRequiredMinRxInterval) * 1000,
+	}
```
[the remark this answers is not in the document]
catherinetcai: BFDConfig is already validated via both paths for its construction to check for these values. Adding another check here is unnecessary.

aauren (approved):
Looks great! Thanks Cat!

Issue #59, opened 2017-07-13
ERROR: logging before flag.Parse


bzub:
With glog updates we're running into this bug, https://github.com/kubernetes/kubernetes/issues/17162

There's a workaround in there that I'll try out.

PR #2125, opened 2026-07-28
Misc Touch Ups and Fixes

aauren:
#### What type of PR is this?

bug
cleanup
feature
regression

This PR does a few things:

* Fixes a regression in the e2e-netpol action flow that I accidentally introduced
* kube-router now fails if people accidentally introduce positional args instead of silently swallowing them
* Refactors `setupKubeBridge()` to be less complex and fixes an IPv6 comment
* Replaces string error checking with `errors.AsType` available since 1.25 for netlink error type checking also add more complete error handling
* Add proper context handling to `NRC.Run()` to `setupKubeBridge()` to ensure that it doesn't get stuck during termination
* Add dual-stack handling for `NodePort` binding (similar to what was done for ClusterIP) to ensure that even if the primary interface is one family or another, that in dual-stack handling it at least binds to both families. This was causing several conformance tests to fail previously and had never been properly handled before
  * Also adds tests to ensure regressions don't happen in the future
* Add a structured way of passing k8s informers into the various kube-router controllers
  * This reduces the complexity of controller signatures and moves handler adding into the controllers which is a bit more idiomatic and self-contained

The dual-stack NodePort handling fixes the following k8s conformance regressions:

```
  [FAIL] [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] [It] should function for endpoint-Service: http [sig-network, Feature:IPv6DualStack]
  [FAIL] [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] [It] should function for endpoint-Service: udp [sig-network, Feature:IPv6DualStack]
  [FAIL] [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] [It] should function for node-Service: http [sig-network, Feature:IPv6DualStack]
  [FAIL] [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] [It] should function for node-Service: udp [sig-network, Feature:IPv6DualStack]
  [FAIL] [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] [It] should function for pod-Service: http [sig-network, Feature:IPv6DualStack]
  [FAIL] [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] [It] should function for pod-Service: sctp [Feature:SCTPConnectivity] [sig-network, Feature:IPv6DualStack, Feature:SCTPConnectivity]
  [FAIL] [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] [It] should function for pod-Service: udp [sig-network, Feature:IPv6DualStack]
  [FAIL] [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] [It] should function for service endpoints using hostNetwork [sig-network, Feature:IPv6DualStack]
  [FAIL] [sig-network] Services [It] should be able to create a functioning NodePort service [Conformance] [sig-network, Conformance]
```

None

* What tool was used: Claude
* To what extent was the tool used? Most coding, with supervision / review by human
* If drafted, how detailed of a plan did you create for the AI? Very Detailed

Fully k8s conformance testing, this resolves 9 conformance test failures. Plus it was deployed to a working k8s cluster.

```release-note
Replacement of individual structured k8s informers does introduce pkg instability in the current v2 version for downstreams
```

This is a semi-large PR with a lot of clubbed changes, I would recommend reviewing it commit-by-commit as it is easier to understand that way.

### DIFF
--- a/cmd/kube-router/kube-router.go
+++ b/cmd/kube-router/kube-router.go
@@ -48,6 +48,10 @@ func Main() error {
 		return fmt.Errorf("failed to set flag: %w", err)
 	}
 
+	if args := pflag.Args(); len(args) > 0 {
+		return fmt.Errorf("unrecognized positional argument(s): %v", args)
+	}
+
 	if config.HelpRequested {
 		pflag.Usage()
 		return nil
--- a/cmd/kube-router/kube-router.go
+++ b/cmd/kube-router/kube-router.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"errors"
-	"flag"
 	"fmt"
 	"net/http"
 	"os"
@@ -27,25 +26,15 @@ func main() {
 }
 
 func Main() error {
-	klog.InitFlags(nil)
-
 	config := options.NewKubeRouterConfig()
 	config.AddFlags(pflag.CommandLine)
 	pflag.Parse()
 
-	// Workaround for this issue:
-	// https://github.com/kubernetes/kubernetes/issues/17162
-	err := flag.CommandLine.Parse([]string{})
-	if err != nil {
-		return fmt.Errorf("failed to parse flags: %w", err)
-	}
-	err = flag.Set("logtostderr", "true")
-	if err != nil {
-		return fmt.Errorf("failed to set flag: %w", err)
-	}
-	err = flag.Set("v", config.VLevel)
-	if err != nil {
-		return fmt.Errorf("failed to set flag: %w", err)
+	// Level.Set() mutates klog's package-global verbosity rather than the receiver, so we can discard
+	// vLevel afterwards. Everything else we need, logging to stderr included, is already klog's default.
+	var vLevel klog.Level
+	if err := vLevel.Set(config.VLevel); err != nil {
+		return fmt.Errorf("failed to set log level to %q: %w", config.VLevel, err)
 	}
 
 	if args := pflag.Args(); len(args) > 0 {
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -41,7 +41,6 @@ const (
 	KubeTunnelIfv4    = "kube-tunnel-if"
 	KubeTunnelIfv6    = "kube-tunnel-v6"
 	KubeBridgeIf      = "kube-bridge"
-	IfaceNotFound     = "Link not found"
 	IfaceHasNoAddr    = "cannot assign requested address"
 	IpvsMaglevHashing = "mh"
 	IpvsSvcFSched1    = "flag-1"
@@ -1850,13 +1849,13 @@ func (nsc *NetworkServicesController) Cleanup() {
 	// delete dummy interface used to assign cluster IP's
 	dummyVipInterface, err := nlretry.LinkByName(context.Background(), KubeDummyIf)
 	if err != nil {
-		if err.Error() != IfaceNotFound {
-			klog.Infof("Dummy interface: " + KubeDummyIf + " does not exist")
+		if _, notFound := errors.AsType[netlink.LinkNotFoundError](err); notFound {
+			klog.Infof("Dummy interface: %s does not exist", KubeDummyIf)
 		}
 	} else {
 		err = netlink.LinkDel(dummyVipInterface)
 		if err != nil {
-			klog.Errorf("could not delete dummy interface %s due to: %v", KubeDummyIf, err.Error())
+			klog.Errorf("could not delete dummy interface %s due to: %v", KubeDummyIf, err)
 			return
 		}
 	}
--- a/pkg/controllers/proxy/network_services_controller.go
+++ b/pkg/controllers/proxy/network_services_controller.go
@@ -165,6 +165,14 @@ type NetworkServicesController struct {
 	nphc *nodePortHealthCheckController
 }
 
+// Informers is the set of shared informers the NSC reads from. We declare it here rather than accepting a
+// full informer factory so that the controller can only reach the resources it actually watches.
+type Informers interface {
+	Services() cache.SharedIndexInformer
+	EndpointSlices() cache.SharedIndexInformer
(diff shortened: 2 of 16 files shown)


pkg/controllers/proxy/network_services_controller.go
```diff
@@ -1850,13 +1857,13 @@ func (nsc *NetworkServicesController) Cleanup() {
 	// delete dummy interface used to assign cluster IP's
 	dummyVipInterface, err := nlretry.LinkByName(context.Background(), KubeDummyIf)
 	if err != nil {
-		if err.Error() != IfaceNotFound {
-			klog.Infof("Dummy interface: " + KubeDummyIf + " does not exist")
+		if _, notFound := errors.AsType[netlink.LinkNotFoundError](err); notFound {
+			klog.Infof("Dummy interface: %s does not exist", KubeDummyIf)
 		}
 	} else {
```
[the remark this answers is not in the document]
aauren: This is an error, but it is one of multiple in the `--cleanup` logic. I'm going to keep this PR smaller and address the myriad problems with `--cleanup` some other time in the future. For now, it is not a critical part of the core functionality.

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -179,6 +184,13 @@ func (nrc *NetworkRoutingController) Run(
 	stopCh <-chan struct{},
 	wg *sync.WaitGroup,
 ) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go func() {
+		<-stopCh
```
twz123: You probably want to have something like this here:

```go
select {
case <-stopCh:
	cancel()
case <-ctx.Done():
}
```

...so that this goroutine doesn't outlive the context it's supposed to cancel.

Didn't have a look, but the question is if it'd be more meaningful to convert `stopCh` into a context and handle this at the call site(s)?

pkg/controllers/routing/network_routes_controller.go
```diff
@@ -39,9 +39,6 @@ import (
 )
 
 const (
-	// Deprecated: Use [netlink.LinkNotFoundError] instead.
-	IfaceNotFound = "Link not found"
```
twz123, line 43: Just a quick double check: kube-router doesn't treat its exported Go API as stable? Not that I'd object, but I've always acted on the safe side so far, and didn't want to smuggle in breaking changes.

cmd/kube-router/kube-router.go
```diff
@@ -48,6 +48,10 @@ func Main() error {
 		return fmt.Errorf("failed to set flag: %w", err)
 	}
 
+	if args := pflag.Args(); len(args) > 0 {
+		return fmt.Errorf("unrecognized positional arguments (did you forget '--'?): %v", args)
```
twz123: Had to read this a couple of times until I understood what this means, because `--` is usually the GNU style marker for "end of flags, everything past this point is positional arguments". But that's completely misleading in how it's meant here: no positional args supported, did you forget to prepend a double-dash in front of your flags?

I'd probably just drop the second part from the error message, or at least print something like `"--"+args[0]`. But then again, this would be strange if kube-router doesn't have such a flag to begin with. Also, I remember having read some style guides mentioning that including questions in error messages is bad practice. But yeah...

pkg/controllers/routing/network_routes_controller.go
aauren: Good call, that would make this a bit more robust to other return conditions that might come up in the future.

I think it would probably be good to lean further into contexts in the future. Probably a bit further than I want to go in this PR, but good to think about for a future refactor.

pkg/controllers/routing/network_routes_controller.go
aauren: Yeah, I would say that we try to take a balanced approach. In general we try not to break compatibility, but if it makes sense, then I think we're open to it.

In general, we don't cut a new major version because those are painful across the board. This is a similar approach to the one that GoBGP has taken.

aauren: As a downstream receiver of kube-router, maybe it would be good to get your feedback on this approach. Do you have any thoughts on this approach?

cmd/kube-router/kube-router.go
aauren: Fair enough, I'll admit that this was a bit of AI slop and it was based on a real condition that I hit while testing, so AI over-indexed on my experience here. And you're right about the style comment as well.

Thanks!

pkg/controllers/routing/network_routes_controller.go
twz123: We're doing basically the same for k0s, as it's not meant and designed to be consumed as a Go dependency. With the the Go knowledge that I have today, I'd love to stuff 90% of the code base into `internal` directories, to make this obvious. But well, that ship has sailed, for the most part. At least I try to do it for newly added stuff.

That said, k0s itself doesn't import kube-router as a Go dependency either, so we can't be affected anyways 🤷

cmd/kube-router/kube-router.go
catherinetcai: I know this is outside of your PR, but figured I'd ask since your PR touches the entrypoint. Is there any reason why we're using both flag and pflag? Should we just consolidate on one?

catherinetcai (approved):

cmd/kube-router/kube-router.go
aauren: Hmm... It predated my time with the project, but going back through the git history, it looks like it was related to: https://github.com/kubernetes/kubernetes/issues/17162 which is an old glog issue from 2016 and #59 ?

Its a good reminder and there is no time like the present to remove stale code, so I have removed it.

PR #2126, opened 2026-07-30
Support bootstrapping CNI configs from a template file

twz123:
#### What type of PR is this?

feature

A CNI configuration file that's copied into the CNI configuration directory by an init container is already visible to the container runtime before kube-router has completed it, so on a freshly booted node, pod sandbox creation may transiently fail with errors like "no IP ranges specified" until kube-router has filled in the missing parts.

To avoid this, introduce CNI configuration bootstrapping via a template file. If the `KUBE_ROUTER_CNI_CONF_TEMPLATE_FILE` environment variable is not empty, kube-router reads the config from the template instead, and writes the completed config to the usual CNI config file path, which doesn't need to exist beforehand. The template file's content is interpreted according to the CNI config file name. The template's own file name has no significance.

As a result, the container runtime won't pick up an incomplete configuration and try to set up pod networking too early. Since the config is re-derived from the pristine template on every start, the written config is wholly owned by kube-router.

Write this up in the user guide and add another example DaemonSet.

The PR can be reviewed commit by commit:

* Quite some test refactoring commits,
* one fix that hardens against panics for some more pathological CNI config inputs (there's still a gap around ipam fields that are null, IIRC),
* and the final feature commit.

The problem itself _should_ dissolve as soon as the correct CNI config landed on disk. However, this is very reliably triggering 
containernetworking/plugins#1265, especially in air-gapped scenarios when there's no images to be pulled and the container runtime can start pods right away.

Yes, for design discussions, patch series review, and the initial draft of the changes to the user guide.

Unit tests so far. I plan to give this a spin in the k0s CI, of course.

No.

```release-note
NONE
```

I deliberately didn't try to do some write-temprary-file-then-rename dance to make the config file appear atomically. The time it takes to write the file is negligible, probably microseconds in practice. Getting this atomic thing right, along with all its edge cases, warrants its own library. (And, no, I don't think renameio gets this right, either.)

### DIFF
--- a/daemonset/generic-kuberouter-cni-conf-template.yaml
+++ b/daemonset/generic-kuberouter-cni-conf-template.yaml
@@ -1,0 +1,217 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kube-router-cfg
+  namespace: kube-system
+  labels:
+    tier: node
+    k8s-app: kube-router
+data:
+  cni-conf.json: |
+    {
+      "cniVersion": "0.3.0",
+      "name": "mynet",
+      "plugins": [
+        {
+          "name": "kubernetes",
+          "type": "bridge",
+          "bridge": "kube-bridge",
+          "isDefaultGateway": true,
+          "ipam": {
+            "type": "host-local"
+          }
+        }
+      ]
+    }
+
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    k8s-app: kube-router
+    tier: node
+  name: kube-router
+  namespace: kube-system
+spec:
+  selector:
+    matchLabels:
+      k8s-app: kube-router
+      tier: node
+  template:
+    metadata:
+      labels:
+        k8s-app: kube-router
+        tier: node
+    spec:
+      priorityClassName: system-node-critical
+      serviceAccountName: kube-router
+      initContainers:
+      - name: install-cni
+        image: docker.io/cloudnativelabs/kube-router
+        imagePullPolicy: Always
+        command:
+        - /bin/sh
+        - -ec
+        - |
+          [ -f /etc/cni/net.d/10-kuberouter.conflist ] || {
+            rm -f /etc/cni/net.d/*.conf
+            rm -f /etc/cni/net.d/*.conflist
+          }
+          /usr/local/bin/cni-install
+        volumeMounts:
+        - name: cni-conf-dir
+          mountPath: /etc/cni/net.d
+        - name: host-opt
+          mountPath: /opt
+      containers:
+      - name: kube-router
+        image: docker.io/cloudnativelabs/kube-router
+        imagePullPolicy: Always
+        args:
+        - --run-router=true
+        - --run-firewall=true
+        - --run-service-proxy=false
+        - --bgp-graceful-restart=true
+        env:
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: KUBE_ROUTER_CNI_CONF_FILE
+          value: /etc/cni/net.d/10-kuberouter.conflist
+        - name: KUBE_ROUTER_CNI_CONF_TEMPLATE_FILE
+          value: /etc/kube-router/cni-conf.json
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 20244
+          initialDelaySeconds: 10
+          periodSeconds: 3
+        resources:
+          requests:
+            cpu: 250m
+            memory: 250Mi
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - name: lib-modules
+          mountPath: /lib/modules
+          readOnly: true
+        - name: cni-conf-dir
+          mountPath: /etc/cni/net.d
+        - name: kube-router-cfg
+          mountPath: /etc/kube-router
+          readOnly: true
+        - name: xtables-lock
+          mountPath: /run/xtables.lock
+          readOnly: false
+      hostNetwork: true
+      hostPID: true
+      tolerations:
+      - effect: NoSchedule
+        operator: Exists
+      - key: CriticalAddonsOnly
+        operator: Exists
+      - effect: NoExecute
+        operator: Exists
+      volumes:
+      - name: lib-modules
+        hostPath:
+          path: /lib/modules
+      - name: cni-conf-dir
+        hostPath:
+          path: /etc/cni/net.d
+      - name: kube-router-cfg
+        configMap:
+          name: kube-router-cfg
+      - name: xtables-lock
+        hostPath:
+          path: /run/xtables.lock
+          type: FileOrCreate
+      - name: host-opt
+        hostPath:
+          path: /opt
+
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: kube-router
+  namespace: kube-system
+
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: kube-router
+  namespace: kube-system
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  - pods
+  - services
+  - nodes
+  - endpoints
+  verbs:
+  - list
+  - get
+  - watch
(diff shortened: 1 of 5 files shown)


docs/user-guide.md
```diff
@@ -578,46 +580,92 @@ $ kubectl annotate service my-service "kube-router.io/service.schedflags=flag-2"
...
-- Update init container command to create `/etc/cni/net.d/10-kuberouter.conflist` file
+## CNI Configuration
+
+kube-router provides pod networking by means of the standard [`bridge`][cni-bridge] and [`host-local`][cni-host-local]
+[containernetworking plugins], optionally accompanied by others like [`portmap`][cni-portmap]. The kube-router Docker
+image ships these plugins, and the [example daemonsets](../daemonset/) install them onto the node in an init container.
+kube-router's own contribution is the routing of pod traffic between nodes and the management of the plugins'
+configuration file, filling in the parts that are node-specific and only known at runtime. The latter is what
+`--enable-cni` controls, which is enabled by default. Disable it to use kube-router's features alongside another CNI
+provider. On startup, kube-router
```
catherinetcai: I think "On startup..." is a good spot to place a line break for readability.

I think adding a colon at the end helps visually for formatting with the bullet points right below.

docs/user-guide.md
```diff
@@ -578,46 +580,92 @@ $ kubectl annotate service my-service "kube-router.io/service.schedflags=flag-2"
...
+
+- inserts the node's pod CIDRs into the `host-local` IPAM ranges of the `bridge` plugin, as allocated to the node by
+  the kube-controller-manager, or as set via the `kube-router.io/pod-cidrs` annotation
+- and sets the `bridge` plugin's MTU, if `--auto-mtu` is enabled (see [MTU](#mtu)).
```
catherinetcai: Nit: I would drop the starting "and" for this second point and drop the ending period since this is a bulleted list and not meant to be read like complete sentences.

pkg/utils/cni_test.go
```diff
@@ -197,404 +233,285 @@ func getConf() []byte {
...
+			ranges:  []string{"10.242.0.0/24", "10.242.1.0/24", "10.242.2.0/24", "10.242.3.0/24"},
 		},
 		{
-			name:       "Ensure conflist subnets get de-deduplicated with ranges when repeats exist",
-			filename:   "10-kuberouter.conflist",
-			isConfList: true,
-			content:    getConfListWithIPv6DuplicateRanges(),
-			err:        nil,
-			ranges:     []string{"10.242.0.0/24", "10.242.1.0/24", "10.242.2.0/24", "2001:db8:42:2::/64"},
+			name:    "Ensure conflist subnets get de-deduplicated with ranges when repeats exist",
```
catherinetcai: I think this test name is missing calling out IPv6

pkg/utils/cni_test.go
```diff
@@ -197,404 +233,285 @@ func getConf() []byte {
...
+			content: getConfListWithNoPlugins(),
+			err:     "10-kuberouter.conflist has no bridge plugin",
 		},
 		{
-			name:       "Ensure error upon reading from conf with no type",
-			filename:   "10-kuberouter.conf",
-			isConfList: false,
-			content:    getConfWithNoType(),
-			err:        errors.New("error load CNI config, file appears to have no type: "),
+			name:    "Ensure conf subnet get consolidated into ranges when only subnet exists",
```
catherinetcai, line 329: I think the args for this test are the same as  the "Attempt reading from conf" test:

```golang
{
			name:    "Attempt reading from conf",
			content: getConf(),
			ranges:  []string{"10.242.0.0/24"},
},
```

pkg/utils/cni_test.go
```diff
@@ -197,404 +233,285 @@ func getConf() []byte {
...
+			content: getConf(),
+			ranges:  []string{"10.242.0.0/24"},
 		},
 		{
-			name:       "Ensure error upon reading from conflist with no plugins",
-			filename:   "10-kuberouter.conflist",
-			isConfList: true,
-			content:    getConfListWithNoPlugins(),
-			err:        errors.New("CNI config list "),
+			name:    "Ensure conflist subnet get consolidated into ranges when only subnet exists",
```
catherinetcai, line 334: Seems to be the same args as a previous test:

```golang
{
			name:    "Attempt reading from conflist",
			content: getConfList(),
			ranges:  []string{"10.242.0.0/24"},
},
```

pkg/utils/cni.go
```diff
@@ -16,35 +16,50 @@ type CNINetworkConfig struct {
 }
 
 func NewCNINetworkConfig(cniConfFilePath string) (*CNINetworkConfig, error) {
+	return NewCNINetworkConfigFromTemplate(cniConfFilePath, "")
```
catherinetcai: Is this calling down to NewCNINtworkConfigFromTemplate just to try and not break the API? I think it's okay to just update the API for NewCNINetworkConfig to accept `cniConfTemplateFilePath`. 

I think this entire utils package really belongs in internal (and we should do that refactor at some point...)

catherinetcai (commented):
Great work @twz123 - just some minor, surface level comments. 

What do you think about having this templating behavior be the user-facing default moving forward? I'm not sure if I see the benefit of continuing to push the legacy/racy behavior, so I think it's worth just updating all of the examples under the daemonsets folder to use the template instead and calling it out in the release notes.

aauren:
Ping @twz123 - I think we're good to merge this as soon as the comments @catherinetcai left are addressed.

docs/user-guide.md
twz123: Done

docs/user-guide.md
twz123: I reformulated this into two full sentences. I feel this reads better than the bullet points. WDYT?

pkg/utils/cni_test.go
twz123: Deduped in 1b652332a18cca295b097f86c1bb1b5fd02be553.

pkg/utils/cni_test.go
twz123: Deduped in 1b652332a18cca295b097f86c1bb1b5fd02be553.

pkg/utils/cni_test.go
twz123: After you've pointed it out, I've checked the test cases and it turned out there were quite some dupes, which the refactoring in this PR made clear. So I've removed the dupes and consolidated the tests in 1b652332a18cca295b097f86c1bb1b5fd02be553. There's indeed a separate test case for IPv6. Is that enough?

pkg/utils/cni.go
twz123: > Is this calling down to NewCNINtworkConfigFromTemplate just to try and not break the API?

Yep.

> I think it's okay to just update the API for NewCNINetworkConfig to accept `cniConfTemplateFilePath`.

Done.

twz123:
> What do you think about having this templating behavior be the user-facing default moving forward? I'm not sure if I see the benefit of continuing to push the legacy/racy behavior, so I think it's worth just updating all of the examples under the daemonsets folder to use the template instead and calling it out in the release notes.

While I think it makes sense to advertise the template method as the recommended setup, I think it makes sense to also show how to do it without, unless you plan to remove support for that completely.

daemonset/generic-kuberouter-cni-conf-template.yaml
```diff
@@ -0,0 +1,217 @@
...
+              fieldPath: spec.nodeName
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: KUBE_ROUTER_CNI_CONF_FILE
+          value: /etc/cni/net.d/10-kuberouter.conflist
+        - name: KUBE_ROUTER_CNI_CONF_TEMPLATE_FILE
+          value: /etc/kube-router/cni-conf.json
+        livenessProbe:
```
catherinetcai, line 89: @aauren pushed some recent changes to the yamls that added a [startupProbe](https://github.com/cloudnativelabs/kube-router/blob/master/daemonset/generic-kuberouter.yaml#L70), and a few new other fields. Could you also update the YAMLs here?

catherinetcai (approved):
LGTM. Thank you!

daemonset/generic-kuberouter-cni-conf-template.yaml
twz123: Done.

PR #2139, opened 2026-08-15
Prep v2.11.0 release

catherinetcai:
@aauren / @mrueg 

Lots of minor version bumps to prep for the v2.11.0 release.

Only thing that I noticed was broken was the Kind version bump, which I had to advanced manually. I'll have a patch to fix the version update for it.


.github/workflows/ci.yml
```diff
@@ -18,9 +18,9 @@ permissions:
   contents: read
 
 env:
-  BUILDTIME_BASE: &buildtime_base "golang:1.26.3-alpine3.23@sha256:91eda9776261207ea25fd06b5b7fed8d397dd2c0a283e77f2ab6e91bfa71079d"
-  RUNTIME_BASE: &runtime_base "alpine:3.23.4@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11"
-  GO_VERSION: &go_version "~1.26.3"
+  BUILDTIME_BASE: &buildtime_base "golang:1.26.6-alpine3.24@sha256:af8d6740070b8906d12eae1c3e3ea0957fb63f492051ea05e354c38ef9fe88df"
+  RUNTIME_BASE: &runtime_base "alpine:3.23.5@sha256:fd791d74b68913cbb027c6546007b3f0d3bc45125f797758156952bc2d6daf40"
```
aauren: @catherinetcai - I think that we need to bump the alpine pinned version: https://github.com/cloudnativelabs/kube-router/blob/master/build/dependency-updater/versions.lock.yaml#L15 so that we pick up 3.24.1

Makefile
```diff
@@ -18,31 +18,31 @@ MAKEFILE_DIR=$(dir $(realpath $(firstword $(MAKEFILE_LIST))))
...
-DOCKER_BUILD_IMAGE?=golang:1.26.3-alpine3.23@sha256:91eda9776261207ea25fd06b5b7fed8d397dd2c0a283e77f2ab6e91bfa71079d
+DOCKER_BUILD_IMAGE?=golang:1.26.6-alpine3.24@sha256:af8d6740070b8906d12eae1c3e3ea0957fb63f492051ea05e354c38ef9fe88df
 ## These variables are used by the Dockerfile as the bases for building and creating the runtime container
 ## During CI these come from .github/workflows/ci.yaml below we define for local builds as well
 GO_CACHE?=$(shell go env GOCACHE)
 GO_MOD_CACHE?=$(shell go env GOMODCACHE)
 BUILDTIME_BASE?=$(DOCKER_BUILD_IMAGE)
 # See Versions: https://hub.docker.com/_/alpine
-RUNTIME_BASE?=alpine:3.23.4@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11
+RUNTIME_BASE?=alpine:3.23.5@sha256:fd791d74b68913cbb027c6546007b3f0d3bc45125f797758156952bc2d6daf40
```
aauren: Same here, but the bump that unpins 3.23 should cause this to pick up 3.24.1 once its run again

aauren (changes requested):
Just the alpine version pinning. Everything else looks good.

If the RELEASE information doesn't say to check the version pinning in that file then we should add that so that no one else hits this.

.github/workflows/ci.yml
catherinetcai: Bumped and re-ran make update-deps and checked in the updated version.

Makefile
catherinetcai: Bumped to 3.24.1 as well here!

aauren (approved):
LGTM! Thanks @catherinetcai !

PR #2140, opened 2026-08-17
chore(deps): Bump GoBGP version in go.mod to match GoBGP binary

catherinetcai:
#### What type of PR is this?

dependency

I updated GoBGP to 4.8.0 through a bunch of dependency bumps, but noticed I didn't advance GoBGP in the go.mod to match.

N/A

N/A

NONE

No

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/moby/moby/client v0.5.0
 	github.com/onsi/ginkgo/v2 v2.25.3
 	github.com/onsi/gomega v1.42.1
-	github.com/osrg/gobgp/v4 v4.7.0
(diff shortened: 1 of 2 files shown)


aauren (approved):

twz123:
Just an idea: k0s has basically a similar problem with several dependencies whose versions are defined in multiple places. We use unit tests to ensure they stay in sync. Might be worth considering for kube-router, as well.

aauren:
> Just an idea: k0s has basically a similar problem with several dependencies whose versions are defined in multiple places. We use unit tests to ensure they stay in sync. Might be worth considering for kube-router, as well.

That's a cool idea, thanks for sharing!

Issue #2123, opened 2026-07-10
Add monitoring phase for NetworkPolicies: log violations instead of dropping via annotation


dottorblaster:
**Is your feature request related to a problem? Please describe.**
In the context of the Kubernetes-native network enforcer that me and others are building, we would like to have a monitoring phase for the deployed network policy so the user can verify everything is right before switching to "enforce mode". Policy violations should be logged, but the traffic should still be allowed.

**Describe the solution you'd like**
From the investigation we conducted:

> - use a custom annotation on the k8s network policy to mark the policy as in "monitor" mode. Not ideal but would avoid a new CRD.
> - if the annotation is present kube-router could change the iptables rules to just log instead of dropping the traffic. This would mean that the notification in monitor mode would be collected through system logs
> - if the annotation is not present the policy is enforced in the cluster.

**Describe alternatives you've considered**
At first we thought about a dedicated CRD for this, but maybe having just the annotation would be a better fit for the current architecture

**Additional context**
We already looked into this from a user standpoint: https://github.com/rancher-sandbox/network-enforcer/issues/58

We would still love to hear feedback about anything of this.

aauren:
I think that this would be an interesting feature.

One nuance, is that it would be using NFLOG targets for the logging as kube-router doesn't have any components in the actual data plane and is purely control plane oriented. That means that users would have to have an NFLOG collector active on their system, not sure how many users have that.

Are you / k3s interested in helping develop this feature? If so, we could probably do some more detailing of the design here and then work towards a PR.

One other thing to note, the netfilter implementation is currently undergoing some pretty extreme changes. There is #2064 that would need to land first, and then after that, the project has ~a month long retrofit that we want to do to remove dependency on iptables across the rest of the controllers and make future implementations more flexible.

We would want to see all of that work land before this feature was worked on.

dottorblaster:
Thanks for this @aauren!

Yes me/k3s is interested in developing this feature, and I'd like toward a design here and towards a PR later.

About the work in #2064, I think we have some time on our end and we can wait, thanks for letting me know!

About the design, this is what I thought: log instead of drop" only works when a pod is selected only by monitored policies. If a pod is selected by both an enforced and a monitored policy (the common case when validating a new policy on a protected workload), flipping the terminal REJECT drops enforcement for the whole pod.

[...]

PR #2141, opened 2026-08-21
Added monitor network policy

rbrtbnfgl:
#### What type of PR is this?

feature

This PR is an initial implementation of #2123 in case a policy is defined with the annotation `kube-router.io/network-policy.monitor: "true"` the packet is marked with an additional mark `0x40000` and logged as it was dropped but even if the mark `0x10000` is set. Right now it works for iptables.

#2123 

* What tool was used: Copilot 
* To what extent was the tool used? To draft some part of the code
* If drafted, how detailed of a plan did you create for the AI? Part of the code human written and used AI to help 

Tested with different policies with and without the annotation

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/netpol/network_policy_controller.go
+++ b/pkg/controllers/netpol/network_policy_controller.go
@@ -126,6 +126,7 @@ type networkPolicyInfo struct {
 	name        string
 	namespace   string
 	podSelector labels.Selector
+	annotations map[string]string
 
 	// set of pods matching network policy spec podselector label selector
 	targetPods map[string]podInfo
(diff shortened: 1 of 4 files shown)


aauren:
Thanks @rbrtbnfgl! Just letting you know that we've seen this. As mentioned over here https://github.com/cloudnativelabs/kube-router/issues/2123#issuecomment-4939590896 I think that this will likely sit for just a bit. We have some rather large changes coming in for the netfilter code and I want to land those first before accepting this work.

But we are grateful for the submission and we'll wrap back around to this as soon as we can!

PR #2162, opened 2026-09-05
Make healthcheck consistent chunked

aauren:
FYI @twz123 / @rbrtbnfgl 

#### What type of PR is this?

bug

Standardizes what a heartbeat means across NPC, NSC, NRC, the route syncer, and the load balancer allocator: every
controller now beats at iteration start and end, unconditionally, instead of skipping the beat when a sync failed.
Gating liveness on sync success was turning unfixable problems (a stuck BGP peer, a rejected Service, an
unreplaceable route) into a CrashLoopBackOff, which takes kube-router off the node instead of leaving a
stale-but-working rule set in place.

Sync outcomes are no longer silently dropped. They're reported through two new metrics,
kube_router_controller_sync_last_success and kube_router_controller_sync_failures_total, via a single
RecordSyncResult entry point. docs/health.md documents the new contract and includes alert guidance for the
staleness case /healthz intentionally no longer catches.

Also fixes a few related bugs found along the way: advertiseVIPs/withdrawVIPs/syncInternalPeers now return
their failures instead of dropping them, and syncInternalPeers defers peer deletion while any Node fails to
parse so a transient address outage can't tear down a working peer.

None

* What tool was used: Claude Code (frequently reviewed by Codex)
* To what extent was the tool used? Drafted the implementation and all four commits from a detailed prompt/plan
* If drafted, how detailed of a plan did you create for the AI? Fairly detailed - the heartbeat contract and
failure-propagation approach were specified up front, Claude worked out the per-controller implementation

Deployed to a live cluster and validated failure scenarios directly: failed a controller's sync and confirmed
kube_router_controller_sync_failures_total incremented while /healthz kept returning healthy for that controller.

```release-note
Controller heartbeats now fire at both the start and end of every sync, regardless of success or failure, instead
of being skipped on a failed sync. This means kube-router will no longer be restarted by its livenessProbe just
because a sync is failing (a restart can't fix a bad BGP peer or rejected Service anyway). If you rely on the
livenessProbe restarting kube-router to recover from sync failures, switch to alerting on the new
kube_router_controller_sync_last_success / kube_router_controller_sync_failures_total metrics instead - see
docs/health.md for example alert rules.
```

docs/health.md has a new section spelling out exactly what a heartbeat does and doesn't assert, plus sample
Prometheus alert rules for the two new metrics - worth a read before reviewing the code changes.


aauren:
Fixes: #2163

pkg/controllers/proxy/network_services_controller.go
```diff
@@ -350,10 +350,12 @@ func (nsc *NetworkServicesController) Run(healthChan chan<- *healthcheck.Control
...
-		err := nsc.doSync()
-		if err != nil {
-			klog.Fatalf("failed to perform initial full sync %s", err.Error())
+		// We come up degraded rather than fatal, because a restart can't fix a rejected service
+		syncErr := nsc.doSync()
+		if syncErr != nil {
+			klog.Errorf("initial full sync did not complete cleanly, continuing anyway: %v", syncErr)
 		}
+		metrics.RecordSyncResult(healthcheck.NetworkServicesController, syncErr)
 		nsc.readyForUpdates = true
```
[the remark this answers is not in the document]
aauren: Fixed!

docs/health.md
```diff
@@ -30,3 +30,84 @@ For example, if kube-router is started with
 
 If the route controller, policy controller or service controller exits its main loop and does not publish a heartbeat
 the `/healthz` endpoint will return a error 500 signaling that kube-router is not healthy.
+
+## What A Heartbeat Does And Does Not Assert
```
twz123: Strange title case here. I'd rather write "What a heartbeat does and does not assert". Same for the other headlines.

docs/health.md
```diff
@@ -30,3 +30,84 @@ For example, if kube-router is started with
...
+
+This is worth being explicit about, because the two shipped daemonsets wire `/healthz` as a **livenessProbe** with
+`periodSeconds: 3` and `failureThreshold: 3`. Roughly nine seconds of unhealthy and the kubelet kills the container,
+so what a heartbeat means decides when kube-router gets restarted.
```
twz123: My gut feeling says this belongs into the respective DaemonSets as a comment? Feels a bit off as the first paragraph in this section? I'd try to detangle the whole section a bit from the concrete Pod's livenessProbe details, or at least defer those things into a separate section at the end of this doc. "What it means for the kube-router DaemonSet" or something along those lines.

docs/health.md
```diff
@@ -30,3 +30,84 @@ For example, if kube-router is started with
...
+
+A heartbeat asserts that **a control loop is turning**. It does not assert that an iteration accomplished anything.
+Every controller sends one heartbeat when an iteration starts and another when that iteration's work returns,
+unconditionally, whether it succeeded or failed.
```
twz123: That's somehow duplicate with the first sentence in this doc. Maybe just add the last half sentence "whether it succeeded or failed" at the beginning of this doc?

docs/health.md
```diff
@@ -30,3 +30,84 @@ For example, if kube-router is started with
...
+sync never sends its end beat, so a wedged goroutine is still detected within one sync period plus grace of the moment
+it stuck.
+
+The tradeoff is that a sync which completes but takes far longer than usual (while staying under the sync period) will
+not show up on `/healthz` at all. That's intentional: slowness while making progress is not a liveness problem, and it
+is visible instead through the sync-time histograms (for example `controller_iptables_sync_time`) and the sync metrics
+described below.
+
+So `/healthz` returning `OK` means the process is alive and its loops are turning. It does **not** mean the node's
+networking is correctly programmed.
```
twz123: I'd drop the first sentence, and move "It does **not** mean the node's networking is correctly programmed." up and replace "It does not assert that an iteration accomplished anything." with it.

docs/health.md
```diff
@@ -30,3 +30,84 @@ For example, if kube-router is started with
...
+heartbeats on the same unconditional contract, but it allocates per service off a queue rather than in one sync pass,
+so it has no single per-iteration outcome to report and does not appear under these two metrics.
+
+Both series are created the first time a controller reports any result, successful or not. That matters, because a
+controller that has failed every sync since startup still needs to be visible: its
+`kube_router_controller_sync_last_success` sits at `0` rather than being absent, so a staleness alert matches it
+instead of silently evaluating against an empty vector.
+
+You'll want an alert on a controller that is alive but not making progress, which is the case `/healthz` no longer
+catches for you. Something like the following:
```
twz123: Might be good to say that the below snippet is an example in how to set this up if you're using Prometheus or some compatible technology.

docs/health.md
```diff
@@ -30,3 +30,84 @@ For example, if kube-router is started with
...
+- alert: KubeRouterSyncFailing
+  expr: time() - kube_router_controller_sync_last_success > 900
+  for: 10m
+  labels:
+    severity: warning
+  annotations:
+    summary: "kube-router {{ $labels.controller }} on {{ $labels.instance }} has not completed a sync in 15m"
+    description: >
+      The control loop is still running (kube-router would be reporting unhealthy otherwise), but its last
+      several sync attempts have failed. Check the pod logs for the abort reason; a restart will not help.
```
twz123: > a restart will not help.

That's a bit over-confident, isn't it :) Pretty sure there's scenarios in which a restart will help, even if the control loops aren't locked up.

docs/health.md
```diff
@@ -30,3 +30,84 @@ For example, if kube-router is started with
...
+```
+
+The 900 second threshold assumes the default 5 minute sync periods of the three main controllers, so it fires after
```
twz123: Maybe move this as inline comments into the YAML snippet?

docs/health.md
```diff
@@ -30,3 +30,84 @@ For example, if kube-router is started with
...
+roughly three consecutive missed syncs for those. `RouteSyncController` syncs every 60 seconds by default, so the same
+threshold represents around 15 missed route syncs; tighten it (or split out a separate alert for that label) if you
+want faster detection there. Adjust the threshold if you've changed `--iptables-sync-period`, `--routes-sync-period`,
+`--ipvs-sync-period`, or `--injected-routes-sync-period`.
+Note that a controller which has never succeeded is stuck at `0` and therefore trips the threshold as soon as the
+`for` window elapses, which is the intent - it has never programmed anything, so it deserves to page sooner than one
+that merely went stale.
+
+If you'd rather alert on the rate of failures than on staleness,
```
twz123: Wouldn't it make sense to do both?

docs/health.md
```diff
@@ -30,3 +30,84 @@ For example, if kube-router is started with
...
+
+The tradeoff is that a sync which completes but takes far longer than usual (while staying under the sync period) will
+not show up on `/healthz` at all. That's intentional: slowness while making progress is not a liveness problem, and it
+is visible instead through the sync-time histograms (for example `controller_iptables_sync_time`) and the sync metrics
+described below.
+
+So `/healthz` returning `OK` means the process is alive and its loops are turning. It does **not** mean the node's
+networking is correctly programmed.
+
+## Monitoring Sync Correctness
```
twz123: That kinda warrants its own file, in a way? I mean, monitoring is not part of health, although you probably really want to monitor the health. That said, just realized there's already `metrics.md`. Maybe move this section over, and just link it here?

go.mod
```diff
@@ -97,6 +97,7 @@ require (
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/k-sone/critbitgo v1.4.0 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
```
twz123, line 100: Weird. No direct dependency changes, but an indirect dependency added?

pkg/controllers/lballoc/lballoc.go
```diff
@@ -474,11 +474,19 @@ func (lbc *LoadBalancerController) Run(healthChan chan<- *healthcheck.Controller
...
+			// walk finished. walkServices feeds addChan, which only this select drains, so it
+			// can't run inline without deadlocking
 			healthcheck.SendHeartBeat(healthChan, healthcheck.LoadBalancerController)
-			if isLeader {
-				go lbc.walkServices()
-			}
+			leader := isLeader
+			go func() {
+				if leader {
+					lbc.walkServices()
```
twz123: Not introduced by this PR but realized when reviewing this: Under pathological circumstances, there can be two goroutines that execute walkServices concurrently, if the ticker fires right after this kube-router process became the leader. Might be worth fixing.

pkg/controllers/lballoc/lballoc.go
```diff
@@ -474,11 +474,19 @@ func (lbc *LoadBalancerController) Run(healthChan chan<- *healthcheck.Controller
 				lbc.allocateChan <- svc
 			}
 		case <-timer.C:
-			timer.Reset(time.Minute)
+			timer.Reset(lbc.syncPeriod)
+			// Beat at walk start as well as walk end, matching the dual-beat contract of the sync
+			// controllers: the first beat asserts this select loop is alive, the second that the
+			// walk finished. walkServices feeds addChan, which only this select drains, so it
+			// can't run inline without deadlocking
 			healthcheck.SendHeartBeat(healthChan, healthcheck.LoadBalancerController)
```
twz123: The walkServices goroutine that's started whenever this kube-router process became the leader doesn't send heartbeats at all.

pkg/controllers/netpol/network_policy_controller.go
```diff
@@ -109,6 +109,19 @@ type NetworkPolicyControllerBase struct {
 	networkPolicyEventHandler cache.ResourceEventHandler
 }
 
+// syncAndReport runs one full policy sync, beating at both the start and the end so the health
+// deadline budgets a full sync period per iteration instead of the first sync's duration, and
+// records the outcome unconditionally so the beats assert only that this goroutine isn't wedged
+func (npc *NetworkPolicyControllerBase) syncAndReport(fullPolicySync func() error) {
```
twz123: I'd rather have expected such a helper in the healthcheck package à la

```
func RunWithBeats(ch, component, f) {
	healthcheck.SendHeartBeat(ch, component)
	defer healthcheck.SendHeartBeat(ch, component)
	f()
}

File: pkg/controllers/netpol/network_policy_controller.go [hunk 15]
```diff
@@ -109,6 +109,19 @@ type NetworkPolicyControllerBase struct {
...
+	healthcheck.SendHeartBeat(npc.healthChan, healthcheck.NetworkPolicyController)
+	err := fullPolicySync()
+	if err != nil {
+		klog.Errorf("aborted full sync of network policies: %v", err)
```
twz123: Is "aborted" the right phrasing? As an outsider, I'd have expected something along the lines of "failed to do a".

pkg/controllers/netpol/npc_iptables.go
```diff
@@ -218,17 +212,18 @@ func (npc *NetworkPolicyControllerIptables) fullPolicySync() {
 		klog.V(1).Infof("Restoring %v iptables rules took %v", ipFamily, restoreEndTime)
 
 		if err != nil {
-			klog.Errorf("Aborting sync. Failed to run iptables-restore: %v\n%s",
-				err.Error(), npc.filterTableRules[ipFamily].String())
-			return
+			klog.Errorf("iptables-restore for %v failed against the following rule set:\n%s",
```
twz123: Wouldn't this end up twice in the logs? I'm assuming that the caller will log the returned error eventually.

pkg/controllers/netpol/npc_nftables.go
```diff
@@ -59,6 +59,11 @@ type NetworkPolicyControllerNftables struct {
 	// Policy chains whose transaction failed last sync; syncPodFirewallChains fails their target
 	// pods closed rather than leaving them default-allow. Guarded by npc.mu.
 	failedPolicyChains map[string]bool
+
+	// Joined errors behind failedPolicyChains, returned at the end of fullPolicySync so a partial
+	// failure still counts as a failed sync without aborting the fail-closed programming. Guarded
+	// by npc.mu.
+	failedPolicyErr error
```
twz123: I'd probably go with `failedPolicyErrs []error` here, and only call `errors.Join(...)` on the slice whenever an opaque error is needed. Same for the other similar error variables in this PR.

pkg/controllers/netpol/npc_iptables.go
aauren: It does sorta go into the logs twice, but I like this one because it includes the full tables rule that failed whereas the upper error does not.

So I kept it, but I rephrased the error just a bit to make it more apparent what the value of it is.

docs/health.md
aauren: Good catch on style non-conformity.

docs/health.md
aauren: Yeah, there were a few dups in here and I liked your advice so I implemented all of them and tried to make the sections flow just a bit better.

docs/health.md
aauren: Fair :wink:

I removed that bit and left it up to the operator to know what to do.

docs/health.md
aauren: Done

docs/health.md
aauren: Also a good recommendation, I moved the second recommendation as a secondary alert example.

docs/health.md
aauren: Agree, moved to metrics.md.

Maybe at some point I'll make a dedicated prometheus alerts example file in there.

go.mod
aauren: Yeah, I was surprised on this one too, but I guess it has to do with the way that modern go.mod semantics handle test dependencies?

pkg/controllers/netpol/npc_nftables.go
aauren: Updated

pkg/controllers/netpol/network_policy_controller.go
aauren: This is a really good point and something I honestly should have caught, especially since the original implementation had the same core set of logic strewn across the code base.

I've now consolidated the logic into a `healthcheck.RunSync()` which does the heart-beating surrounding a sync. And a `metrics.RunObservedSync()` which calls `RecordSyncResult()` after a `healthcheck.RunSync()`.

As a bonus this also forced many of the more lengthy `select` clauses in some controllers to actually get a function written that encapsulates the logic, something that I haven't liked for a bit now.

pkg/controllers/netpol/network_policy_controller.go
aauren: Updated phrasing.

docs/metrics.md
```diff
@@ -31,6 +31,64 @@ By enabling
 your Prometheus configuration & adding the required annotations, Prometheus can automatically discover & scrape
 kube-router metrics.
 
+## Monitoring sync correctness
+
+Because kube-router's /healthz endpoint deliberately does comment on sync correctness, kube-router emits two metrics to
```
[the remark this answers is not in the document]
aauren: Updated!

docs/health.md
aauren: I added a comment to each of the daemonsets we provide that contained a livenessProbe.

docs/health.md
aauren: Done

pkg/controllers/lballoc/lballoc.go
aauren: This should be a bit better now, rather than maintaining a completely separate set of logic, Go 1.26 allows us to just call the timer with a timer.Reset(0). This doesn't completely remove the pathological case here, but it should reduce it significantly. Beyond that, additional remediation for this should probably extend beyond this controller, because a lot of our other controllers contain this same fragility.

pkg/controllers/lballoc/lballoc.go
aauren: Fixed!

